PR #1030, opened 2026-02-23
⚡ Bolt: Optimize trade ledger loading

rozavala:
Optimization of `get_trade_ledger_df` in `performance_analyzer.py`.
Previously, the function loaded cached dataframes (with string timestamps) and then re-parsed the entire concatenated dataframe's timestamps on every call.
Now, timestamps are parsed once when the file is loaded into the LRU cache. The concatenation then preserves the datetime objects.
This significantly reduces the overhead of loading trade history, especially as the archive grows.

Verified with a new unit test `tests/test_ledger_loading.py` and existing tests.
Benchmark simulation showed a 2.88x speedup.

---
*PR created automatically by Jules for task [10271956764523808430](https://jules.google.com/task/10271956764523808430) started by @rozavala*

### DIFF
--- a/performance_analyzer.py
+++ b/performance_analyzer.py
@@ -15,6 +15,7 @@ from notifications import send_pushover_notification
 # Decision signals: lightweight summary of Council decisions
 from trading_bot.decision_signals import get_decision_signals_df
 from trading_bot.performance_graphs import generate_performance_charts
+from trading_bot.timestamps import parse_ts_column
 from config_loader import load_config
 
 logger = logging.getLogger("PerformanceAnalyzer")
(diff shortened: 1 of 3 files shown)


rozavala:
## Review

The optimization in `performance_analyzer.py` is solid — parsing timestamps in the LRU-cached loader avoids redundant `pd.to_datetime` on every call. Clean approach, ~3x speedup is meaningful.

**However, the tests introduce `sys.modules` mocking at module level:**

```python
sys.modules["matplotlib"] = MagicMock()
sys.modules["matplotlib.pyplot"] = MagicMock()
sys.modules["plotly"] = MagicMock()
sys.modules["plotly.graph_objects"] = MagicMock()
sys.modules["trading_bot.performance_graphs"] = MagicMock()
```

This is a known anti-pattern in this codebase — module-level `sys.modules` hacks pollute all downstream tests that import the same modules. All dependencies (matplotlib, plotly, etc.) are installed in the test environment, so these mocks are unnecessary.

**Closing** because the `sys.modules` pollution would break test isolation. The core optimization is good — will re-implement without the mocking.

PR #1104, opened 2026-02-27
⚡ Bolt: Parallelize Sentinel network I/O with asyncio.gather

rozavala:
💡 **What**: Refactored the core loops in `WeatherSentinel` and `MacroContagionSentinel` to use `asyncio.gather`.
🎯 **Why**: Previously, these sentinels awaited asynchronous I/O calls sequentially (e.g. iterating over API calls for each region or ticker in a for loop). In an event-driven architecture, reducing cycle time is critical for responsiveness. Parallelizing these fetches dramatically reduces the wait time.
📊 **Impact**: Reduces total network latency for these checks from $O(N)$ (sum of all fetch times) to $O(1)$ (bounded by the single slowest request), ensuring the Sentinel loop iterates faster.
🔬 **Measurement**: Validated by running tests in `tests/test_sentinels.py` using `pytest`. No assertions were broken.

---
*PR created automatically by Jules for task [7601707124066284777](https://jules.google.com/task/7601707124066284777) started by @rozavala*

### DIFF
--- a/trading_bot/sentinels.py
+++ b/trading_bot/sentinels.py
@@ -2322,9 +2322,13 @@ class MacroContagionSentinel(Sentinel):
             # OPTIMIZATION: Fetch historical data for all tickers concurrently.
             # This reduces total latency for the contagion check.
             tasks = [self._get_history(ticker, period="10d", interval="1d") for ticker in tickers.values()]
-            results = await asyncio.gather(*tasks, return_exceptions=False)
+            results = await asyncio.gather(*tasks, return_exceptions=True)
 
-            for name, hist in zip(names, results):
+            for name, result in zip(names, results):
+                if isinstance(result, Exception):
+                    logger.debug(f"Insufficient data for {name} ({tickers[name]}), skipping from contagion check: {result}")
+                    continue
+                hist = result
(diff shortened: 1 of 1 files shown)


rozavala:
Good optimization — parallelizing weather region fetches and contagion ticker history with `asyncio.gather` is the right pattern.

**Issue to fix before merge:**

The contagion check at L2327 uses `return_exceptions=False`:
```python
results = await asyncio.gather(*tasks, return_exceptions=False)
```

This means if **any single ticker** fails (network timeout, yfinance rate limit, etc.), the entire contagion check crashes with an unhandled exception. The original sequential code had per-ticker `try/except` that would skip failures gracefully.

**Fix:** Change to `return_exceptions=True` and add the same exception-handling pattern used in the weather sentinel:
```python
results = await asyncio.gather(*tasks, return_exceptions=True)

for name, result in zip(names, results):
    if isinstance(result, Exception):
        logger.debug(f"Insufficient data for {name} ({tickers[name]}), skipping from contagion check")
        continue
    hist = result
    if len(hist) < 2:
        ...
```

The weather sentinel pattern (L718-724) is correct — please replicate it for the contagion check.

PR #1427, opened 2026-03-26
Add 'executed' flag to council_history

rozavala:
## Summary

council_history records every signal, but there was no way to tell which ones resulted in actual orders vs which were blocked. This adds an `executed` column.

### How it works
1. When a council decision is logged → `executed='False'`
2. When order_manager queues an order → `mark_council_executed()` updates matching `cycle_id` rows to `executed='True'`
3. Legacy rows (pre-feature) → `executed=''` (empty = unknown)

### Downstream uses (not implemented yet, but now possible)
- **Scorecard**: Filter to executed-only for "real" win rate vs "signal accuracy" win rate
- **Funnel**: Show signal→execution conversion rate directly from council_history
- **Signal Overlay**: Color-code executed (solid) vs blocked (faded) markers
- **Reconciliation**: Skip non-executed rows to save IB API calls

### No breaking changes
- Empty string backfill for legacy data
- All existing pages/processes continue working — they don't filter on `executed` yet
- Dashboard pages can adopt the flag incrementally

## Test plan
- [x] All 1042 tests pass
- [ ] Verify new council_history rows have `executed=False` by default
- [ ] Verify executed signals show `executed=True` after order queue
- [ ] Verify schema auto-migration adds column to existing CSV


rozavala:
Critical Review: APPROVE. All 1042 tests pass. Clean v7 schema addition following established patterns. order_manager.py change is fail-open and post-execution. Backward compatible with legacy data. No consumers yet (data-layer-first). Minor gap: no unit tests for mark_council_executed(). Labeled Ready to be merged.

PR #1422, opened 2026-03-26
Fix P&L units: convert from price units to USD across all reporting

rozavala:
## Summary

P&L values in `council_history.csv` were stored in **price units** (e.g., `0.082` for NG meaning an $820 move), not dollars. This caused every downstream consumer — 7 dashboard pages, Pushover reports, and trade journals — to display incorrect values.

### Root Cause
`reconciliation.py` computed `pnl = exit_price - entry_price` without applying the contract's dollar multiplier (375× for KC, 10,000× for NG).

### Fixes

1. **reconciliation.py** — Apply `get_dollar_multiplier()` before storing `pnl_realized`
2. **performance_analyzer.py** — Compute daily realized P&L from trade ledger (`total_value_usd`) instead of `commissionReport.realizedPNL` (which is zero for combo fills)
3. **Migration script** — `scripts/migrations/migrate_pnl_to_usd.py` converts existing historical data. Idempotent (won't double-convert).

### Impact

| Before | After |
|--------|-------|
| Dashboard: "$0.66 realized" | Dashboard: "$6,600 realized" |
| Pushover: "Realized $0.00" | Pushover: "Realized $-370.00" |
| Win rate: correct (sign-based) | Win rate: still correct |

### Deployment Steps

After merge, run migration on each environment:
```bash
DATA_DIR=/path/to/data python scripts/migrations/migrate_pnl_to_usd.py
```

## Test plan
- [x] All 1037 tests pass (reconciliation test updated for new units)
- [x] Migration script dry-run verified against production data
- [ ] Run migration on DEV data after deploy
- [ ] Run migration on PROD data after deploy
- [ ] Verify dashboard Financials shows dollar-scale P&L
- [ ] Verify Pushover daily report shows non-zero realized P&L

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1428, opened 2026-03-27
Fix P&L calculation for cross-window positions

rozavala:
## Summary

Fixes a critical bug in the Financials page P&L calculation where positions opened before the date filter window showed as false wins.

### Bug
When selecting "Mar 25-26" and a position was opened on Mar 24 and closed on Mar 25:
- The date filter dropped the entry fills (Mar 24 = outside window)
- P&L was computed from close fills only → the close credit looked like profit
- Every loss showed as a "win"

**NG Mar 25-26**: Dashboard showed +$760, 55.6% win rate. Actual: -$850, 0.0% win rate.

### Fix
Keep the unfiltered `trade_df_all`. Use the date-filtered view only to identify which positions CLOSED in the window, then compute full P&L from ALL fills (including entries before the window).

## Test plan
- [x] All 1042 tests pass
- [x] Verified NG P&L changes from +$760 → -$850 (correct)
- [x] Verified KC P&L unchanged at -$288.75 (was already correct — entry and close both in window)
- [ ] Verify on live dashboard


rozavala:
## Code Review — Approved

### What was reviewed

Single-file change to `pages/4_Financials.py` (16 additions, 9 deletions). No changes to order execution, compliance, or IB interface code.

### Bug confirmed

When the date filter window excluded entry fills (e.g., position opened Mar 24, closed Mar 25, filter set to Mar 25-26), P&L was computed from close fills only. Since close fills are credits (positive `total_value_usd`), every loss appeared as a win. The PR description's example — NG showing +$760/55.6% when actual was -$850/0.0% — is consistent with this bug.

### Fix is correct

The fix preserves `trade_df_all` (unfiltered) alongside the date-filtered `trade_df`. It uses the filtered view only to identify which `position_id`s have close fills within the window, then computes full P&L from ALL fills via `trade_df_all`. The accounting identity (`sum(total_value_usd)` where BUY=debit, SELL=credit) is preserved.

### Edge cases verified

- **No date filter active**: `trade_df` equals `trade_df_all`, behavior identical to before
- **All fills within window**: Same position_ids and fills used, same result
- **No close fills in window**: `_closed_pids` is empty, `_ledger_pnl` stays empty — correct
- **Partial fills**: All fills for a position_id included via `.isin()` on `trade_df_all`
- **Zero P&L**: Handled by existing downstream code (unchanged)
- **Positions closed before window**: Correctly excluded (close fill not in filtered `trade_df`)

### Not redundant with recent PRs

- PR #1422 fixed P&L *units* (price to USD conversion)
- PR #1426 switched from theoretical signal P&L to trade ledger actual P&L
- This PR #1428 fixes a *date filtering* bug that #1426 introduced — it is a necessary follow-up

### Note on test coverage

No unit tests exist for the P&L computation logic on this page (pre-existing gap). Adding a test for the cross-window scenario would be valuable as a follow-up.

### Unable to run tests in this environment

pytest was not available in the review environment. The PR description claims all 1042 tests pass. Recommend verifying CI passes before merging.

**Verdict: Ready to merge.** This is a clean, minimal, correct fix for a critical financial reporting bug that showed losses as wins.

PR #1457, opened 2026-03-30
🛡️ Sentinel: [MEDIUM] Prevent stack trace exposure in UI

rozavala:
This commit modifies the error handling blocks in `pages/5_Utilities.py` to prevent full python stack traces (`traceback.format_exc()`) from being displayed to the UI in production. This closes a potential information disclosure vulnerability which could leak sensitive system internals or dependency versions. Error traces will only be displayed when `current_env` is not set to `prod`, and otherwise logged securely to the application backend. Included a learning entry to `.jules/sentinel.md`.

---
*PR created automatically by Jules for task [14868151879531551701](https://jules.google.com/task/14868151879531551701) started by @rozavala*


rozavala:
Closing. While the code change is technically correct, the Streamlit dashboard is not publicly exposed (runs on localhost / Tailscale), so stack trace exposure is not a real security risk in our deployment.

Hiding tracebacks from the UI in production actually degrades the operator debugging experience — when an order or position close fails, seeing the stack trace immediately is more valuable than SSHing in to grep log files. The tracebacks only contain Python file paths and dependency versions, which are not sensitive for a system we operate ourselves.

If we want defense-in-depth later, a better approach would be to always log the traceback to the backend while still showing it in the UI expander for immediate debugging.

PR #1468, opened 2026-03-31
fix: scope emergency-close TMS sweep + fix IBKR option symbol filter

rozavala:
## Summary

- **Bug 1 — IBKR option symbol filter**: `p.contract.symbol == commodity_symbol` used `'SB'` but IBKR reports SB options with `symbol='SO'` (same pattern: KC options=`'KO'`, NG=`'LNE'/'LN1-5'`). Fixed in `emergency_hard_close` (orchestrator.py) and all three filters in `close_stale_positions` + post-close verification (order_manager.py) using a `_IBKR_OPTION_SYMBOLS` prefix map.

- **Bug 2 — TMS sweep too broad**: `emergency_hard_close` invalidated ALL active TMS theses when `closed > 0`, not just the theses for positions that were actually filled. The fix tracks `closed_local_symbols` during the fill loop, resolves them to `position_id`s via the trade ledger, and only invalidates matching TMS theses.

- **Data fix script**: `scripts/reactivate_sb_theses.py` — one-time script that re-activated 6 incorrectly deactivated SB theses against the production TMS (already run). Included for audit trail.

## Impact

On 2026-03-31, `emergency_hard_close` closed 1 SON6 position but then swept all 7 active SB TMS theses, leaving 6 still-open positions (4×SOV6 + 2×SOH7) with `active=false`. This caused Portfolio page to show 0 SB positions and Cockpit to fall back to the stale IBKR leg count.

## Test plan

- [x] All 1082 existing tests pass
- [x] Production TMS re-activated: 6 active theses now match 6 open IBKR positions; SON6 (legitimately closed) remains inactive


rozavala:
**Code Review — Claude Code**

The core logic of both bug fixes is **correct and addresses a real production incident**. The IBKR symbol mapping fix and the scoped TMS sweep are both well-targeted. However there are two minor issues that should be fixed before merge:

---

### Issue 1 — `_IBKR_OPTION_SYMBOLS` duplicated in two files (maintenance hazard)

The lookup table is defined as a local variable in **two separate places**:
- `orchestrator.py` inside `emergency_hard_close`
- `trading_bot/order_manager.py` inside `close_stale_positions`

If a new commodity symbol mapping needs to be added or corrected in the future, both must be updated in sync. This will inevitably drift. The map should be extracted to a single shared constant — `trading_bot/utils.py` or `trading_bot/ib_interface.py` are natural homes.

**Suggested fix:** Define `IBKR_OPTION_SYMBOLS: dict[str, tuple[str, ...]]` once in `utils.py` and import it in both `orchestrator.py` and `order_manager.py`.

---

### Issue 2 — Python 3.9 incompatibility in `reactivate_sb_theses.py`

```python
def reactivate_theses(data_dir: str | None = None) -> int:
```

The `X | Y` union type syntax in annotations requires Python 3.10+. The project targets Python 3.9+. This will raise a `SyntaxError` on Python 3.9. Since the script has reportedly already run in production it's low impact, but it should be corrected for consistency.

**Suggested fix:** Change to `Optional[str]` from `typing`:
```python
from typing import Optional
def reactivate_theses(data_dir: Optional[str] = None) -> int:
```

---

Both issues are minor and confined to non-execution-path code (a shared constant extract and a type annotation fix). The critical bug fixes in `orchestrator.py` and `order_manager.py` are correct. Please fix and re-push.

---

PR #1580, opened 2026-04-13
fix: lightweight VaR recompute post-close to prevent overnight order submission risk

rozavala:
## Summary
The previous attempt to sync weekend VaR state introduced a dangerous regression by running a full `run_position_audit_cycle` after the market close. That function actively submits orders to IBKR, which would queue as GTC orders and blindly execute at the next session open.

## Fixes
1. Removed `run_position_audit_cycle` from the post-close tasks.
2. Created a dedicated `recompute_var_state` async function that purely computes portfolio VaR and saves state to disk without validating theses or interacting with orders.
3. Registered the new task in `FUNCTION_REGISTRY` and `RECOVERY_POLICY`.
4. Fixed the test suite's blind spot by adding the post-close task to the `_make_session_config` test fixture, successfully verifying that no order-submitting tasks run after the market closes.


rozavala:
Addressed review comments: Added configure_market_data_type(ib) and removed redundant _save_state() call.

Issue #1594, opened 2026-04-14
bug: Sentinel Effectiveness Alert wrongly reports 0 council decisions due to Pandas date-matching bug


rozavala:
## Summary
The Sentinel Effectiveness Alert in `orchestrator.py` incorrectly reports `Trades from sentinels: 0` and `Diagnosis: Alerts fired but no council decisions`, even on days when dozens of council decisions occurred (e.g., 48 today for SB).

## Root Cause
In `orchestrator.py` (around line 3510), the code tries to count today's decisions by parsing `council_history.csv` with Pandas. It creates a `today_mask` by comparing the formatted timestamp column to `today_str`. However, because it forces `utc=True` during `to_datetime()`, the timezone-aware timestamps might not string-match exactly, or the dates could roll over the UTC boundary differently than expected, resulting in a count of `0`.

## Proposed Fix
Fix the Pandas date-matching logic in `sentinel_effectiveness_check` to ensure the timestamps in `council_history.csv` properly match `today_str` (e.g., by ensuring both are normalized to the same timezone and format before comparison).

Issue #1591, opened 2026-04-13
bug: stale news headlines (4+ months old) persisting in news cache


rozavala:
## Observed on
2026-04-13

## Log evidence
```
2026-04-13 23:12:36 - trading_bot.sentinels - DEBUG - Skipping stale headline (2025-12-16): Caffe Nero issues price warning to customers buyin...
```

## Description
A headline from **2025-12-16** (almost 4 months ago) is still present in the news cache and being evaluated by the NewsSentinel on each cycle. The sentinel correctly identifies and skips it, but the fact it is still in the cache indicates the news cache is not being pruned by age.

## Impact
- Over time, the cache will accumulate months of stale entries, increasing memory usage and evaluation overhead on every sentinel cycle.
- If the staleness check threshold is ever loosened, old headlines could re-surface as signals.
- It obscures whether the cache is receiving fresh data — a full cache of stale items looks the same as a healthy cache at a glance.

## Suggested fix
Add a TTL-based eviction to the news cache: on each write or on a scheduled basis, remove entries older than N days (e.g. 7 days). The `semantic_cache` already has `ttl_minutes` configured in `config.json` — apply a similar pattern to the news headline cache.

```python
# Example pruning in NewsSentinel._load_cache or equivalent
cutoff = datetime.utcnow() - timedelta(days=7)
self._cache = {k: v for k, v in self._cache.items() if v['published'] > cutoff}
```

## Files to investigate
- `trading_bot/sentinels.py` — `NewsSentinel` cache load/save logic
- `config.json` — any existing `news_cache_ttl` config key

rozavala:
**Proposed Fix:** I will add a TTL eviction step inside `NewsSentinel`. On each cache load or save, we can iterate through the cache dictionary and remove any headlines where the `published` date is older than a configurable threshold (e.g., 7 days).

Issue #1589, opened 2026-04-13
bug: MicrostructureSentinel 'Unsubscribe error: Not connected' recurring after market close


rozavala:
## Observed on
2026-04-13 (post-market)

## Log evidence
```
2026-04-13 17:00:25 - trading_bot.microstructure_sentinel - ERROR - Unsubscribe error: Not connected
2026-04-13 17:30:40 - trading_bot.microstructure_sentinel - ERROR - Unsubscribe error: Not connected
2026-04-13 21:00:48 - trading_bot.microstructure_sentinel - ERROR - Unsubscribe error: Not connected
```

## Description
MicrostructureSentinel is attempting to unsubscribe from IB market data ticks after the IB connection has already been closed (post-market disconnect). The `Not connected` error fires on a ~30-minute interval, suggesting it is tied to a periodic cleanup or subscription refresh loop that does not first check connection state before calling unsubscribe.

## Impact
- Noisy ERROR-level log entries that can obscure real errors during log review.
- If the unsubscribe failure prevents the sentinel from re-subscribing correctly on the next market open, it could cause the microstructure monitor to be silently inactive.
- The error reporter may count these as actionable errors and file spurious GitHub issues overnight.

## Suggested fix
In `MicrostructureSentinel.unsubscribe()` (or equivalent cleanup method), guard the IB call with a connection check:
```python
if self.ib.isConnected():
    self.ib.cancelMktData(contract)
```
Or wrap in a `try/except` that logs at WARNING (not ERROR) when disconnected, since this is expected behaviour post-close.

## Files to investigate
- `trading_bot/sentinels.py` — `MicrostructureSentinel` subscription/unsubscription logic
- `trading_bot/microstructure_sentinel.py` — if it exists as a separate module (per the log source name)

rozavala:
This is a harmless but noisy bug. The `MicrostructureSentinel` attempts to call `ib.cancelMktData(contract)` to clean up its subscriptions, but if the market has closed or the IB connection was already torn down by the orchestrator, it throws `Unsubscribe error: Not connected`.

**Proposed Fix:** Add a simple guard `if ib.isConnected():` before attempting to cancel the market data subscription in `trading_bot/microstructure_sentinel.py`.

Issue #1587, opened 2026-04-13
bug: drawdown guard fail-closed during KC check at 15:00 UTC — IB timeout during market hours


rozavala:
## Observed on
2026-04-13 15:00 UTC

## Log evidence
```
2026-04-13 15:00:15 - ib_insync.client    - ERROR - API connection failed: TimeoutError()
2026-04-13 15:00:15 - connection_pool     - ERROR - IB (KC_drawdown_check) connection failed: . Next retry backoff: 10.0s
2026-04-13 15:00:15 - Orchestrator        - ERROR - Drawdown guard check failed (fail-closed):
```

## Description
During active US trading hours, the IB connection timed out while the drawdown circuit breaker was performing its periodic check for KC. The system correctly failed closed (blocking any new trades), but:

1. **The error message is empty** — `"Drawdown guard check failed (fail-closed):"` with nothing after the colon. The underlying exception is being swallowed or stringified to `""`. This makes post-hoc diagnosis impossible.
2. **It is unclear whether the fail-closed state was cleared after the IB connection recovered**, or whether KC trading remained suppressed for the rest of the session.

## Impact
If the drawdown guard stayed in fail-closed mode after the IB connection recovered (~10s later), KC was locked out of trading for the rest of the day even though no drawdown limit was actually breached. This would silently suppress valid signals.

## Action items
1. Fix the exception logging: capture and log `repr(e)` or the full traceback, not just `str(e)`.
2. Check `drawdown_circuit_breaker.py` — does a fail-closed timeout auto-reset when the next successful check runs, or does it latch?
3. Add a Pushover alert when the drawdown guard fails closed during market hours (this is a high-impact silent failure).
4. Investigate why the IB connection timed out at 15:00 specifically — was there a connection pool exhaustion event? Check pool state around that time.

## Files to investigate
- `trading_bot/drawdown_circuit_breaker.py` — fail-closed logic and reset path
- `trading_bot/connection_pool.py` — pool exhaustion / timeout behavior
- `orchestrator.py` — drawdown guard call site and exception handling

rozavala:
This issue was already successfully fixed in **PR #1575** (which resolved the parent issue #1536). 

The PR implemented a 5-minute TTL grace window in `DrawdownGuard.check()` so that if an IBKR timeout occurs during market hours, the system safely falls back to the most recently cached state rather than tripping the circuit breaker and failing closed. The fix is live in PROD.

rozavala:
[...]

PR #1595, opened 2026-04-14
fix: 7-issue safety audit — VaR bypass, regime UNKNOWN, sentinel matching, drawdown, rate limits

### LINKED ISSUE #1583: bug: var_95 = 0.0 despite 17 open positions — VaR 95th percentile calculation failure
## Observed on
2026-04-13

## Description
`var_state.json` shows `var_95 = 0.0` and `var_95_pct = 0.0` while `var_99 = 3725.87` and `var_99_pct = 0.074965`. This is not credible: with 17 open positions across NG and SB, a zero 5th-percentile loss implies the portfolio is perfectly hedged at that quantile, which does not match the actual position composition.

## Observed state (`data/var_state.json`)
```json
{
  "var_95": 0.0,
  "var_95_pct": 0.0,
  "var_99": 3725.87,
  "var_99_pct": 0.074965,
  "equity": 49701.12,
  "position_count": 17,
  "commodities": ["NG", "SB"],
  "timestamp": "2026-04-13T18:32:10.296302+00:00",
  "last_attempt_status": "OK"
}
```

## Hypothesis
The VaR 95 simulation path likely produces a positive PnL (i.e. a net gain at the 5th percentile), which gets clipped or coerced to 0.0 instead of being reported as a negative VaR (zero risk). This can happen if:
- The P&L distribution is calculated as loss (positive = bad) but the 95th-percentile scenario lands in profit territory, and the code `max(0, ...)` clamps it.
- A sign convention mismatch between how individual leg deltas are summed and how the quantile is taken.
- The position set is genuinely net-short-vega / net-long-theta and the 95% scenario (small move) favors the portfolio — but then `var_99_pct` should also be small, which it isn't.

## Impact
- Compliance Guardian uses VaR to gate new trades. A reported `var_95 = 0.0` may cause it to under-estimate risk and allow positions it should block.
- Dashboard Portfolio page (page 9) will show misleading risk metrics.

## Files to investigate
- `trading_bot/risk_management.py` — VaR simulation and quantile logic
- `orchestrator.py` — `recompute_var_state` and `configure_market_data_type` calls
- `data/var_state.json` — confirm whether `last_attempt_status: OK` masked a silent calculation error

## Steps to reproduce
1. Check current `data/var_state.json` — if `var_95 == 0.0` with positions open, the bug is still present.
2. Run `recompute_var_state` manually and inspect the raw P&L distribution before quantile selection.

rozavala:
## Summary
Comprehensive safety audit fixing 7 open issues, prioritized by financial risk:

- **#1583 (HIGH)** VaR(95%)=0 compliance bypass: when 5th-percentile P&L is positive but 99th-percentile shows real tail risk ($3,725), compliance gate was wide open. Now promotes var_99 to var_95 so Article V enforcement reflects actual portfolio risk.
- **#1594** Sentinel effectiveness wrongly reported 0 council decisions: Pandas `.dt.strftime()` on tz-aware columns silently mismatched. Replaced with `.dt.normalize()`.
- **#1584** All regimes UNKNOWN: `fundamental_regime.json` doesn't exist for any commodity. Added fallback to council_history.csv `fundamental_regime`/`entry_regime` columns.
- **#1587** Drawdown guard empty error: `str(TimeoutError())` = `""`. Now uses `str(e) or repr(e)` for log + Pushover messages.
- **#1586** Perplexity 429 burst: multiple agents fire queries simultaneously. Added process-wide `asyncio.Semaphore(2)` to serialize concurrent calls.
- **#1589** MicrostructureSentinel post-market errors: `unsubscribe_all()` called `cancelMktData` on disconnected IB. Guarded with `isConnected()`, downgraded to WARNING.
- **#1591** Stale news re-evaluation: 4-month-old RSS entries re-parsed every cycle. Now adds stale links to seen_cache to stop the loop.

## Test plan
- [x] Full test suite: 1121 passed, 0 failures
- [x] Targeted tests for changed modules (var_calculator, compliance, system_digest, sentinels, regime): 228 passed
- [ ] Verify VaR reporting on next trading session — var_95 should now reflect tail risk
- [ ] Verify regime shows actual value (not UNKNOWN) in next system health digest
- [ ] Confirm no Perplexity 429 bursts during multi-commodity cycle
- [ ] Confirm clean post-market logs (no unsubscribe ERROR spam)

Closes #1583, #1584, #1586, #1587, #1589, #1591, #1594.


rozavala:
## Review Feedback Addressed (9f17cca)

### Accepted (2 fixes applied):

**1. VaR falsification — VALID, fixed.** Reverted the `var_95 = var_99` promotion in `var_calculator.py`. The VaR calculator now reports strict mathematical truth. The tail-risk check moved to `compliance.py` Article V: when `var_95_pct == 0.0` but `var_99_pct > var_limit` with positions open, compliance rejects the trade directly. This is architecturally correct — the calculator computes, compliance decides.

**2. `.iloc` vs `.loc` — VALID, fixed.** Changed `ch_df['timestamp'].iloc[recent.index[-1]]` to `ch_df.loc[recent.index[-1], 'timestamp']`. The `.iloc` call used a label as a positional index, which works by accident on freshly-loaded CSVs but would crash on any filtered/sorted DataFrame.

### Rejected (2 points with evidence):

**3. seen_cache "memory leak" — INVALID.** The existing `_load_seen_cache` (line 194) already has **7-day TTL eviction**: `cutoff = time_module.time() - (7 * 24 * 3600)` with entries pruned at `{k: v for k, v in data.items() if v > cutoff}`. Stale links added by the fix get a "now" timestamp via `_save_seen_cache` (line 219: `self._seen_timestamps.get(link, now)`), persist for 7 days (during which the RSS entry is silently skipped), then get pruned. The reviewer's suggestion to "implement a proper time-based TTL eviction" already exists in the code.

**4. asyncio.Semaphore cross-loop — INVALID for Python 3.12.** Event-loop binding for asyncio synchronization primitives was removed in Python 3.10 ([BPO-42392](https://bugs.python.org/issue42392)). Our system runs Python 3.12.3. `asyncio.Semaphore()` is not loop-bound at creation time — it binds at `acquire()`. The `RuntimeError: Task got Future attached to a different loop` is a pre-3.10 issue only. Empirically: **1121 tests passed, 0 failures** across both commits.

### Test results: 1121 passed, 0 failures (7m28s)

rozavala:
## Second Review Feedback Addressed (4092952)

### Accepted (3 fixes applied):

**1. Float epsilon** — Changed `cached.var_95_pct == 0.0` to `< 1e-6` in compliance.py. The `max(negative, 0.0) → round(x, 6)` path produces exactly 0.0 today, but `< 1e-6` guards against simulation noise at zero cost.

**2. Inconsistent return shape** — Added `'source'` key to all `_build_regime_context` return paths (fundamental_regime.json, council_history fallback, UNKNOWN, and error). Prevents `KeyError` if downstream code accesses `context['source']`.

**3. Removed debug log** — Restored the `Skipping stale headline` debug log. Now logged once per link (on first encounter, before adding to seen_cache) to preserve feed health diagnostics without the per-cycle repetition the original fix addressed.

### Acknowledged but not changed:

**4. `entry_regime` semantics** — Valid concern: it's the regime at trade entry, not a live assessment. However:
- It's strictly better than UNKNOWN (the current bug) — provides actual signal
- The `updated_at` timestamp shows when the value was recorded (staleness is visible)
- The `source: council_history.entry_regime` field identifies it as a historical value
- The `fundamental_regime` column is checked first but is always NaN in production data

**5. Drawdown latch** — **No latch exists.** `DrawdownGuard.update_pnl()` recalculates status from scratch on every call (lines 215-222: unconditional threshold check). When IB recovers, the next `guarded_generate_orders` call succeeds and entry is allowed. The fail-closed is per-invocation, not persistent.

**6. Perplexity `max_concurrent` config** — Valid follow-up. Low risk to hardcode at 2 for now.

**7. Microstructure re-subscribe gap** — Pre-existing concern (before this fix, `tickers.clear()` was also called on error path). Out of scope.

### Test results: 1121 passed, 0 failures (7m33s)

Issue #1604, opened 2026-04-14
investigate: ZW adaptive walk produces off-tick price 63.2 despite profile tick_size=0.25


rozavala:
## Summary

The adaptive price walker generated `63.2` as the next step for a ZW straddle (from `63.0`, cap `63.8`). IB rejected it as \"Invalid Price\". `config/profiles/zw.json` declares `tick_size: 0.25`, which means valid prices from 63.0 should be 63.25, 63.50, etc. — not 63.2.

This means `get_tick_size(config)` is returning something other than 0.25 for ZW at runtime, or the step calculation is bypassing the tick rounding.

## Evidence

```
2026-04-14 16:33:49 - OrderManager - INFO - Adaptive Update for Order 38 (ZW-C600.0+P600.0): 63.0 -> 63.2 (Cap/Floor: 63.8)
2026-04-14 16:33:49 - ib_insync.wrapper - ERROR - Error 201, reqId 38: Order rejected - reason:Invalid Price
```

Price 63.2 is not a multiple of 0.25, confirming it violates CBOT ZW option tick rules. Yet `round_to_tick(proposed, 0.25, 'BUY')` should have produced 63.25.

## Investigation needed

1. Add a debug log of the actual TICK_SIZE value used inside monitor_and_execute_orders for ZW
2. Check if get_tick_size(config) reads from the correct profile key at runtime vs what zw.json declares
3. Verify step calculation: with remaining_gap=0.8, target_steps=6, gap_step≈0.133, TICK_SIZE=0.25, step_amount=max(0.133,0.25)=0.25 → 63.0+0.25=63.25. This is correct. So TICK_SIZE must be < 0.133 at runtime.

## Impact

Even after fixing the counter-reset bug (#1602), the walk will generate invalid tick prices on every step, causing repeated rejections until the order expires.

## Related

- #1602 — error counter reset (the flood of 238 rejections)
- #1603 — ZW yfinance wrong suffix

Issue #1603, opened 2026-04-14
fix: ZW yfinance ticker resolves to wrong exchange suffix (ZWN26.NYB instead of ZWN26.CBT)


rozavala:
## Summary

`data_providers.py` constructs `ZWN26.NYB` for ZW (Wheat) contract lookups. NYB is the NYBOT/ICE suffix; ZW is CBOT-listed and should use `.CBT`. Yahoo Finance returns 404 for `ZWN26.NYB`, so all yfinance-based price data for specific ZW contracts fails silently.

## Evidence

Repeated in both dashboard and orchestrator logs today:

```
2026-04-14 15:27:11 - yfinance - ERROR - HTTP Error 404: Quote not found for symbol: ZWN26.NYB
2026-04-14 18:43:00 - yfinance - ERROR - $ZWN26.NYB: possibly delisted; no price data found
2026-04-14 20:46:31 - yfinance - ERROR - $ZWN26.NYB: possibly delisted; no price data found
```

## Root Cause

`trading_bot/data_providers.py` `_resolve_yf_ticker` (line 356):

```python
_EXCHANGE_SUFFIX = {
    'ICE': 'NYB',
    'NYBOT': 'NYB',
    'NYMEX': 'NYM',
    'CME': 'CME',
    'COMEX': 'CMX',
    # CBOT is missing — falls to default 'NYB'
}

suffix = self._EXCHANGE_SUFFIX.get(exchange.upper(), 'NYB')  # default NYB is wrong for CBOT
```

ZW's exchange in `config.json` is `"CBOT"`, which is not in the dict, so it silently falls to `'NYB'`.

## Fix

Add `'CBOT': 'CBT'` to `_EXCHANGE_SUFFIX`. Also consider changing the default fallback from `'NYB'` to raising a warning, since silently using the wrong exchange suffix can mask data quality issues.

## Impact

All yfinance lookups for specific ZW contract months (not the continuous `ZW=F`) return no data. Falls back to Databento or disk cache, but any gap in those sources means no price data for ZW at all.

rozavala:
Fixed in PR #1605 (merged). Added `'CBOT': 'CBT'` to `_EXCHANGE_SUFFIX` in `trading_bot/data_providers.py`. ZW contract lookups now resolve to `ZWN26.CBT` instead of `ZWN26.NYB`. Also added a `logger.warning` for any exchange not in the map so future unmapped exchanges surface immediately rather than silently defaulting to NYB.

PR #1605, opened 2026-04-14
fix: stop infinite order-rejection loop and fix ZW yfinance exchange suffix

### LINKED ISSUE #1602: fix: modification_error_count resets on every adaptive attempt, causing infinite rejection loop
## Summary

A ZW straddle order (reqId 38) received **238 consecutive \"Invalid Price\" rejections** from IB between 16:33 and 17:03 today, only stopping when the 30-minute monitoring timeout expired.

## Root Cause

`trading_bot/order_manager.py` line 2894 resets the error counter to zero on *every* new adaptive price attempt:

```python
details['modification_error_count'] = 0  # Reset on attempt
```

The guard at line 2856 (`if modification_error_count >= max_modification_errors`) is therefore never reached, because the cycle is:

1. Current price 63.0 → compute new price 63.2 → **reset counter to 0** → place modification
2. IB rejects (Error 201): counter increments to 1, ib_insync reverts order back to Submitted at 63.0
3. Next 15-second interval: sees 63.0 again → computes 63.2 again → **resets counter to 0** → repeat
4. Logs show `Error count: 1/3` on every single rejection — it never reaches 3

## Evidence

```
2026-04-14 16:33:49 - OrderManager - WARNING - ⚠️ Error 201 for Order 38: Modification rejected. Reverting internal tracking from 63.00 to 63.20. Error count: 1/3
2026-04-14 16:34:04 - OrderManager - WARNING - ⚠️ Error 201 for Order 38: Modification rejected. Reverting internal tracking from 63.00 to 63.20. Error count: 1/3
... (238 times total, ending at 17:03:22)
2026-04-14 17:03:34 - OrderManager - WARNING - Monitoring loop timed out. Some orders may not be in a terminal state.
```

## Fix

Do not reset `modification_error_count` to 0 before knowing the modification succeeded. Instead, only reset it on a confirmed successful fill or when the price advances past the previously-rejected value. Remove line 2894 (or guard it so it only resets after a fill).

## Impact

Any order whose adaptive walk target is an invalid tick price will be rejected hundreds of times until the monitoring window expires, flooding logs and wasting IB API quota.

rozavala:
## Summary

- **Bug #1602** — `modification_error_count` was reset to 0 on every new adaptive price attempt, and `last_known_good_price` was updated optimistically (before IB confirmed). A single ZW straddle order got rejected 238 times in 30 minutes. Full fix implemented (see Changes).
- **Bug #1603** — `CBOT` was missing from the yfinance exchange-suffix map, causing ZW contract lookups to resolve to `ZWN26.NYB` (invalid) instead of `ZWN26.CBT`.
- **Investigation #1604** — Added a `TICK_SIZE=<value>` log line at the start of the adaptive walk to surface the off-tick ZW price root cause in the next run.

## Changes

**`trading_bot/order_manager.py`**

Two module-level helpers extracted for testability and correctness:

- `_update_tracking_on_acceptance(order_id, confirmed_price, live_orders)` — called from `on_order_status` on every `PreSubmitted`/`Submitted` event. Advances `last_known_good_price` and resets `modification_error_count` **only when the confirmed price is different** from the last known good price (epsilon comparison). Returns `False` and is a no-op when IB sends the post-rejection revert Submitted at the same price — this is the key guard that prevents counter reset.

- `_handle_modification_rejection(req_id, live_orders)` — called from `on_error` on Error 201. Increments counter, snaps `lmtPrice` back to `last_known_good_price`, and freezes the adaptive ceiling when `max_modification_errors` is reached.

Updated `on_order_status` to call `_update_tracking_on_acceptance` so `last_known_good_price` tracks only IB-confirmed prices.

Added `logger.info(f"Adaptive walk TICK_SIZE={TICK_SIZE} for ...")` for #1604 diagnosis.

**`trading_bot/data_providers.py`**
- Added `'CBOT': 'CBT'` to `_EXCHANGE_SUFFIX`.
- Added `logger.warning(...)` for any exchange not in the map.

**`tests/test_order_manager.py`**
- `TestError201Handling` — 7 direct unit tests for both helpers and a combined walk→accept→reject→revert-no-op→freeze scenario.

## Test plan

- [x] `pytest tests/test_order_manager.py::TestError201Handling` — 7 passed
- [x] `pytest tests/` — 1132 passed, 0 failures

## Related issues

Fixes #1602, #1603. Helps diagnose #1604.


rozavala:
## Automated Review: Ready to Merge ✓

**Tests**: Full suite run on this branch — **1,125 passed, 0 failures**.

### Bug #1602 — `modification_error_count` reset (the critical fix)

Root cause is correct. `modification_error_count = 0` inside the adaptive walk loop meant the guard counter was wiped before it could ever reach the threshold (3), allowing rejection loops to run indefinitely. Removing that reset is the right fix — the counter now accumulates across consecutive rejections and the walk will freeze as intended.

### Bug #1603 — CBOT exchange suffix

`'CBOT': 'CBT'` is correct (CME Group Board of Trade uses `.CBT` in yfinance). The warning log for unmapped exchanges is a good defensive addition.

### Diagnostic #1604 — TICK_SIZE log line

Harmless and useful for diagnosing off-tick price issues.

---

### One minor comment inaccuracy (non-blocking)

The new comment at `order_manager.py` ~line 2895 reads:
> _"It tracks the last confirmed-accepted price and is only updated after a fill."_

`last_known_good_price` is never actually updated after a fill — it is set once at order initialization to `order.lmtPrice` and never modified again. The comment should read something like _"It is set to the initial order placement price and never modified."_ This doesn't affect behavior, but will mislead the next reader. Worth a one-line cleanup in a follow-up.

---

**Verdict**: Both bugs are real production issues (238 rejections in 30 minutes), the fixes are minimal and correct, tests are clean. Labeled **Ready to be merged**.

---

PR #1662, opened 2026-04-19
feat: trend-alignment-based holding periods (WITH_TREND 10d / COUNTER_TREND 1d)

rozavala:
## Summary

- Replaces the flat 3-day hold for all positions with a two-regime system validated by simulations across KC, NG, SB, and ZW
- **WITH_TREND** (signal rides the SMA-200 macro trend) → hold up to 10 days, Friday close disabled
- **COUNTER_TREND** (signal fights the macro trend) → hold at most 1 day, Friday close applies
- Unknown/legacy positions fall back to the existing `max_holding_days` (3d) + Friday close behavior

## Changes

| File | Change |
|---|---|
| `trading_bot/utils.py` | `compute_trend_alignment()` helper; `trend_alignment` column in trade_ledger |
| `trading_bot/signal_generator.py` | Computes trend_alignment at decision time; propagates to council log, decision_signals, and signal dict |
| `trading_bot/decision_signals.py` / `schema.py` | v10 `trend_alignment` column in both CSV schemas |
| `trading_bot/order_manager.py` | Fill logger forwards `trend_alignment` to ledger; `close_stale_positions` applies per-position hold days and Friday close exemption |
| `trading_bot/exit_reasons.py` | `WITH_TREND_MAX_HOLD` / `COUNTER_TREND_MAX_HOLD` constants |
| `config.json` | `with_trend_max_holding_days=10`, `counter_trend_max_holding_days=1` |
| `scripts/migrations/backfill_trend_alignment.py` | Idempotent backfill for pre-existing CSV rows |
| `tests/test_trend_alignment_hold.py` | 33 tests covering all alignment/hold/Friday-close combinations |

## Test plan
- [ ] `pytest tests/test_trend_alignment_hold.py` — 33/33 pass
- [ ] `pytest tests/` — 1251/1251 pass (no regressions)
- [ ] Run backfill migration on production data dirs before deploying: `python scripts/migrations/backfill_trend_alignment.py data/KC data/NG data/SB data/ZW`
- [ ] Verify `trend_alignment` column appears in `trade_ledger.csv` and `decision_signals.csv` on next signal cycle

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1666, opened 2026-04-20
feat: trend-alignment follow-ups (stacked on #1662)

rozavala:
Stacked on #1662. Six strategic follow-ups so the doubled WITH_TREND holding period resolves correctly through Brier scoring and is observable end-to-end.

## Summary
- **Brier orphan window 168h → 336h** (`enhanced_brier.py`, `brier_bridge.py`): 10-day WITH_TREND holds now resolve before auto-orphaning.
- **Brier half-life 14d → 28d** (`brier_scoring.py`): reliability decay tracks the doubled resolution latency so long-hold agents don't drift.
- **Per-cycle trend_alignment log** (`signal_generator.py`): structured counts of WITH_TREND / COUNTER_TREND / UNKNOWN per cycle.
- **Scorecard breakdown** (`pages/2_The_Scorecard.py`): new Trend Alignment Performance section with trades, win-rate, avg/total P&L. Small-sample caption at n<15; warns if WITH_TREND underperforms COUNTER_TREND by >5pp.
- **Wiring tests** (`tests/test_trend_alignment_hold.py`): guard `close_stale_positions → _classify_lot_for_close` delegation, the `trend_alignment` column name, and the DataFrame `to_dict('records')` round-trip. Clarified the mixed-alignment test docstring (helper vs position-level semantics).
- **Docs + cleanup**: CLAUDE.md watch-out bullet; moved ad-hoc `simulate_holding_period.py` into `backtesting/` with a module docstring; deleted scratch `pr1662.txt`.

## Test plan
- [x] `pytest tests/` — 1262 passed in ~8min
- [x] New wiring tests fail if delegation is removed or the column is renamed (verified via source inspection)
- [ ] Monitor first cycle after deploy for the new trend_alignment counts log line
- [ ] Confirm Scorecard page renders the new section without layout regressions
- [ ] Spot-check `enhanced_brier.json` 14 days after deploy — WITH_TREND predictions should resolve rather than orphan


rozavala:
## Review: Ready to merge (second commit resolves all three prior concerns)

The second commit (59b69f8d) addressed the three issues from the original review:

**1. Orphan window divergence — FIXED**
`enhanced_brier.py` now has `_compute_orphan_window_hours()` using `max_hold * 24 * 1.4` (cap 504h), and `auto_orphan_stale_predictions` / `auto_orphan_enhanced_brier` both default to `None` which triggers this helper. The formula is a flat 1.4× uplift vs `brier_scoring`'s per-prediction calendar-aware walk, but both target the same window and share the 504h cap — close enough for operational purposes.

**2. Fragile `inspect.getsource` test — FIXED**
The column-name guard was dropped (the FIFO-walk integration test already catches column renames end-to-end). The delegation guard was kept with justification — it's the only test that would catch `_classify_lot_for_close` being re-inlined.

**3. Half-life 14d → 28d global change — DOCUMENTED**
The brier_scoring docstring now explicitly states: *"Known trade-off: 28d also slows decay for COUNTER_TREND predictions (which resolve in 1 day)... The principled fix is per-alignment half-life — tracked as a future iteration. The current uniform 28d is acceptable because decay preserves the RELATIVE ordering across agents; only absolute drift speed changes."*

This is an honest accounting of the trade-off. The relative ordering preservation argument is valid for the current use case (agent weighting in voting). Per-alignment half-life remains the right long-term improvement.

Other components all clean: per-cycle trend_alignment log, Scorecard section, backtesting simulator, CLAUDE.md watchout. 1262 tests passing.

PR #1676, opened 2026-04-20
fix(deferred-triggers): read-then-ack so Gateway outages don't lose triggers (#1671)

### LINKED ISSUE #1671: bug: deferred-trigger read-and-clear loses triggers on connection failure
## Summary

`StateManager.get_deferred_triggers` **atomically reads AND clears** the deferred trigger file under a single lock. If the subsequent code path fails before successfully placing orders, **the triggers are silently lost**. On 2026-04-20, this cost us 30 deferred triggers across two engines.

## Evidence from 2026-04-20 outage

**NG engine — 17 triggers lost:**

```
orchestrator_ng.log:374  2026-04-20 13:00:37,266 - Orchestrator - INFO - SENTINEL SAFETY NET: First ACTIVE transition — processing deferred triggers
orchestrator_ng.log:376  2026-04-20 13:00:37,274 - Orchestrator - INFO - Processing 17 deferred triggers from overnight
orchestrator_ng.log:380  2026-04-20 13:00:52,787 - Orchestrator - ERROR - Failed to process deferred triggers:
```

**ZW engine — 13 triggers lost:**

```
orchestrator_multi.log:2883  2026-04-20 13:30:44,892 - Orchestrator - INFO - Processing 13 deferred triggers from overnight
orchestrator_multi.log:2887  2026-04-20 13:31:00,404 - Orchestrator - ERROR - Failed to process deferred triggers:
```

Both failures were caused by an upstream IB Gateway outage (see #1669), but the mechanism matters: **by the time the exception fires in `process_deferred_triggers`, the triggers have already been cleared from disk.** There is no retry, no re-queue, no dead-letter — they are gone.

### `trading_bot/state_manager.py:455–495`

```python
def _do_get_and_clear():
    triggers = cls._load_deferred_triggers()
    if not triggers:
        return []
    # Clear immediately under the same lock
    cls._write_deferred_triggers([])   # <-- line 462: destructive clear
    ...
    return valid_triggers
```

The lock correctly prevents two callers from racing, but it **couples "read" and "commit-as-done" into one atomic step**. The contract is effectively: *"If you take these triggers, you must succeed — there is no way to put them back."*

### `orchestrator.py:3870–3944` — `process_deferred_triggers`

```python
deferred = StateManager.get_deferred_triggers()   # line 3879 — ALREADY CLEARED
if not deferred:
    return
logger.info(f"Processing {len(deferred)} deferred triggers from overnight")
ib_conn = await IBConnectionPool.get_connection("deferred", config)   # line 3885 — can raise TimeoutError
# ...
except Exception as e:
    logger.error(f"Failed to process deferred triggers: {e}")   # line 3938 — no re-queue
```

Any failure between line 3879 (clear) and the end of the loop (line 3935) drops the in-memory list on the floor. Connection timeout is the most common trigger, but any exception mid-loop has the same effect (partial processing → remaining triggers lost).

## Why this is critical

- **Real data loss on every Gateway blip that coincides with a deferred-processing window** (market open, first ACTIVE transition, overnight catch-up)
- Deferred triggers represent **overnight sentinel events** that accumulated while the market was closed — prediction market moves, weather events, macro contagion alerts, etc. Losing them means trading the next session blind to what happened overnight
- Silent failure — no alert fires, no reconciliation catches it. The only evidence is the ERROR log line, which may be overlooked during busy incidents

## Proposed fix (two-phase commit)

Replace "read-and-clear" with "read + ack-on-success":

### `state_manager.py`

```python
@classmethod
def get_deferred_triggers(cls, max_age_hours: float = 72) -> list:
    \"\"\"Read triggers without clearing. Caller must call ack_deferred_triggers on success.\"\"\"
    def _do_read():
        triggers = cls._load_deferred_triggers()
        # ... age filtering (unchanged) ...
        return valid_triggers
    return cls._with_deferred_lock(_do_read)

@classmethod
def ack_deferred_triggers(cls, processed_ids: list[str]) -> None:
    \"\"\"Remove successfully-processed triggers by id. Unprocessed triggers remain.\"\"\"
    def _do_ack():
        remaining = [t for t in cls._load_deferred_triggers() if t.get('id') not in processed_ids]
        cls._write_deferred_triggers(remaining)
    cls._with_deferred_lock(_do_ack)
```

Each trigger needs a stable `id` (hash of source + reason + timestamp, or a UUID assigned at deferral time).

### `orchestrator.py:3870` — `process_deferred_triggers`

```python
deferred = StateManager.get_deferred_triggers()
# ...open connection, fail early if no Gateway...
ib_conn = await IBConnectionPool.get_connection(\"deferred\", config)

processed_ids = []
try:
    for t in deferred:
        try:
            if _get_deduplicator().should_process_historical(trigger):
                await run_emergency_cycle(trigger, config, ib_conn)
            processed_ids.append(t['id'])  # ack even skip/reject — they've been *decided*
        except Exception as per_trigger_err:
            logger.error(f\"Deferred trigger {t.get('id')} failed: {per_trigger_err}\")
            # do NOT ack — leave in queue for next attempt
finally:
    if processed_ids:
        StateManager.ack_deferred_triggers(processed_ids)
```

Key properties:
- Connection failure at open time → no triggers cleared → retried on next `process_deferred_triggers` invocation
- Mid-loop failure → already-decided triggers acked, remaining left in queue
- Expired triggers are still filtered out (age check happens at read time), but we can ack them too so they stop accumulating

## Edge cases / bounded retry

To prevent a poison trigger from infinitely retrying:

```python
t['attempts'] = t.get('attempts', 0) + 1
if t['attempts'] > MAX_ATTEMPTS:  # e.g., 3
    logger.error(f\"Trigger {t.get('id')} exceeded max attempts, discarding\")
    processed_ids.append(t['id'])  # ack to remove
    continue
```

## Acceptance criteria

- [ ] `get_deferred_triggers` no longer clears the file
- [ ] New `ack_deferred_triggers(ids)` method removes triggers by id
- [ ] Every trigger has a stable `id` (backfilled on read if missing)
- [ ] `process_deferred_triggers` acks on success (per-trigger) and leaves unprocessed triggers in the queue
- [ ] Connection-failure at open time → zero triggers cleared (observable in test)
- [ ] Mid-loop failure → only successful triggers acked
- [ ] Max-attempts cap prevents poison triggers from blocking the queue
- [ ] Unit tests:
  - `test_deferred_preserved_on_connection_failure`
  - `test_deferred_partial_ack_on_mid_loop_failure`
  - `test_deferred_max_attempts_discards_poison`
- [ ] Integration test: simulate Gateway timeout during `process_deferred_triggers`; verify triggers survive

## Migration

- New `id` and `attempts` fields are additive to the JSON schema — missing fields are tolerated via `.get(...)` defaults
- No data migration required; existing on-disk triggers get an id assigned on first read
- Backwards compatible if rolled back (fields ignored)

## Risks

| Risk | Mitigation |
|---|---|
| Triggers retry forever if ack logic has a bug | Max-attempts cap (hard ceiling of 3); dead-letter log line |
| Race between read and ack if another process starts | `_with_deferred_lock` serializes (existing mechanism) |
| Duplicate processing if crash happens between `run_emergency_cycle` and ack | Existing dedup in `should_process_historical` already handles this |
| Expired triggers accumulate in queue | Acked alongside processed ones; reader still filters them |

## Related

- #1669 — Phase 1 connection pool consolidation (reduces the failure surface that triggers this bug, but doesn't fix the data-loss mechanism)
- Contributing factor in today's (2026-04-20) data loss; should ship independently of #1669

rozavala:
## Summary
- Closes #1671 (critical).
- `StateManager.get_deferred_triggers` no longer clears the file; callers now explicitly `ack_deferred_triggers(ids)` after each trigger is decided.
- `process_deferred_triggers` bumps per-trigger `attempts` up-front so connection failures still count toward the poison cap (default 3), then acks per-trigger inside its own try/except. Anything undecided survives for the next invocation.
- Legacy on-disk triggers get a stable uuid backfilled on first read so they become acknowledgeable.

## Why
On 2026-04-20, IB Gateway dropped for ~90 min. The old read-and-clear meant 30 overnight triggers (17 NG + 13 ZW) were wiped from disk before the first successful trade could be placed. With read-then-ack those triggers would have survived the outage.

## Test plan
- [x] `pytest tests/test_deferred_triggers.py` — 15/15 pass (added: preserved-on-connection-failure, partial-ack, poison cap, legacy backfill, id/attempts on queue, empty-ack noop).
- [x] `pytest tests/test_sentinel_loop.py tests/test_session_schedule.py` — 26/26 pass.
- [ ] DEV deploy: verify queue length goes to zero after a successful run and recovers after a forced Gateway restart.


rozavala (commented):
**Code review + test run: ✅ All 43 tests pass** (17 in `test_deferred_triggers.py`, 26 across `test_sentinel_loop` + `test_session_schedule`). **Ready to merge.**

The read-then-ack pattern is correct and well-implemented:

- `get_deferred_triggers` now reads without clearing — triggers survive Gateway connection-level failures intact
- `ack_deferred_triggers(ids)` provides precise per-trigger removal; partial acks correctly leave unprocessed triggers on disk
- `bump_deferred_attempts` counts only _post-connection_ processing failures toward the poison cap — not Gateway outages — so N cycles of connection failure will never poison triggers that were never actually tried
- `DEFERRED_MAX_ATTEMPTS = 3` as a monkeypatchable class attribute is a clean testability decision
- Legacy on-disk triggers are backfilled with stable UUIDs on first read, making the transition seamless
- The `finally` block correctly acks decided triggers and bumps failed ones even when the per-trigger loop raises, preventing double-processing

Minor: the PR description claims "15/15 new tests" but `test_deferred_triggers.py` has 17 — the two async integration tests (`test_connection_outage_does_not_bump_attempts`, `test_per_trigger_failure_bumps_and_eventually_poisons`) were added after the description was written. The extra coverage is welcome.

---

Issue #1637, opened 2026-04-16
refactor: connection_pool has no portfolio-level cross-purpose outage detection


rozavala:
## Summary

`IBConnectionPool` uses per-purpose `_consecutive_failures` counters. When the IB Gateway goes fully down, each of the ~15 active purposes (KC_sentinel, KC_drawdown_check, SB_sentinel, ZW_orders, etc.) accumulates its own counter — but since each purpose is called at different cadences, no individual purpose is likely to hit `FORCE_RESET_THRESHOLD=5` before the Gateway recovers or before a 1102 resets the counter.

During the Apr 16 outage (16 connection errors, 12:15–14:28 UTC), zero ZOMBIE alerts fired. The system saw cascading failures across all commodities + purposes, but with ~15 separate counters none crossed the threshold.

## What's Missing

A portfolio-level view: if **any N out of M purposes** fail within a time window, fire a single "IB Gateway appears fully down" alert.

## Proposed Detection Logic

```python
# In connection_pool.py — class-level tracking
_all_failures_window: list[float] = []  # timestamps of all purpose failures
PORTFOLIO_OUTAGE_WINDOW = 300  # seconds
PORTFOLIO_OUTAGE_MIN_FAILURES = 5   # distinct purposes failing in window

@classmethod
def _check_portfolio_outage(cls, config):
    import time
    now = time.time()
    cls._all_failures_window = [t for t in cls._all_failures_window if now - t < cls.PORTFOLIO_OUTAGE_WINDOW]
    cls._all_failures_window.append(now)
    if len(cls._all_failures_window) >= cls.PORTFOLIO_OUTAGE_MIN_FAILURES:
        # deduplicate: fire alert only once per window
        ...
```

Call `_check_portfolio_outage` at the end of every failed `get_connection`.

## Impact

- Apr 16 outage: 16 failures in 133 minutes → would have triggered portfolio alert in first 5 failures (~first 30 min)
- Single-purpose failures (normal): 1–2 failures per 5-minute window → below threshold, no false alarm

## Acceptance Criteria

- [ ] Portfolio-level alert fires when ≥ 5 distinct connection failures occur within 5 minutes
- [ ] Alert fires at most once per outage window (debounce)
- [ ] Single-purpose transient failures do not trigger portfolio alert
- [ ] Tests cover the multi-purpose failure scenario
- [ ] Integrates cleanly with fix for #1636 (flapping reset issue)

## Related

[...]

PR #1679, opened 2026-04-20
fix(order_manager): drop empty missed_orders rows and escalate via Pushover (#1675)

### LINKED ISSUE #1675: bug: missed_orders.json is a write-only orphan with stale/empty rows and no consumer
## Summary

`data/missed_orders.json` is a **write-only orphan**: `_persist_missed_orders` (`trading_bot/order_manager.py:2322`) writes to it when IB connection fails during order placement, but **nothing consumes it** — no dashboard page, no alert, no replay job, no cleanup. Entries from 2026-04-15 / 2026-04-16 are still sitting in the file untouched. The comment on the write path (*"review manually; orders are stale after 30 min"*) is aspirational: the staleness is never enforced and no operator-facing surface exists.

Separately, 3 of the 4 current entries have empty `contract_local_symbol` and empty `signal` dict, suggesting the writer can still produce useless rows under edge cases even after the recent BAG-combo fallback fix.

## Evidence (active state)

```
$ stat data/missed_orders.json
Modify: 2026-04-16 16:20:43 (+93 hours ago, no writes since)

$ cat data/missed_orders.json  (4 entries)
  2026-04-15T06:06:21Z  contract=""        signal={}
  2026-04-16T05:46:13Z  contract=""        signal={}
  2026-04-16T05:48:28Z  contract=""        signal={}
  2026-04-16T16:20:43Z  contract="BAG:KC"  signal={}
```

```
$ grep "Persisted.*missed order\|unexpected item type" logs/orchestrator_*.log
(no matches today — but we have 4 unreviewed entries sitting in the file)
```

```
$ grep -rn "missed_orders" pages/ dashboard*.py
(no hits — no UI consumer)
```

## Failure modes this creates

1. **Silent pileup** — next production outage (cf. today's 12:45 UTC Gateway outage) could write dozens of rows here, and nobody would know until someone inspects the file.
2. **Empty rows are indistinguishable from missing data** — `{"contract_local_symbol":"","signal":{}}` gives an operator no way to reconstruct what was supposed to trade. Even the `missed_at` timestamp without context is useless.
3. **No replay path** — per the docstring, a compliance-approved signal should not be silently lost. But writing it to a file nobody reads is silent loss with extra steps.
4. **No TTL** — the comment says 30-min staleness but 93-hour-old entries remain. Anyone looking would be tempted to replay a stale trade.

## Root cause of empty rows (suspected)

In `_persist_missed_orders`:

```python
# order_manager.py:2344-2349
for item in target_queue:
    try:
        contract, order, signal = item
    except (ValueError, TypeError):
        logger.warning(f"_persist_missed_orders: unexpected item type {type(item)!r}, ...")
        contract, order, signal = None, None, item
```

When `signal` is not a dict (e.g., when the fallback path sets `signal = item` and `item` is some non-dict), the downstream filter yields empty `signal_scalars = {}`. Combined with `contract = None` (empty localSymbol/secType/symbol), you get `contract_local_symbol=""` + `signal={}`.

The Apr 15 entries predate the `_dropped_nested_keys` / `unexpected item type` instrumentation (added around the 2026-04-16 fix). The Apr 16 16:20 entry has `"BAG:KC"` from the secType-fallback, so by that point the writer had been partially hardened. But the empty-signal case still slipped through — the writer didn't refuse to persist a row with no reconstructable content.

## Proposed fix

Three small, independent changes:

### 1. Refuse empty rows

```python
# order_manager.py _persist_missed_orders
if not local_sym and not signal_scalars:
    logger.error(
        f"_persist_missed_orders: refusing to persist empty row "
        f"(contract={contract!r}, signal={signal!r}) — raising alert instead"
    )
    send_pushover_notification(
        config.get('notifications', {}),
        "Missed Order — Unrecoverable",
        f"Order generation produced a signal with no contract and no content. "
        f"Raw: {str(item)[:500]}"
    )
    continue  # don't write the row
```

Rationale: a row we can't replay is worse than no row — it takes up review space and implies something can be done.

### 2. Add TTL on read and a scheduled cleanup

Add a helper `_load_missed_orders(max_age_minutes=30)` that filters out stale rows when reading. Call it from the dashboard consumer (below). Separately, add a light cleanup task to the orchestrator schedule that rewrites the file without stale rows daily.

### 3. Add a dashboard panel

Add a small section to `pages/1_Cockpit.py` (or a new `pages/10_MissedOrders.py`) that:
- Lists recent entries with age, contract, and signal summary
- Flags rows older than 30 min in red
- Provides a manual "dismiss" button that removes the row
- Shows count in the cockpit health header

Without this consumer the file is pure noise.

### 4. Notify on every write

Each `_persist_missed_orders` call that produces ≥1 valid row should send a Pushover notification (the code already does this for the connection-failed message — make sure it mentions the row count and the local symbol of the first entry). That pulls the signal into the place operators actually watch.

## Acceptance criteria

- [ ] Empty rows refused at write time; Pushover alert fired instead
- [ ] `_load_missed_orders(max_age_minutes=30)` helper added, used by UI
- [ ] Cockpit panel (or new page) renders current missed orders with age and replay-readiness
- [ ] Scheduled cleanup task registered via `build_schedule` (daily at 00:05 UTC)
- [ ] Pushover notification on every write includes row count + first local symbol
- [ ] New test: `test_persist_missed_orders_refuses_empty_row`
- [ ] New test: `test_load_missed_orders_filters_stale`
- [ ] Today's 4 stale entries cleared (one-time migration)

## Risks

| Risk | Mitigation |
|---|---|
| Refusing empty rows could swallow a real missed trade with weird metadata | The alert on refusal is the safety net; we don't lose the signal, we escalate it |
| Dashboard panel adds cognitive load | It's a count; zero is the norm. Non-zero is the signal operators actually want |
| TTL deletes a row during active remediation | Keep the daily cleanup; 30-min TTL is for *display*, not deletion |

## Non-goals

- No auto-replay of missed orders (too risky; pricing and IV have moved)
- No change to when `_persist_missed_orders` is called — the write trigger is fine; the read/consumption side is what's broken
- No change to the underlying IB-retry logic (that's covered by #1669 Phase 1)

## Related

- #1534 — original PR that added `_persist_missed_orders` (write-only)
- #1669 — connection pool consolidation (reduces the frequency of IB-connect failures that trigger this path)
- #1671 — deferred-trigger read-and-clear (different code path; same shape of "silent data loss")

rozavala:
## Summary
Closes #1675.

`missed_orders.json` had become a write-only orphan: `_persist_missed_orders` would append rows even when the contract symbol and the reconstructable signal fields were both empty — rows that cannot be replayed or even investigated. No consumer cleans them up, so they just dilute the file.

- `_persist_missed_orders` now refuses a row only when it carries neither a contract nor any scalar signal content. Each refusal fires a `Missed Order — Unrecoverable` Pushover so the operator still learns something was lost.
- A successful persist fires a `Missed Orders Persisted (N)` Pushover with the first symbol and a reminder that rows go stale after 30 min.
- The file is only written when at least one row survived (empty-queue calls no longer touch the file).
- Returns the persisted count (backward-compatible — callers ignored the old `None`).
- New `_load_missed_orders(max_age_minutes)` helper centralizes the freshness filter so future consumers don't have to re-parse the file themselves.

## Test plan
- [x] `pytest tests/test_order_manager.py` — 50 passed (10 new, 40 existing, no regressions)
  - `TestPersistMissedOrdersEmptyRowGuard` (4): empty row refused + Pushover, non-dict signal refused, mixed queue writes only good row, persisted alert fires.
  - `TestLoadMissedOrdersTtl` (3): missing file returns `[]`, TTL keeps fresh / drops stale, unparseable timestamp skipped under TTL.


rozavala:
Addressed the reviewer's "alert-storm during sustained outage" concern in f510f24.

**What changed**
- New module-level dedupe helper `_notify_with_cooldown(kind, notif_config, title, message, cooldown_seconds)` in `trading_bot/order_manager.py`. Tracks last-sent timestamp and suppressed count per notification kind; emits a `"[N similar alert(s) suppressed during Xmin cooldown]"` prefix on the next permitted send so nothing silently disappears.
- Both Pushover sites in `_persist_missed_orders` now go through the helper:
  - `kind="missed_order_unrecoverable"` (empty row refusal)
  - `kind="missed_orders_persisted"` (successful persist)
- Cooldown is config-driven: `notifications.missed_order_alert_cooldown_seconds`, default 1800s (30 min). Setting to 0 disables dedupe (operator override, covered by test).

**Pattern parity**
Modeled after `connection_pool.CONNECTIVITY_ALERT_COOLDOWN` / `PORTFOLIO_ALERT_COOLDOWN` (#1637) — same dict-keyed-by-key shape, same "timestamp + cooldown window" check. Kept local to `order_manager` for now; if a second caller materializes (e.g. #1676 deferred-trigger alerts), we can hoist into `notifications.py`.

**Tests** — 5 new cases in `TestNotifyWithCooldown`:
- `test_first_call_sends_repeat_suppressed` — happy path
- `test_suppressed_count_reported_on_next_send` — backlog surfaces on the next allowed send
- `test_different_kinds_are_independent` — distinct kinds don't share a bucket
- `test_persist_missed_orders_dedupes_across_cycles` — integration: 5 cycles → 1 Pushover (the flagged scenario)
- `test_cooldown_zero_disables_dedupe` — operator escape hatch

Existing `TestPersistMissedOrdersEmptyRowGuard` got a `setUp` that clears the registry so prior-test cooldowns can't bleed across cases. Full `tests/test_order_manager.py` is green (55 passed).

rozavala (commented):
**Code review + test run: ✅ All 55 tests pass** (12 new, 43 existing — no regressions). **Ready to merge.**

The changes are solid and address a real operational gap:

- The empty-row guard (`not local_sym and not signal_scalars`) correctly targets unreplayable rows — a row with no contract and no signal content is worse than no row, as it wastes review space and obscures queue depth
- `_notify_with_cooldown` mirrors the existing `CONNECTIVITY_ALERT_COOLDOWN` pattern from `connection_pool.py` — consistent with the codebase conventions
- The 30-minute default cooldown prevents Pushover flooding during extended outages while ensuring the first event always gets through
- Suppressed-count reporting (`"N similar alert(s) suppressed during Xmin cooldown"`) ensures nothing disappears silently — operators know how many alerts were collapsed
- `_load_missed_orders(max_age_minutes)` centralizes freshness filtering so future consumers don't re-parse the file themselves
- Return value change `None → int` is backward-compatible (all existing callers ignored the return value)

The "50 passed / 10 new" claim is slightly off — actual result is 55 passed / 12 new. No impact on correctness.

---

Issue #1672, opened 2026-04-20
bug: drawdown grace window is outage-unaware, keeps trading armed during sustained Gateway outages


rozavala:
## Summary

The drawdown-guard fail-closed logic has a 2-hour grace window that lets signal generation proceed using **cached** drawdown state when the IB Gateway is temporarily unreachable. This is correct for 30-second blips, but during today's ~80-minute sustained outage it kept `signal_open` armed on stale data until the cached state itself aged out — then fail-closed fired mid-outage.

The grace window should be **outage-aware**: if the portfolio-wide outage detector has fired recently, shorten the grace aggressively (or fail-closed immediately). A brief connection blip is not the same shape as a full Gateway outage.

## Evidence from 2026-04-20

The portfolio outage detector first fired at **13:34:47**:

```
orchestrator_multi.log:2908  2026-04-20 13:34:47,488 - CRITICAL - PORTFOLIO OUTAGE DETECTED: 5 IB connection failures across all purposes in the last 5 min.
```

At **14:06:00**, NG engine's `signal_open` scheduled task tried to run. The drawdown check attempted a fresh connection, timed out after 15s, and because the cached state was already 22+ minutes old (previous update at 10:06 via `KC_drawdown_check`), it blew through the grace-window fallback and fail-closed:

```
orchestrator_ng.log:413  2026-04-20 14:06:15,916 - Orchestrator - ERROR - Drawdown guard check failed (fail-closed): TimeoutError()
```

**What should have happened**: At 13:34, when the portfolio outage was detected, the drawdown guard should have *either* immediately entered fail-closed mode *or* blocked entry to signal generation pre-emptively. Instead, the grace window kept NG's trading intent armed for 30 more minutes, then fail-closed kicked in mid-signal-cycle — the worst of both worlds (stale state for 30min, then crash at critical moment).

### `orchestrator.py:6356–6396`

```python
except Exception as e:
    # Grace window: if cached drawdown state is recent enough, allow entry
    # instead of fail-closed. Drawdown doesn't change dramatically within
    # the signal cycle interval (2 hours). Default grace = 7200s
    grace_seconds = config.get('drawdown_circuit_breaker', {}).get('grace_window_seconds', 7200)
    last_updated = _dg.state.get('last_updated')
    if last_updated and _dg.is_entry_allowed():
        age = (datetime.now(timezone.utc) - datetime.fromisoformat(last_updated)).total_seconds()
        if age <= grace_seconds:
[...]

PR #1680, opened 2026-04-20
feat(scheduler): per-engine deterministic tick jitter (#1674)

### LINKED ISSUE #1674: refactor: jitter scheduled-task tick times per engine to break cross-engine :00 collisions
## Summary

Every scheduled task fires at exactly `HH:MM:00` across all 4 commodity engines. This creates a thundering herd at round minutes and was a direct contributor to the 2026-04-20 12:45:00 UTC Gateway outage (3 handshakes in the same second overwhelmed IB's API handshake thread). **#1669 Phase 1** reduces *how many* sockets can handshake per engine; this issue addresses *when* those handshakes land, so the two together close the thundering-herd hole.

Called out as a deferred complement in #1669 ("Staggered scheduled ticks with jitter — complementary fix, separate ticket"). Filing it now.

## Evidence (active logs only, 2026-04-20)

All engines schedule tasks on minute boundaries. Sample of "Running scheduled task" timestamps (across all engines, same calendar second):

```
4x collisions today, >=2 engines firing within the same second:
  12:15:00   (2 engines)
  12:45:00   (2 engines) — start of 93-min Gateway outage
  16:55:00   (2 engines)
  17:05:00   (2 engines)
```

Every per-engine fire timestamp was `HH:MM:00.XXX` — never `:02`, `:15`, `:47`, etc. Sample:

```
17:25:00,405 | kc  ← same second
17:25:00,248 | sb  ←
17:05:00,200 | kc  ← same second
17:05:00,248 | sb  ←
12:45:00,xxx | ng cleanup     \
12:45:00,xxx | zw start_monitor > three handshakes, one second
12:45:00,xxx | zw deferred    /
```

## Root cause

`build_schedule` in `trading_bot/` (called from `commodity_engine.py:221`) returns task times at minute granularity (`HH:MM`). The scheduler loop wakes at each due time and fires `next_task.function()` immediately at `:00.0xx`. No jitter is applied per-engine or per-task.

Result: whenever two engines share a configured slot (or adjacent cleanup/audit slots collide across engines), the handshake thread gets >=2 concurrent `connect()` calls. IB Gateway's API handshake thread is **single-threaded** — 3 concurrent connects = 2 queued behind a potential 15s timeout.

## Why Phase 1 alone doesn't close this

#1669 consolidates each engine's N purposes into 2 sockets (`reader` + `orders`). That cuts sockets per engine from ~7 to 2 and prevents *within-engine* handshake storms. But **across engines**, the :00-tick alignment is unchanged:

- Today: 4 engines × 2 scheduled tasks near :00 = up to 8 concurrent handshakes
- After Phase 1: 4 engines × 2 sockets, same 8 ticks aligned = still up to 8 concurrent handshakes
- After Phase 2 (#1670) adds 5 more commodities: 9 engines × 2 = up to 18 concurrent handshakes at :00

The tick alignment scales linearly with engine count; connection consolidation doesn't help.

## Proposed fix

Per-engine deterministic jitter applied in `apply_schedule_offset` (the same function that already handles DEV offset). For each engine, derive a stable offset in `[0, 45]` seconds from `hash(ticker) % 46`:

```python
# trading_bot/schedule_builder.py (approx)
JITTER_WINDOW_SECONDS = 45

def apply_engine_jitter(task_list, ticker: str):
    """Shift all scheduled times by a deterministic per-engine offset
    so cross-engine collisions at :00 no longer pile up on IB's
    handshake thread. Offset is stable across restarts."""
    offset = hash(ticker) % JITTER_WINDOW_SECONDS
    return [task.shift_seconds(offset) for task in task_list]
```

Call sites in `commodity_engine.py:244`:

```python
task_list = apply_schedule_offset(task_list, offset_minutes=schedule_offset_minutes)
task_list = apply_engine_jitter(task_list, ticker=self.ticker)  # NEW
```

### Why deterministic, not random

- **Deterministic** = reproducible scheduler logs, predictable post-mortem analysis, dashboard `active_schedule.json` still meaningful
- **Per-engine** (not per-task) = each engine's internal task ordering preserved; only cross-engine boundaries shift
- **45s window** = below the 1-minute task granularity so tasks don't cross into the next slot; comfortably above IB Gateway's ~1s handshake serialization

### Expected outcome

With 4 engines and a 45s window, expected max concurrent handshakes at any second drops from 4 (today) to 1–2 (if two engines happen to hash near each other). After Phase 2 with 9 engines, expected max drops from 9 → 2–3 (birthday-collision math).

## Acceptance criteria

- [ ] `apply_engine_jitter` implemented with deterministic hash-based offset
- [ ] Wired into `commodity_engine._setup` after `apply_schedule_offset`
- [ ] `_write_active_schedule` reflects the jittered times so the dashboard matches reality
- [ ] Test: `test_engine_jitter_is_deterministic` — same ticker yields same offset across runs
- [ ] Test: `test_engine_jitter_different_tickers_differ` — no two current tickers (KC/NG/ZW/SB) collide at 0 offset
- [ ] After DEV deploy for 3 trading days: zero `:XX:00.XXX` collisions across engines in `logs/orchestrator_*.log`
- [ ] After PROD deploy: no repeat of 3-simultaneous-handshakes pattern seen on 2026-04-20 12:45:00

## Risks

| Risk | Mitigation |
|---|---|
| Jitter shifts a task out of its market-open window | Window is ≤45s; all production tasks have ≥1min margin from open/close |
| Jitter changes performance baseline for backtests | Jitter lives in runtime scheduler, not signal/decision code; no effect on backtest determinism |
| Hash collisions on future ticker additions | Easy to detect in CI with the "different tickers differ" test, fail-fast if a new ticker collides |
| Someone reads dashboard and expects :00 ticks | Dashboard reads `active_schedule.json` which includes the jittered times — UI stays consistent |

## Non-goals

- No change to the scheduler loop itself; jitter is applied at task build time
- No global rate limiter for handshakes (separate and more complex lever; not needed once jitter lands)
- No change to deferred-trigger scheduling; those are event-driven, not time-driven

## Related

- #1669 — Phase 1 connection pool consolidation (per-engine reader/orders keys)
- #1670 — Phase 2 fixed-pool architecture (scales better once engine count grows)
- #1671 — deferred-trigger persistence fix (orthogonal; different code path)

rozavala:
## Summary
Closes #1674.

Every engine's scheduled tasks fire at exactly `HH:MM:00`. When 2+ engines share a slot — signals at 13:00, cleanup/start-monitor pairs, etc. — IB Gateway's single-threaded handshake queue takes concurrent `connect()`s in the same calendar second. This was the direct contributor to the 2026-04-20 12:45 UTC outage that #1672 already hardened the drawdown guard against.

- `engine_jitter_seconds(ticker, window=45)` returns a stable per-ticker offset in `[0, 45)` using `hashlib.blake2b` — intentionally **not** Python's salted `hash()`, so the offset survives process restarts (reproducibility is a hard requirement for post-mortem timestamps and the dashboard).
- `apply_engine_jitter(task_list, ticker, window)` shifts every `ScheduledTask` uniformly, so intra-engine ordering is preserved and only the cross-engine boundaries fan out.
- `get_next_task` and missed-task detection now honor `time_et.second` instead of truncating to `:00` (otherwise the scheduler loop would round the jitter away).
- `active_schedule.json` renders `%H:%M:%S` and reports `jitter_seconds` so the dashboard matches reality.
- Wired in `commodity_engine.start()` after `apply_schedule_offset` so DEV offset + per-engine jitter compose cleanly.

## Test plan
- [x] `pytest tests/test_engine_jitter.py` — 11 pass
  - bounds, determinism across calls, determinism across subprocess restarts (the real check that `blake2b` is in use, not `hash()`), case-insensitive ticker, current tickers don't all collapse, custom window respected, uniform shift, different schedules across tickers, 45s window stays within the same minute.
- [x] `pytest -k 'schedule or jitter'` — 55 pass, no regressions.

## Acceptance criteria (from #1674)
- [x] `apply_engine_jitter` implemented with deterministic hash-based offset.
- [x] Wired into `commodity_engine._setup` after `apply_schedule_offset`.
- [x] `_write_active_schedule` reflects the jittered times so the dashboard matches reality.
- [x] `test_engine_jitter_is_deterministic` — same ticker yields same offset across runs (and across process restarts).
- [x] `test_engine_jitter_different_tickers_differ` — current tickers don't all collapse to the same bucket.
- [ ] DEV/PROD observations — requires deploy + 3 trading days of log capture (orthogonal to code review).

## Risks
- Jitter window is sub-minute, so no task crosses into the next slot (test covers this).
- Changes are runtime-only; backtest determinism is untouched.
- `get_next_task` now waits for `HH:MM:SS` instead of `HH:MM:00`. Day-one shifts each engine's fire time by up to 45s — verified safe against all existing `is_trading_day` / market-window guards.


rozavala:
Addressed the near-midnight wrap review item in 0a628c4.

**Bug shape**
The old `apply_engine_jitter` used `datetime.combine(today, t) + timedelta(seconds=offset)` followed by `.time()`. When the shift crossed midnight — e.g. task at `23:59:50` + `30s` jitter — the `.time()` call silently discarded the day and returned `00:00:20`. The scheduler keys off HH:MM:SS within the current day, so that task would register as an already-passed morning slot rather than a late-evening one. No production tasks run that close to midnight today, but the latent footgun would bite anyone who adds one.

**Fix**
Detect cross-midnight shifts and leave those specific tasks unjittered (with a `logger.warning`). Other tasks in the same engine still get their per-engine offset — only the wrap-prone tail is held back. Chose "skip shift for wrappers" over "clamp to 23:59:59" because clamping creates a new collision attractor (every wrapping engine converges on the same second), which defeats the fan-out purpose entirely.

**Tests** — 3 new cases in `tests/test_engine_jitter.py`:
- `test_apply_jitter_skips_shift_when_wrap_would_cross_midnight` — direct regression
- `test_apply_jitter_partial_wrap_only_affects_late_tasks` — mid-day tasks still shift, only the wrap-prone one is preserved
- `test_apply_jitter_task_exactly_at_midnight_minus_offset_is_preserved` — the `00:00:00 next day` boundary case

14/14 green.

**Not addressed here**: the same wrap hazard exists in `apply_schedule_offset` (the DEV `-30m` offset, which can wrap *backward* across midnight for morning tasks). Different enough semantically — backward wrap, minutes not seconds, different call site — that I'd rather file it as its own follow-up issue if you agree. Happy to roll it in if you'd prefer to bundle.

The other reviewer item — "dashboard seconds display vs. document minute-granularity" — is mostly a doc/UI decision rather than a code fix. Want me to tackle that next, or file-and-move-on?

rozavala:
Follow-up: dashboard now correctly handles the seconds-granular times this PR writes.

**Bugs fixed**
- `pages/1_Cockpit.py:752` parsed `%H:%M`, which raised on the new `HH:MM:SS` format; a bare `except` swallowed the exception and the entire "Next Up" countdown block silently disappeared.
- `dashboard_utils.py:880` compared scheduled vs. current time in minutes-since-midnight, so a task at `09:00:30` was bucketed to the same minute as `now=09:00:29` and mis-classified as *overdue* one second before it was actually due to fire.

**Changes**
- Cockpit parses both `HH:MM:SS` and legacy `HH:MM`; countdown reports seconds when <60s out.
- Dashboard classifier uses seconds-since-midnight.
- Cockpit caption now surfaces `jitter_seconds` from `active_schedule.json` ("DEV (jitter +3s)") so the operator can see the per-engine offset at a glance.
- New `tests/test_dashboard_seconds_schedule.py`: 4 tests covering seconds parsing, legacy fallback, the 09:00:29 vs 09:00:30 boundary, and `jitter_seconds` passthrough.

**Tests**
`pytest tests/test_engine_jitter.py tests/test_schedule_config.py tests/test_dashboard_seconds_schedule.py tests/test_dashboard_loading.py tests/test_dashboard_state_caching.py` → 41 passed.

rozavala (commented):
**Code review + test run: ✅ All 18 primary tests pass** (14 in `test_engine_jitter.py`, 4 in `test_dashboard_seconds_schedule.py`). 62 schedule/jitter tests pass across the full suite. **Ready to merge.**

The implementation is careful and production-safe:

- Using `hashlib.blake2b` over Python's salted `hash()` is the correct call — restart-stable offsets are a hard requirement for post-mortem timestamp reconciliation. `test_jitter_seconds_stable_across_process_restarts` guards this by spawning actual subprocesses.
- The 45s window is safely sub-minute — tasks cannot bleed into adjacent scheduler slots. `test_apply_jitter_stays_within_same_minute_slot` guards this invariant for all production tickers.
- The midnight-wrap guard in `apply_engine_jitter` correctly leaves late-night tasks unjittered rather than rolling them to morning slots where they'd register as already-missed.
- The active schedule is written **after** `apply_engine_jitter` in `commodity_engine.start()`, so dashboard-displayed times match actual fire times.
- Dashboard backward-compatibility (HH:MM fallback alongside new HH:MM:SS) is handled correctly in both `1_Cockpit.py` and `dashboard_utils.py`.
- `jitter_seconds` is exposed in the schedule JSON so the cockpit caption can display `(jitter +Xs)` for operator awareness.

The "11 pass" claim for `test_engine_jitter.py` is off — the file has 14 tests (3 midnight-wrap-guard tests added after the PR description). Broader "55 pass" is also off (62 match). All present tests pass.

---

rozavala:
Follow-up note on Round 2 reviewer items — several were already addressed in later commits on this branch before the review landed. Quick accounting:

- **Midnight-wrap preservation** → commit `0a628c4f6`. Tasks with `time_et` like `23:59:58` now use `dt_shifted.date() != today` to detect near-midnight wrap and keep the original time rather than silently drop the task.
- **Dashboard seconds rendering** → commit `a426619a0`. `dashboard_utils.load_task_schedule_status` switched minute-since-midnight classification to seconds-since-midnight, and `pages/1_Cockpit.py` parses both `HH:MM:SS` and legacy `HH:MM`. Countdown now surfaces seconds when < 60s.
- **Surface `jitter_seconds`** → same commit. Env caption now renders `(jitter +Ns)` when the active schedule file carries a non-zero `jitter_seconds` key.

Regression coverage for all three lives in `tests/test_dashboard_seconds_schedule.py` (4 tests: HH:MM:SS parsing, legacy fallback, 09:00:29 vs 09:00:30 boundary, jitter_seconds passthrough).

Let me know if the review was against an older revision — happy to rebase and squash if the history is getting noisy.

rozavala:
**Blocked — needs rebase on `main` after #1683.**

Seven of the eight open PRs just merged. #1683 (cleanup: remove dormant DEV schedule offset) landed, which removed `apply_schedule_offset` and its import from `trading_bot/commodity_engine.py` and `orchestrator.py`. This PR still imports and references `apply_schedule_offset`, so GitHub now reports `CONFLICTING`.

**Rebase checklist:**
- `trading_bot/commodity_engine.py`: drop `apply_schedule_offset` from the import block; the `if not is_prod:` offset branch has been removed upstream — keep only the jitter call.
- `orchestrator.py`: the `apply_schedule_offset` function definition is gone; ensure `apply_engine_jitter` placement doesn't re-introduce it.
- `_write_active_schedule` signature upstream is now `(task_list, env_name, is_prod)` (3 args). Re-thread `jitter_seconds` field without bringing back `schedule_offset_minutes`.

Everything else in the PR (blake2b, midnight-wrap guard, dashboard seconds fix) is green in review. Ping when rebased and CI is clean.

rozavala:
Rechecked: GitHub's CONFLICTING status was stale — the branch is actually based on current `main` (merge-base = main HEAD after the #1683 cleanup landed). Ran `tests/test_engine_jitter.py` (14 passed) and `tests/test_schedule_config.py` (15 passed) locally. `grep apply_schedule_offset` is empty across the tree. Merged. 🚀

PR #1686, opened 2026-04-21
fix(deploy): manage dashboard via systemd to avoid port-bind race

rozavala:
## Summary
- `deploy.sh` was killing the dashboard with `pkill -f "streamlit run dashboard.py"` and then starting a fresh copy via `nohup streamlit ...`. The dashboard is already managed by `coffee-dashboard.service` (`Restart=always`, `RestartSec=5`), so after `pkill` systemd respawned it in ~5s — beating the script's `nohup` start to port 8501. Result: misleading `"Dashboard started (PID: X, port: 8501)"` in deploy logs for a process that immediately died with `"Port 8501 is not available"`.
- Replace `pkill` + `nohup` with `systemctl stop` / `systemctl start coffee-dashboard`. Explicit `systemctl stop` suppresses `Restart=always` until we issue the matching `start` in step 9, eliminating the race.
- Both services (`trading-bot.service` and `coffee-dashboard.service`) are now managed the same way — via systemd — so the deploy verb is consistent. Restart policies intentionally kept asymmetric: `on-failure` for trading-bot (don't hammer a broken bot), `always` for the dashboard (silent UI outage is worse than auto-revive).
- Added backward-compat fallback: if `coffee-dashboard.service` is absent on a host (`systemctl start` returns non-zero), deploy.sh falls back to the old `nohup` path. No box gets bricked by this change.
- Same fix applied to the rollback path for consistency.

## Context
Observed in the 2026-04-21 DEV deploy after merging #1685. The orchestrator side was fine (`trading-bot.service` started cleanly, pool-key logs showed healthy `SB_reader`/`NG_reader`/`ZW_reader` client IDs). Only the dashboard startup was misreported.

Root cause is a latent dual-management bug: deploy.sh treats the dashboard as unmanaged (`nohup`), but the systemd unit exists and owns it. The two compete.

## Test plan
- [x] `git apply --check` clean, `git diff` shows only the three intended hunks (rollback, step 1 stop, step 9 start).
- [ ] DEV deploy: confirm `"coffee-dashboard.service started (PID: N, port: 8501)"` replaces the spurious PID line, and `logs/dashboard.log` shows a single clean startup (no "Port 8501 is not available").
- [ ] DEV deploy: confirm `systemctl status coffee-dashboard` returns `active (running)` after deploy completes.
- [ ] Rollback path: not easy to exercise without forcing a verify_deploy failure — will validate by inspection only.
- [ ] Fallback path: N/A on DEV/PROD (unit is installed); preserved for hosts without the unit file.


rozavala:
Thanks for the review. Second commit `d1dba09a9` addresses the substantive items:

**#1 (medium) — Silent dashboard failure escapes** → Fixed. The `is-active` check now `exit 1`s instead of warning. Chose `exit 1` over `rollback_and_restart` deliberately: by the time we reach this point the trading-bot has already started cleanly on the new code, and a dashboard-only bug shouldn't drag it back. CI marks the deploy failed; operator investigates via `journalctl -u coffee-dashboard -n 100`. Happy to switch to `rollback_and_restart` if you disagree on the tradeoff.

**#2 (medium) — Stray streamlit on stop path** → Fixed. `pkill -f "streamlit run dashboard.py"` now runs unconditionally after `systemctl stop`, not only on its failure. Closes the transition gap you described.

**#3 (low) — DASH_PID="0"** → Fixed. `[ "$DASH_PID" = "0" ] && DASH_PID="?"` guard added.

**#4 (low) — Silent `2>/dev/null` on systemctl stop** → Kept silent. Rationale: on hosts without the unit installed (legitimate fallback case), systemctl prints `Failed to stop coffee-dashboard.service: Unit coffee-dashboard.service not loaded` which is noise. And the unconditional `pkill` that runs after now provides self-correction for the non-noise failure modes — if stop genuinely fails to stop a running unit, step 9's `systemctl start` will fail loudly (with the new `exit 1` from #1), so it doesn't silently escape. Open to changing if you prefer.

**#5 (low) — Log path divergence** → Added a one-line comment above the dashboard block noting `journalctl -u coffee-dashboard` is the new log source and `logs/dashboard.log` is only populated in the fallback path. Also included `journalctl` invocation in the #1 error message.

**#6 (nit) — `systemctl show --value` ≥ 230** → Not changed. Deploy targets are modern Ubuntu on DO droplets (systemd 249+); 230 dates to 2016. If we add older targets this can be adapted using the `${VAR#MainPID=}` pattern already used for the orchestrator.

**#7 (nit) — `curl localhost:8501` reachability check** → Out of scope for this PR; agreed as a good follow-up.

PR #1687, opened 2026-04-21
fix(order_manager): skip BAG summary fills in on_exec_details

rozavala:
## Summary

- `on_exec_details` now ignores IB's BAG-level summary fill so `is_filled` can latch True on multi-leg combo orders again
- Matches the filter `_handle_and_log_fill` already applies at `order_manager.py:1947`
- Restores correct summary Pushover text after fills (regression introduced by ef763fadc)

## Root cause

For every multi-leg combo (bear/bull spreads, iron condors, straddles) IB emits **one fill event for the BAG contract plus one per real leg**. `on_exec_details` was tracking all of them in `filled_legs`, so:

- Bear put spread (2 legs) → `filled_legs` dict ends up with 3 keys (BAG + 2 legs)
- Gate check `len(filled_legs) == total_legs` is `3 == 2` → False
- `is_filled` never becomes True
- The final summary loop (L3618-3779) finds zero "filled" entries → Pushover reports `\"Order monitoring complete. No orders were filled or needed cancellation.\"`

Before ef763fadc (\"record all lots in multi-lot combo fills\"), `filled_legs` was a `set` and the gate fired on the first event whose add made `len==total_legs`. Because BAG fires before leg2, the check latched True at that midpoint by coincidence. That commit fixed multi-lot tracking but broke the coincidence.

## Impact

- **Trading logic: unaffected.** Ledger, TMS, council_history, execution_funnel all write correctly — they go through `_handle_and_log_fill`, which already filters BAG.
- **User-visible:** every spread fill silently reported \"no fills\" in the summary Pushover. Affects every strategy (spreads, condors, straddles).

## Evidence from today's DEV logs (2026-04-21)

Two SB bear put spreads filled successfully:

| Order | Time (UTC) | Spread | Qty | Fill | Logged to ledger/TMS/council | Pushover said |
|---|---|---|---|---|---|---|
| 245 | 09:29:45 | SON6 P0.1325/P0.125 | 2 | 0.27 | ✅ | ❌ \"no fills\" |
| 250 | 13:28:47 | SON6 P0.1375/P0.1275 | 2 | 0.40 | ✅ | ❌ \"no fills\" |

Neither order produced the `\"All 2 legs of combo order X fully filled\"` log line that fires when `is_filled` latches — which was present for every fill on April 13 and 17 (pre-regression).

## Test plan

- [x] `pytest tests/test_order_manager.py` — 55/55 pass
- [ ] Verify in DEV after deploy: next spread fill should produce `\"All 2 legs of combo order X fully filled\"` log line and a Pushover listing the filled spread by name and price
- [ ] Confirm ledger/TMS/council writes continue to land (should be unchanged — this path is untouched)


rozavala:
## Review summary — ship it

**Tests:** 1362/1362 passing (order_manager: 55/55).

### Verified claims
- **BAG filter at `_handle_and_log_fill` L1947** exists and short-circuits on \`isinstance(fill.contract, Bag)\`. The fix mirrors it exactly at L2688-2690 (4 lines).
- **Regression origin confirmed** via \`git show ef763fadc\`: \`filled_legs\` was a \`set()\` pre-commit. The gate latched by coincidence when \`size == total_legs\` happened to hit mid-fill (BAG+leg1 made size 2 for a 2-leg spread). The dict change preserves BAG entries indefinitely, so \`3 != 2\` for every 2-leg combo afterward. Theory holds.
- **Downstream trading sinks are genuinely unaffected.** \`log_trade_to_ledger\` (L1977), \`log_funnel_event POSITION_OPENED\` (L1999), and TMS thesis writes (L2008+) all live inside \`_handle_and_log_fill\`, dispatched unconditionally per non-BAG leg at L2713. They are NOT gated by \`is_filled\`. Ledger, TMS, council_history, execution funnel all correct pre- and post-fix.
- **Arithmetic post-fix:** BAG never enters \`filled_legs\`, so \`len(filled_legs) == total_legs\` compares (n real legs) == (n comboLegs). Clean.

### Edge cases — all safe
- Single-leg orders: no BAG event, filter is a no-op.
- IB omits BAG for some combos: pre-fix this was the *only* working case; filter doesn't affect it.
- Partial fills spread across events: multi-lot tracking via \`current_fills\`/\`expected_qty\` untouched.
- Ordering of BAG vs legs: irrelevant now — BAG always skipped.

### One small thing (non-blocking)
The \`is_filled\`-gated block in the summary loop (L3644-3670) also drives ORDER_FILLED funnel events and catastrophe-stop cleanup post-fill. Pre-fix, these were also silently skipped alongside the Pushover summary — the PR description slightly understates impact (cat-stop cleanup may have been orphaned until other cleanup paths caught it). The fix resolves all three with one change. Worth a one-liner in the PR body but no code change needed.

A regression test simulating a BAG fill event in \`on_exec_details\` would lock this behavior down — follow-up, not a blocker.

No CLAUDE.md violations. Minimal scope, exact symmetry with the existing L1947 filter.

Approving.

Issue #1689, opened 2026-04-21
chore(notifications): PR-B — Deduplicate Tier-2 Pushover notifications


rozavala:
## Context

Follow-up to #1688 (PR-A: Tier-3 demotion). After the Tier-3 noise is gone, the next-highest-volume sources are **Tier-2 duplicates and flapping events** — notifications that fire real signal but too often or redundantly with other pushes.

**Blocked on:** observing #1688 on DEV for 1–2 days to confirm the expected ~70% daily-volume reduction before touching more notification logic.

## Scope

Three surgical dedup changes, each isolated enough to revert independently if needed.

### 1. Drop \`Trading Orders Queued\` (pure duplicate)

- **Current:** Every order generation triggers a \`Trading Orders Queued\` push, followed seconds later by \`Order Placed\` with the same underlying info.
- **Proposal:** Remove the \`Queued\` push entirely. The \`Placed\` push is the authoritative signal.
- **Risk:** None — the \`Placed\` event already contains all the same info plus IB order ID.

### 2. Rate-limit \`ORDER REJECTED\` per order-id

- **Current:** IB can emit multiple rejection events per order (e.g., exchange rejection followed by connection re-sync). Each becomes its own Pushover.
- **Proposal:** Keep first rejection per \`orderId\` immediately; suppress subsequent rejections for the same \`orderId\` within a 60s window. Log all suppressed rejections at WARNING.
- **Risk:** Higher. A legitimate second-order rejection could be masked. Mitigations:
  - Scope to **same \`orderId\`** only (not per-ticker) — distinct orders still push immediately.
  - 60s window is short enough that any resubmit would come after.
  - Suppressed events still hit logs for forensic trail.
- **Suggest this lands in a separate PR from change 1 and 3** so it can be reverted cleanly.

### 3. Suppress IB Gateway Lost/Reconnected flap pairs < 120s

- **Current:** Brief network blips cause a \`Gateway Lost\` push immediately followed by \`Reconnected\` within seconds. User sees two alerts for a non-event.
- **Proposal:** Delay the \`Gateway Lost\` push by 120s. If \`Reconnected\` fires in that window, suppress both (log-only). If still disconnected at 120s, push \`Gateway Lost\` + duration.
- **Risk:** Legitimate outages are pushed 120s later. Acceptable tradeoff — if the gateway is down >2min the alert is still actionable; if <2min, no action was needed anyway.

## Expected impact

[...]

PR #1688, opened 2026-04-21
chore(notifications): demote 15 Tier-3 Pushover sites to log-only

rozavala:
## Summary
- Exhaustive audit of the ~60 Pushover call sites found Tier-3 routine lifecycle events (sentinel triggers gated by severity, startup banners, periodic reconciliations, health checks) generating the bulk of daily notifications while duplicating log content. This converts 15 such sites to `logger.info`/`warning`, expected to drop ~70% of daily push volume while preserving Tier-1 trade actions and Tier-2 infra-critical alerts.
- Critical alerts kept: REMOTE-GW+LIVE-MODE misconfig (commodity_engine), FLASH CRASH, ORDER REJECTED, all trade entry/exit pushes.
- Paired test updates: `test_phantom_ledger_reconciliation` and `test_prediction_market_sentinel` no longer assert on the demoted Pushover calls.

### Demoted sites
**orchestrator.py** — Sentinel Trigger push-gate, Restart Recovery, OFF-mode startup, Thesis Cleanup, Thesis Reconciliation, Phantom Ledger Reconciliation, Position Audit Complete, Sentinel Effectiveness (HIGH/CRITICAL), State Sync Warning, Health Check Error, Brier Reconciliation Stall
**trading_bot/commodity_engine.py** — Engine OFF-mode startup
**trading_bot/sentinels.py** — Prediction Market Rollover (now `_sentinel_diag.info`)
**scripts/error_reporter.py** — Issues Created summary (issues themselves still created in GitHub)
**scripts/check_model_availability.py** — both OK and problem paths (CI job still exits 1 on problems, surfacing via GitHub Actions)

## Test plan
- [x] `pytest tests/` — 1362 passed
- [x] Syntax check on all 5 modified source files
- [ ] Post-deploy: observe daily Pushover count on DEV; expect ~70% reduction
- [ ] Confirm Tier-1 trade-action pushes still fire on next live cycle (ORDER PLACED, TRADE OPENED, EXIT notifications)
- [ ] Confirm critical alerts unaffected (FLASH CRASH, ORDER REJECTED, REMOTE-GW misconfig)

## Follow-ups (deferred)
- PR-B: Tier-2 deduplication (drop `Trading Orders Queued`, rate-limit `ORDER REJECTED`, suppress Gateway Lost/Reconnected pairs <120s)
- PR-C: Quality fixes (add outcome context, normalize emojis, include contract month/strike in ORDER REJECTED)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1701, opened 2026-04-22
chore(notifications): #1689 PR-B1 — drop Queued push + Gateway flap suppression

rozavala:
## Summary

Two of the three dedup changes from #1689 (PR-B). Change 2 (ORDER REJECTED rate-limit) lands in a **separate PR (B2)** per the issue's explicit request for clean revertability.

### Change 1: drop \`Trading Orders Queued\` Pushover

Fires at signal generation and is superseded seconds later by \`Orders Placed & Monitored\`, which carries richer per-order context (strategy, direction, confidence, thesis_strength) plus IB order IDs. Queued content preserved at \`logger.info\` for the forensic trail.

### Change 3: Gateway Lost/Reconnected flap pairs < 120s

Error 1100 now schedules the \`IB Gateway Connectivity Lost\` Pushover via \`loop.call_later(GATEWAY_FLAP_WINDOW=120)\`. If 1101/1102 arrives inside the window, the pending Lost push is cancelled and the paired Reconnected is suppressed (log-only). If still disconnected at 120s, Lost fires with \`Still disconnected after Ns\` and a later 1102 sends the paired Reconnected with duration.

Existing 300s \`CONNECTIVITY_ALERT_COOLDOWN\` is unchanged — it stacks below the flap window.

### Latent bug fixed

Existing Gateway notifications used \`from trading_bot.notifications import send_pushover_notification\`, but that symbol only lives in the top-level \`notifications\` module. The surrounding \`except Exception: pass\` swallowed the \`ImportError\`, so Gateway Lost/Reconnected Pushovers have **silently never fired**. Fixed all four import sites — otherwise the flap-suppression has nothing to suppress.

## Acceptance criteria (from #1689)

- [x] \`Trading Orders Queued\` push site removed; log entry preserved at \`logger.info\`
- [x] Gateway Lost/Reconnected pair suppression keyed on 120s reconnect window
- [x] Tests updated; full suite passes
- [ ] \`ORDER REJECTED\` rate-limiter — **deferred to PR-B2 per issue scope**

## Tests

- 3 new tests in \`TestGatewayFlapSuppression\`: short flap absorbed, long flap fires Lost + Reconnected, duration in body.
- Full suite: **1369 passed**.

## Expected impact

Combined with PR-B2, targets the ~15–20% additional reduction in daily Pushover volume on top of PR-A (#1688). Change 3 alone eliminates almost all brief-flap noise — the most common source of low-signal Gateway pairs.

## Test plan

- [x] Unit tests green
- [ ] DEV observation for 1 day post-deploy (per #1689)
- [ ] Verify Pushover counts drop for Queued (→0) and Gateway Lost pairs with short flaps (→0)
- [ ] Confirm real outages (>120s) still push Lost + Reconnected with duration


rozavala:
APPROVE — clean, correct, well-scoped PR-B1.

**Verified:**
1. Scope is tight: 3 files, no ORDER REJECTED rate-limit logic leaks in from B2. Grep on `ORDER REJECTED` sites (`order_manager.py:2788/2819/2851`) shows they are untouched.
2. Drop Queued push is safe — forensic trail preserved at `logger.info`, and the authoritative `Orders Placed & Monitored` push still fires downstream at `order_manager.py:3424` with richer context (strategy, direction, confidence, IB order IDs).
3. Flap-suppression lifecycle is correct: `_pending_lost_handle[purpose]` is popped both on timer-fire and on `1102` cancel path; `_lost_at[purpose]` is popped via `.pop(purpose, now)` with a safe default. Keying by `purpose` avoids cross-engine collision on shared class-level dicts.
4. Fail-closed on ambiguity: `_fire_lost_notification` early-returns if `_connectivity_lost` was cleared — only reachable if 1102 arrived after timer started. In that scenario the Reconnected path owns the user message (either suppresses both or sends Reconnected pair), so no notification is truly lost.
5. Restart semantics: class-level dicts reset on orchestrator restart, so a restart in the middle of an outage gets a fresh 120s window. No re-spam; if Gateway is still down post-restart, next 1100 triggers a fresh delayed Lost push.
6. Latent bug fix confirmed — `send_pushover_notification` only exists in top-level `notifications.py`, not `trading_bot/notifications.py` (which only defines `NotificationTier`). The prior `from trading_bot.notifications import send_pushover_notification` raised `ImportError` swallowed by `except Exception: pass`, meaning Gateway Lost/Reconnected Pushovers had silently never fired. Fixing all four sites is required for the flap-suppression to have anything to suppress.

**Tests:** Full suite 1369 passed (matches PR description). New `TestGatewayFlapSuppression` (3 tests) exercises short-flap absorb, long-flap fires Lost+Reconnected, and duration-in-body.

**Minor non-blocking notes:**
- `asyncio.get_event_loop()` is deprecated in 3.12; the `except RuntimeError` fallback handles the no-loop case so it's safe, but `get_running_loop()` would be the modern preference.
- A theoretical microsecond race where `1102` cancel() loses to an already-firing timer is handled by the defensive early-return — rare and acceptable given 1102 == connectivity back.

Labeling Ready to be merged.

PR #1702, opened 2026-04-22
chore(notifications): #1689 PR-B2 — rate-limit ORDER REJECTED per orderId

rozavala:
## Summary

The third and final dedup change from #1689 (PR-B), split into its own PR per the issue's explicit request so it can be reverted independently of PR-B1 (#1701).

IB can emit multiple rejection events for the same order (exchange rejection echoed by a connection re-sync, margin re-check, etc.), each of which currently becomes its own Pushover. First rejection per \`orderId\` now pushes immediately; any subsequent rejection for the same \`orderId\` within a 60s dedup window is logged at WARNING with an explicit \"suppressed\" marker. Distinct orders still push.

## Scope

Three rejection sites in \`order_manager.py\`, all keyed on the IB \`reqId\`:
- \`ORDER REJECTED - INVALID PRICE\`  (Error 201, first-attempt reject)
- \`ORDER REJECTED - MARGIN\`         (Error 201, margin/insufficient)
- \`ORDER REJECTED - OFF-TICK\`       (Error 110)

A module-level \`_should_push_order_rejection(order_id)\` helper owns the window state with opportunistic cleanup of entries older than the window to bound memory.

## Risk mitigations (from #1689)

- Keyed on **same \`orderId\` only** — different orders still push immediately.
- Window is short (60s); any resubmit arrives after.
- Suppressed events are logged at WARNING with the full errorString, so the forensic trail is intact.
- Malformed orderIds (None, non-int) never suppress — fail open on notify.

## Tests

New \`TestOrderRejectionDedup\` (5 tests):
- first pushes, subsequent suppressed within window
- distinct orderIds push independently
- after window elapses, next rejection pushes again
- malformed orderId (None, str) never suppresses
- stale entries cleaned up to bound memory

Full suite: **1371 passed**.

## Acceptance criteria (from #1689)

- [x] \`ORDER REJECTED\` rate-limiter keyed on \`orderId\` with 60s window
- [x] Suppressed events logged at WARNING
- [x] Tests added; no regressions
- [ ] DEV observation for 1 day post-deploy

## Test plan

- [x] Unit tests green
- [ ] DEV observation: confirm Pushover volume drops for repeat-rejection storms while distinct rejections still push
- [ ] If a resubmit-after-60s is masked by mistake, revert this PR independently of #1701


rozavala:
Approved. Reviewed end-to-end; safety-critical checklist satisfied.

**Safety verification**
- 60s rolling window per `orderId`; first rejection pushes, subsequent within window suppressed. For a sustained storm the effective cadence is ~1 push/min, which is bounded and still actionable.
- Full fidelity preserved: every `logger.error(...)` call at all three sites (INVALID PRICE / MARGIN / OFF-TICK) fires unconditionally — only the Pushover is gated. Suppressed events log at WARNING with the full errorString.
- Memory bounded: opportunistic cleanup on each successful push purges entries older than the window. Dict size is O(distinct orderIds rejected in the last 60s), which is tiny.
- Fail-open on malformed ids: `None` / non-int `orderId` → `int(...)` raises → returns True → we over-notify rather than under-notify. Correct for a safety-critical surface.
- In-memory only — resets on restart. Acceptable for a 60s window; a restart within the window would at worst re-push an already-seen rejection, which is safer than the reverse.

**Merge-order vs #1701**
No file overlap. #1701 touches `connection_pool.py` + `test_connection_pool.py`; this PR touches `order_manager.py` + `test_order_manager.py`. Either can merge first; the issue-1689 authors explicitly split them so this one can be reverted independently.

**Tests**
- 5 new `TestOrderRejectionDedup` tests cover: first-pushes-subsequent-suppressed, distinct-orderId isolation, window-expiry re-push, malformed-id fail-open, and stale-entry cleanup.
- Full suite: 1371 passed, 0 failed (12m33s).

**Non-blocking note for DEV observation**
There is no escalation path for a genuinely sticky rejection (one push per 60s forever while the storm continues). If DEV shows a pathological case where the same rejection legitimately persists for many minutes, a log-volume check or an eventual-re-push at N*window could be a follow-up — but that is a tuning concern, not a correctness one.

Approving. Ready to merge.

PR #1703, opened 2026-04-22
chore(eval): PR 1 — grounded packet archival

rozavala:
## Summary

- New module `trading_bot/grounded_packet_archive.py` with `archive_packet`, `load_archived_packet`, `cleanup_old_grounded_packets`. Pure functions, fail-open, path-traversal safe (regex-validated cycle_id and persona_key), atomic writes.
- One call wired into `TradingCouncil.research_topic` after Phase 1 and **before** the inventory/supply_chain numeric-content early-exit, so packets that cause a bail are still captured.
- New pre-open scheduled task `cleanup_old_grounded_packets` (-14m) with 30-day default retention. Refuses to run with `max_age_days < 1` as a safety clamp.
- Multi-engine aware via `data_dir_context.get_engine_data_dir()`; silent no-op outside production (tests, DSPy, scripts).

## Why now

This is the foundation for the eval-harness work. The single biggest discovery from the prior eval-design pass was that we **don't persist the full `GroundedDataPacket` that was fed to each agent** — `council_history.csv` has the agent output but not the agent input. That means April 13 and every incident before it cannot be reconstructed for offline replay. Every minute of delay compounds — cases that could be mined from history have to be hand-written instead.

Landing this first, as pure addition, means every cycle starting today generates eval-ready artifacts.

## Rollout (safe by design)

1. Merge this PR. `grounded_packet_archival.enabled: false` in `config.json` — zero files written. Scheduled task runs but no-ops.
2. On DEV KC, flip `enabled: true`. Verify files appear in `data/KC/grounded_packets/` after the next cycle. Verify cleanup task logs "removed 0 file(s)" on first run.
3. After a week on DEV with no disk issues, flip PROD. Disk budget: ~285MB per commodity steady-state at 12 cycles/day × 8 personas × ~5KB/packet × 30 days.

## Test plan

- [x] 25 new tests in `tests/test_grounded_packet_archive.py` — happy path, path-traversal rejection (cycle_id and persona_key), disabled flag, no-engine-context no-op, I/O error swallow, cleanup age/clamp/missing-dir/unreadable-file behavior.
- [x] Full test suite: 1391 passed.
- [x] AST-parse all edited files.
- [x] ContextVar fixture teardown to prevent state leak into other tests (caught during full-suite run).


rozavala:
Approved — clean, safe-by-default foundational PR. Shipping with `enabled: false` means zero behavior change on merge.

## Safety concerns — all 7 addressed

| # | Concern | Status | Evidence |
|---|---------|--------|----------|
| 1 | Path-traversal safety | Addressed | `_CYCLE_ID_RE = ^[A-Z]{2,4}-[a-z0-9]{4,}$` and `_PERSONA_KEY_RE = ^[a-z_][a-z0-9_]*$` reject `/`, `.`, `..`, spaces. 13 parametrized rejection tests cover the obvious escapes. `.resolve()` check is belt-and-suspenders and not needed given the regex. |
| 2 | Atomic writes | Addressed | `open(tmp_path) → fsync → os.replace(tmp, final)` in `grounded_packet_archive.py:136-140`, with `.tmp` cleanup on failure. |
| 3 | Fail-open safety | Addressed | Outer `try/except` in `agents.py` downgrades any archive error to `logger.debug`. `archive_packet` itself catches every exception and returns `False`. Test `test_archive_packet_swallows_io_errors` locks this in. |
| 4 | Multi-engine aware | Addressed | Resolved via `data_dir_context.get_engine_data_dir()`, so each engine writes under its own `data/<TICKER>/grounded_packets/`. `LookupError` yields silent no-op outside engine context (tests, DSPy, scripts). |
| 5 | Bounded storage | Addressed | 30-day retention default, daily pre-open `cleanup_old_grounded_packets` scheduled task. Zero/negative `max_age_days` is clamped back to default to prevent wipes. PR summary provides the ~285MB/commodity/30d disk budget. |
| 6 | PII / secrets risk | Addressed (via inheritance) | `raw_findings` originates from `_gather_grounded_data` (search APIs, sentinel alerts) — the same content already persisted in `council_history.csv` and logs. No IB account data, no API keys in the packet payload. No new PII surface. |
| 7 | Hook point | Addressed | Placed in `TradingCouncil.research_topic`, the single choke point used by all scheduled + emergency + reflexion flows. Deliberately placed BEFORE the inventory/supply_chain numeric-content early-exit so bail cases are still captured. |

## Additional checks
- `schema_version: "1.0"` present in payload — replay tooling can version-gate.
- Serialization is deterministic-enough for the fixed schema (Python 3.7+ dict order). `sort_keys=True` would be nice-to-have for byte-level diff tooling, but not a blocker.
- Config entry documented inline via `_comment` with sensible defaults (`enabled: false`, `retention_days: 30`).
- Independently mergeable: `load_archived_packet` is unused outside tests; nothing breaks if #1704/#1705 never land.

## Tests
- `pytest tests/test_grounded_packet_archive.py`: **25/25 passed** (0.50s)
- Full suite (excluding live-IB): **1391 passed** (one unrelated brier-integration test was a flake under 30s timeout; passes at 60s).

Ship it.

Issue #1717, opened 2026-04-22
[Daily Summary] [PROD] 2026-04-22: 15 errors across 1 categories


rozavala:
## Daily Error Summary — 2026-04-22

**Environment:** PROD
**Total errors:** 15
**Categories:** `uncategorized`

### `uncategorized` — 15 occurrences

- **8x** `Make sure API port on TWS/IBG is open`
- **7x** `Peer closed connection.`

> The `claude-fix` label will be added after triage — Claude will automatically attempt a fix.

---
*Automated report by `scripts/error_reporter.py`*

rozavala:
Duplicate of #1711 — auto-summary of the same Apr 22 IB Gateway crash (API port / Peer closed connection errors). Consolidating investigation there.

Issue #1716, opened 2026-04-22
[CRITICAL] uncategorized: PORTFOLIO OUTAGE DETECTED: 8 IB connection failures across all purposes in th...


rozavala:
## Error Report

| Field | Value |
|-------|-------|
| **Category** | `uncategorized` |
| **Severity** | CRITICAL |
| **Environment** | DEV |
| **First seen** | 2026-04-22 16:06:40,320 |
| **Last seen** | 2026-04-22 16:16:40,765 |
| **Occurrences** | 4 |

### Sample Log Entry

```
PORTFOLIO OUTAGE DETECTED: 8 IB connection failures across all purposes in the last 5 min. IB Gateway may be fully unreachable.
```

### Impact

Investigate to determine impact on trading operations.

> The `claude-fix` label will be added after triage — Claude will automatically attempt a fix.

---
*Automated report by `scripts/error_reporter.py` — fingerprint: `991ead3f09a7`*

rozavala:
Duplicate of #1711 — same Apr 22 IB Gateway crash incident (16:06-16:16 UTC wave within the documented timeline). Consolidating investigation there.

Issue #1712, opened 2026-04-22
[CRITICAL] uncategorized: PORTFOLIO OUTAGE DETECTED: 10 IB connection failures across all purposes in t...


rozavala:
## Error Report

| Field | Value |
|-------|-------|
| **Category** | `uncategorized` |
| **Severity** | CRITICAL |
| **Environment** | DEV |
| **First seen** | 2026-04-22 13:06:26,633 |
| **Last seen** | 2026-04-22 13:06:26,633 |
| **Occurrences** | 4 |

### Sample Log Entry

```
PORTFOLIO OUTAGE DETECTED: 10 IB connection failures across all purposes in the last 5 min. IB Gateway may be fully unreachable.
```

### Impact

Investigate to determine impact on trading operations.

> The `claude-fix` label will be added after triage — Claude will automatically attempt a fix.

---
*Automated report by `scripts/error_reporter.py` — fingerprint: `5482909e3efd`*

rozavala:
Duplicate of #1711 — same Apr 22 IB Gateway crash incident (13:06 UTC wave within documented timeline). Consolidating investigation there.

Issue #1711, opened 2026-04-22
investigate: IB Gateway crash 2026-04-22 — SB_reader connection stuck for 54min preceded full Gateway crash


rozavala:
## Incident Summary

**Date:** 2026-04-22
**Duration:** ~12:16 UTC → ~13:35 UTC (~79 minutes total)
**Resolution:** Manual IB Gateway restart
**Impact:** All IB connections lost from ~13:10 UTC; sentinel loops blind, thesis cleanup and deferred trigger processing both failed

---

### Phase 1 — SB_reader stuck (12:16–13:10 UTC, ~54 min)

At **12:15:45 UTC**, the SB engine completed a post-audit VaR cycle and cleanly disconnected `SB_reader`:

```
12:15:45 - trading_bot.connection_pool - INFO - Disconnecting IB (SB_reader)...
12:15:49 - Engine.SB - INFO - Next task 'signal_early' scheduled for 13:23:48 UTC
```

Only **35 seconds later**, the sentinel loop fired and attempted to reconnect `SB_reader` (client ID 821):

```
12:16:20 - trading_bot.connection_pool - INFO - Connecting IB (SB_reader) ID: 821 [REMOTE/PAPER]...
12:16:35 - ib_insync.client - ERROR - API connection failed: TimeoutError()
12:16:35 - trading_bot.connection_pool - ERROR - IB (SB_reader) connection failed. Next retry backoff: 10.0s
```

This immediately went into a 15-second-timeout loop. **All other readers (KC, NG, ZW) continued connecting successfully throughout this phase** — the problem was isolated to SB_reader on the Gateway side.

The connection pool's `FORCE RESET` mechanism fired four times (12:22, 12:29, 12:37, 12:44) without effect:

```
12:22:38 - trading_bot.connection_pool - WARNING - FORCE RESETTING connection: SB_reader
12:22:38 - trading_bot.connection_pool - ERROR - IB (SB_reader) connection failed. Next retry backoff: 5.0s
```

The outage notifier fired at **12:26:24 UTC** ("IB outage notification sent after 11 minutes"), but the outage at this stage was only SB_reader — the other commodities were unaffected.

### Phase 2 — Full Gateway degradation (13:10–13:35 UTC)

At **~13:10 UTC**, KC_reader, NG_reader, and SB_reader all began timing out simultaneously — the failure was no longer SB-specific:

```
13:10:11 - ib_insync.client - ERROR - API connection failed: TimeoutError()   [SB_reader]
[...]

PR #1725, opened 2026-04-22
fix(connection-pool): break Gateway compounding-failure loop (#1711)

rozavala:
## Summary

Three additive mitigations in `IBConnectionPool` to stop reconnect storms
from compounding an already-degraded IB Gateway. Root-causes the
Apr 22 2026 portfolio outage (#1711) and the downstream symptom storm.

- **Q7** — `_force_reset_connection` no longer snaps backoff to 5s. Floor
  at `FORCE_RESET_BACKOFF_FLOOR = 60s`; preserve larger in-flight values;
  stamp `_last_reconnect_attempt = now` so the floor is actually enforced
  by the next backoff check.
- **Q8** — `release_connection` / `release_all` now stamp
  `_last_clean_disconnect[pool_key]`. `get_connection` rejects requests
  within `RECONNECT_HOLDOFF_AFTER_CLEAN_DISCONNECT = 60s` with a
  `ConnectionError("in post-disconnect hold-off")` — prevents shutdown /
  cycle-transition open/close churn.
- **Q9** — when `is_portfolio_outage_active()` is True, `get_connection`
  floors the effective backoff at `PORTFOLIO_OUTAGE_BACKOFF_FLOOR = 120s`.
  All pool keys wait out the outage window instead of drumming on a dead
  socket.

All three gates are non-invasive (floors only; no existing longer
backoff is shortened; no happy-path behavior changes).

## Test plan

- [x] `pytest tests/test_connection_pool.py` — 54/54 passing (42 pre-existing + 12 new)
- [x] One existing test updated for Q7's changed contract
      (`TestForceResetIdempotentMapping::test_raw_purpose_and_pool_key_reset_same_state`
      now asserts that a seeded 120s backoff is preserved rather than
      reset to 5s, and that `_last_reconnect_attempt > 0`).
- [x] New coverage across all three gates:
  - `TestForceResetBackoffFloor` — applies floor from below, preserves
    above, stamps `_last_reconnect_attempt`.
  - `TestCleanDisconnectHoldoff` — release_connection / release_all
    stamp the key; get_connection respects the window; allowed past it.
  - `TestPortfolioOutageBackoffFloor` — floors under active outage,
    normal path past elapsed backoff when outage inactive, larger
    per-pool backoff preserved.
- [ ] Post-merge: watch the SB_reader / NG_reader zombie-detection path
      during the next cycle transition — expect to see one
      `in post-disconnect hold-off` log line per pool key rather than a
      burst of reconnect attempts.

## Context

Root cause: #1711. Symptom issues expected to resolve once deployed:

- #1715 (CRITICAL: TimeoutError during order generation prod)
- #1716 (CRITICAL: 8 IB connection failures dev)
- #1712 (CRITICAL: 10 IB connection failures dev)
- #1709 (CRITICAL: 5 IB connection failures dev)
- #1708 (CRITICAL: IB outage notification after 11 min)
- #1707 (CRITICAL: 11 IB connection failures dev)
- #1717 (Daily Summary prod 2026-04-22)
- #1713 (Daily Summary dev 2026-04-22)


rozavala:
Thorough review complete. Verdict: **FIX**. The three mitigations are individually well-reasoned and the test coverage is strong, but Q8 (clean-disconnect hold-off) as written creates a serious regression on the primary signal-cycle flow.

## Root cause of #1711 (understood correctly by PR)
Repeated reconnect attempts on a wedged Gateway compound pressure. Three vectors: (a) `_force_reset_connection` snapping backoff to 5s, (b) shutdown/cycle transitions reopening what was just cleanly closed, (c) all pool keys continuing to pound during an already-declared portfolio outage.

## Concern — Q8 regresses the signal cycle (BLOCKING)

`order_manager.generate_and_execute_orders` (the top-level scheduled signal handler) synchronously chains:

1. `generate_and_queue_orders(connection_purpose="orchestrator_orders")` — releases in `finally` (line 1924)
2. `place_queued_orders(connection_purpose="orchestrator_orders")` — acquires (line 2629)

Both use the **same pool key** (`{ticker}_orders`). With Q8, the release in step 1 stamps `_last_clean_disconnect[KC_orders]`. Step 2's `get_connection` immediately raises `ConnectionError("post-disconnect hold-off")`. `place_queued_orders` has `_max_conn_attempts=2` with 15s sleeps (total 30s), both still well inside the 60s hold-off, so both fail. Result:
- Compliance-approved orders land in `missed_orders.json`
- "Order Placement Connection Failed" Pushover fires every signal cycle
- No trades on a healthy Gateway

I verified this with a minimal test against the PR branch (reset state, release, then get → the `post-disconnect hold-off` ConnectionError is raised). **Test coverage gap:** the PR's `TestCleanDisconnectHoldoff` tests the mechanism in isolation but no test exercises the realistic release→get pattern on the same pool key.

This is fail-closed (visible, persisted, notified) rather than silent skip, so it's not catastrophic, but every scheduled cycle would now miss orders on a healthy Gateway. Shipping as-is would turn signal_early / signal_euro / signal_us_open / signal_peak / signal_settlement into noise.

## Other checklist items (all OK / addressed)

1. **Call sites / blast radius** — 22 files reference the pool. The `release_connection` → `get_connection` pattern recurs in `generate_and_execute_orders` (broken per above). Other paths (`audit` at 13:30, `var_audit`, `reconciliation`, `cleanup`, `close_stale_*`, `emergency_hard_close`) use different pool keys or are minutes/hours apart — safe.
2. **Failure counter semantics** — Consecutive failure counter unchanged. New counters: `_last_clean_disconnect` (stamp on release, checked on acquire, never cleared automatically — see observation below), `_portfolio_alert_at` (existed pre-PR, now read by Q9). Reset paths unchanged.
3. **Fail-closed vs fail-open** — Q9 is correctly fail-closed (blocks reconnects during known outage). Q7 is correctly fail-closed (floor prevents hammering). Q8 is fail-closed but over-eager per above.
4. **Race conditions** — Stamps are dict writes protected by the per-key `_locks[purpose]` inside `release_connection` / `release_all`, and the read on `get_connection` is also inside its lock. OK.
5. **Observability** — Hold-off log is `DEBUG` (invisible in prod). Consider `logger.info` on the first rejection per pool key so the Cockpit `grep` finds it. No Pushover on hold-off rejection (correct — it would spam the dedup).
6. **Test coverage** — 54/54 pass. Missing: back-to-back release→get on same pool key. Add a test like `test_release_then_get_on_same_key_is_allowed` (if intent is to allow it) or `test_release_then_get_on_same_key_respects_holdoff_and_callsite_adapts`.
7. **Interaction with #1701's flap suppression** — No conflict. Flap suppression lives in `_register_connectivity_handler` (Error 1100/1102) and operates on Pushover dedup timers. Q7/Q8/Q9 operate on reconnect backoff floors. Independent domains.
8. **CLAUDE.md conventions** — No tick/regime/state-atomicity concerns. State is in-memory (restart = fresh start). Consistent with the rest of `IBConnectionPool`.

## Suggested fix

Option A (minimal, preserves intent): Skip the hold-off check for pool keys where the caller released and re-acquired within the same orchestrator task — but that requires task-local tracking.

Option B (cleaner): Don't stamp `_last_clean_disconnect` in `release_connection` when the call is made from a code path that intends to immediately reacquire. Introduce a `release_connection(purpose, quiet=True)` or `release_connection(purpose, holdoff=False)` flag and pass it from `generate_and_queue_orders`'s `finally` block (and similar mid-cycle releases). `release_all` at shutdown would keep the default (holdoff=True).

Option C (behavioral): `generate_and_execute_orders` could share a single connection across the two calls instead of releasing between them. The pool's zombie detection on next `get_connection` would catch a dead socket. This avoids the churn Q8 was designed to stop, without needing Q8 to police it.

Option B is the tightest fit — it lets Q8 still catch the real shutdown/cycle-transition churn the author targeted while letting legitimate intra-cycle reuse proceed.

Please also add a regression test that pins down the intended semantics of back-to-back release→get on the same pool key (whatever you choose).

rozavala:
Addressed in eaf84821f — thanks for the catch on the `generate_and_queue_orders` → `place_queued_orders` flow; you're right that Q8 as-written would have killed every signal cycle on a healthy Gateway.

Went with **Option B** (minimal signature change, preserves Q8's intent for real shutdown/cycle-transition churn):

1. `IBConnectionPool.release_connection(purpose, holdoff: bool = True)` — default preserves #1711 Q8 behavior for all callers that do not opt out.
2. The one caller that immediately re-acquires the same pool key — `generate_and_queue_orders`' `finally` block at `order_manager.py:1930` — now passes `holdoff=False` with a comment pointing at this review. Audited all 14 other `release_connection` call sites; each is a genuine shutdown / cleanup path or has a multi-minute gap before the next acquire on the same key (emergency_hard_close, close_positions, close_stale_positions, cancel_all_orders, audit, var_audit, reconciliation, sentinel, cleanup, monitor, test utilities, verify_system_readiness), so they correctly keep the default.
3. Upgraded the hold-off rejection log from `DEBUG` to `INFO` (observation #5) so the Cockpit grep finds the first rejection without DEBUG enabled.

Regression tests added per your request:
- `test_release_then_get_same_key_with_holdoff_false_allowed` — pins down the signal-cycle semantics (release + immediate re-acquire must work with the opt-out).
- `test_release_then_get_same_key_with_default_blocks` — pins down the shutdown/cycle-transition contract (default still gates).

Suites: 56/56 `test_connection_pool.py` + 60/60 `test_order_manager.py` passing.

rozavala:
Re-review of commit `eaf84821` — fix verified, approving.

**Q8 over-application addressed cleanly:**
- `release_connection` gained `holdoff: bool = True` parameter with conditional `_last_clean_disconnect` stamp — default preserves the #1711 compounding-failure protection everywhere else (shutdown, cycle transitions, `release_all`).
- The one intra-cycle site (`generate_and_queue_orders` `finally` block) now passes `holdoff=False` with a clear comment explaining the `place_queued_orders` re-acquisition flow.
- DEBUG→INFO log level upgrade on the hold-off rejection so Cockpit greps find the first rejection per pool key.

**Test coverage:**
- `test_release_then_get_same_key_with_holdoff_false_allowed` — regression test for the signal-cycle flow.
- `test_release_then_get_same_key_with_default_blocks` — mirror confirming the default still gates.
- 56/56 `tests/test_connection_pool.py` pass locally.

Ready to merge.

PR #1695, opened 2026-04-21
🎨 Palette: Professionalize Position Detail table on Financials page

cristobalmoena:
This PR implements a micro-UX enhancement to the "Position Detail" table on the Financials page (`pages/4_Financials.py`) to improve "time-to-insight" for trade auditing.

### 💡 What
- **Iconography**: Prepend semantic emojis to all table headers (📦 Ticker, 🛡️ Strategy, 📜 Legs, ⏱️ Hold, 📐 Width, 💸 Entry Fill, 📊 Entry % Width, 💰 P&L).
- **Progress Bars**: Converted the `Entry % Width` column from a raw number to a `ProgressColumn`. This provides an immediate visual representation of capital utilization/cost relative to spread width.
- **Status Indicators**: The `Hold` column now distinguishes between `⏳ Open` positions and `⏱️ Closed` positions using icons, making the ledger easier to scan.
- **Context**: Added `help` tooltips to columns like `Hold`, `Width`, and `Entry % Width` to explain the underlying calculations.

### 🎯 Why
The previous table was a dense grid of numbers and text. By introducing visual hierarchy through progress bars and icons, users can now distinguish position status and relative trade costs at a glance without reading every cell.

### ♿ Accessibility
- Added descriptive `help` tooltips to all headers to provide screen-reader accessible context for abbreviated financial terms.
- Maintained high color contrast by using standard Streamlit components.

### ✅ Verification
- **Automated Tests**: Added `test_financials_position_detail_config` to `tests/test_ui_ux.py`. This test uses AST inspection to ensure the `st.dataframe` call in `pages/4_Financials.py` always includes the required emojis and the `ProgressColumn` type. All 53 UI/UX tests passed.
- **Frontend Verification**: Generated a screenshot using Playwright which confirms that the progress bars render correctly and the icons improve scannability.
- **Linting**: Verified code adheres to repository standards.

---
*PR created automatically by Jules for task [17675030842707120532](https://jules.google.com/task/17675030842707120532) started by @cristobalmoena*

PR #1726, opened 2026-04-22
🎨 Palette: Professionalize Scorecard tables

cristobalmoena:
Professionalized the data tables on the Scorecard page by adding semantic iconography, descriptive tooltips, and progress bars for key metrics. Added automated AST tests to ensure these UX standards are maintained.

---
*PR created automatically by Jules for task [7900979215235973898](https://jules.google.com/task/7900979215235973898) started by @cristobalmoena*


rozavala:
Approving. Clean, scoped palette work that adds real value to the Scorecard forensic tables.

**What it does well**
- Tooltips explain non-obvious forensic concepts (e.g., Vote WR % = "Hypothetical win rate if the overridden agent's vote had been followed", Net Value = "net value added (or destroyed) by the Master's override, in percentage points"). This is genuine context, not label restatement.
- ProgressColumn for bounded 0–100 ratios (Follow Rate, Override Rate) matches the pattern established by the merged Financials palette PR (#1695).
- No overlap with prior Scorecard palette work — #1304 handled metrics/Recent Decisions, #1574 handled Feedback Loop Health. The 10 dataframes touched here (Strategy, Alignment, Consensus, Regime, Thesis, Confidence, Trigger, Agent Override, Debate Quadrant, Regime-by-tilt) were previously rendered with default column configs.
- AST test locks the convention in.

**Scope check**
- Only `pages/2_The_Scorecard.py` (read-only dashboard), `tests/test_ui_ux.py`, and `.Jules/palette.md` touched. No execution / compliance / order_manager / signal code.
- Full test suite passes (1480 passed) in a clean worktree off the PR branch.

**Minor nit (non-blocking)**
- The "Net Value (pp)" column now uses `format="%+d pp"`, but the values are stored as `round(net, 1)` — floats with 1 decimal. Integer formatting will drop the decimal (e.g., `-15.8` displays as `-15 pp`). The `destructive < -15` threshold downstream still uses the underlying float, so no functional regression, just a small display precision loss versus the default render that was there before. If Jules wants a follow-up, `"%+.1f pp"` would preserve the decimal. Not a merge blocker.

PR #1728, opened 2026-04-23
feat(eval): activate cached-mode gate with per-agent cheap-model overrides

rozavala:
## Summary

Turns the eval harness from mock-only plumbing check into a real prompt-regression gate. Cached mode runs real LLM calls on first hit and replays from disk thereafter; per-agent overrides swap production flagship models (gpt-5.2, o3, Sonnet, Pro) for cheap-tier equivalents (Gemini Flash, gpt-4o-mini, Haiku, grok-fast-non-reasoning) — ~95% of regression signal at ~3% of cost.

## What ships

- `evals/runner/eval_models.py` + `tests/test_eval_models.py` — load + apply per-role overrides to `HeterogeneousRouter.assignments`; gracefully skips unknown roles, providers, and providers without API keys.
- `evals/config/eval_models.json` — the override map, with inline cost rationale.
- `evals/runner/harness.py` — wires override application at live-router init, surfaces the applied snapshot via `get_eval_overrides()`; `write_report` now includes it in the JSON payload.
- `evals/runner/__main__.py` — passes overrides into the report.
- `.github/workflows/eval-gate.yml` — existing `fast` (mock) job unchanged; new `full` (cached) job gated on PRs targeting `production` from the main repo, with response-cache restore/save keyed on prompt-shaping inputs.
- `.github/workflows/eval-drift-weekly.yml` — Monday 14:00 UTC cron; on red severity opens a deduped GitHub issue labeled `eval-drift-weekly` / `priority:high`.
- `scripts/eval_bootstrap_baselines.py` — one-shot: lint → verify API keys → run all cases in cached mode → aggregate → snapshot real baselines. Refuses to snapshot over failing cases unless overridden.
- `evals/README.md` — documents cached-mode activation, cache invalidation triggers, cost table, flagship-tier known limitation.

## Activation (post-merge)

Separate, deliberate step — not part of this PR:

\`\`\`bash
python scripts/eval_bootstrap_baselines.py --mode cached --snapshot
\`\`\`

Budget: ~\$0.50-\$1, ~8-12 min wall-clock. Replaces mock-mode baselines with real LLM values. Any `directional_accuracy` or `contrast_consistency` drop from the mock baseline indicates a real prompt bug that must be fixed before committing.

## Test plan

- [x] `pytest tests/test_eval_models.py -q` — 14/14 pass
- [x] `python -m evals.runner --lint-only` — 123 cases clean
[...]

PR #1723, opened 2026-04-22
chore(evals): extend inventory eval cases to NG/SB/ZW (+12)

rozavala:
## Summary

- 12 inventory eval cases added across NG, SB, ZW (4 per commodity: canonical bullish, canonical bearish, 2-file contrast pair).
- Each commodity uses the appropriate authoritative balance-sheet data source: EIA weekly storage for NG, ISO + Czarnikow + USDA FAS for SB, USDA WASDE + IGC for ZW.
- Contrast pairs hold production/supply fixed and flip only the balance-sheet metric, testing that the agent reads the stock-side variable rather than anchoring on crop size.
- must_not_cite forces balance-sheet-only evidence: blocks weather, FX, chart, positioning, supply-chain, and geopolitical tokens.

## Case inventory

| Commodity | Authority | Bullish | Bearish | Contrast axis |
|-----------|-----------|---------|---------|----------------|
| NG | EIA weekly storage | 3,412 Bcf, -268 vs 5y, DoS 52 | 2,485 Bcf, +412 vs 5y (30yr mid-winter record) | Storage vs 5y avg (-192 Bcf ↔ +244 Bcf) at fixed production + LNG feedgas |
| SB | ISO + Czarnikow + USDA FAS | ISO -4.8 Mt deficit, C/U 21.4%, India 4.2 Mt | ISO +6.2 Mt surplus, C/U 27.2%, India 9.6 Mt | C/U ratio (21.8% ↔ 27.4%) at fixed CS crop |
| ZW | USDA WASDE + IGC | World stocks 254.8 Mt (lowest since 2015/16), US S/U 32.1% | World stocks 292.2 Mt (highest since 2019/20), US S/U 39.2% | World S/U (33.4% ↔ 39.2%) at fixed US production |

## Test plan

- [x] \`python -m evals.runner --lint-only --agent inventory\` — 16/16 clean
- [x] \`python -m evals.runner --mode mock --agent inventory --snapshot\` — 16/16 pass, baseline refreshed
- [ ] CI eval-gate runs --lint-only + --compare-to-baseline on this PR

## Completes the eval-coverage expansion

This PR closes the planned expansion to 4-commodity coverage across all specialist analysts:
- sentiment (#1710), geopolitical (#1714), supply_chain (#1718), agronomist (#1719), macro (#1720), technical (#1721), inventory (#1722).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1729, opened 2026-04-23
fix(eval): scope commodity profile per case, not once per harness

rozavala:
## Summary

Fixes a latent correctness bug in `EvalHarness` that made all non-KC cases render against the KC agronomist prompt. Uncovered on the very first cached-mode smoke test after #1728 landed — mock mode had been papering over it since the harness was written.

## The bug

`EvalHarness.__init__` called `_load_commodity_profile()` once, pulling `config.symbol` (which is `"KC"`), and stored the result on `self.commodity_profile`. Every subsequent `run_case(...)` passed that single profile into `build_specialist_prompt`, ignoring `case.commodity`. So ZW and SB cases were prompted *as if they were Coffee (Arabica)* — regions, agronomy_context, and inventory sources all wired to KC.

Mock mode couldn't detect it: mock responses are synthesized from each case's own `expect:` block without touching the rendered prompt. The first real-LLM smoke test surfaced it — agronomist correctly said "no impact on Coffee (Arabica)" on wheat-data cases, dragging `directional_accuracy` and triggering false `forbid_hallucination` hits (the model was citing coffee-specific percentages while reasoning about wheat data).

## The fix

- Replace eager `self.commodity_profile` with a lazy, per-ticker cached loader (`_commodity_profile_for`).
- 4 profile loads for a 128-case run instead of 128.
- Unknown tickers degrade to `None` (already handled by `build_specialist_prompt`).

## Tests

Three new regression tests in `tests/test_eval_harness.py`:

- `test_harness_renders_commodity_specific_prompts` — **this is the one that would have caught the original bug**: same harness instance, two cases with different commodities, assert the builder receives two different profile objects.
- `test_harness_caches_profile_lookups` — 5 cases on the same ticker → 1 profile load.
- `test_harness_tolerates_missing_profile` — unknown ticker → None profile, no crash.

Also tightens `.gitignore`: adds `evals/.response_cache/` alongside the existing `evals/cache/` entry (the actual default cache path uses the hidden-dir prefix and wasn't previously excluded — discovered while confirming the PR-#1728 cache design).

## Smoke-test evidence (11-case agronomist run, \$0.05)

| Case | Pre-fix | Post-fix |
| --- | --- | --- |
| `bearish.003_sb_monsoon` | JSON error | BEARISH ✓ (conf over cap) |
| `bearish.004_zw_plains_moisture` | BULLISH ✗ | BEARISH ✓ (conf + lane) |
| `bullish.001_minas_frost` (KC) | 30% fabrication | 30% fabrication |
| `bullish.003_sb_centro_sul_drought` | 3× fabrication | BULLISH ✓ (conf over cap) |
| `bullish.004_zw_drought_hrw` | JSON error | BULLISH ✓ (28% detector FP on 45.2→32.5 math) |
| `contrast.zw_crop_condition_004_a` | BEARISH ✗ | **BULLISH ✓ PASS** |
| `contrast.zw_crop_condition_004_b` | JSON error | BEARISH ✓ (lane) |

All 3 JSON truncations, both directional misses, and all 3 false-hallucination hits resolved or recharacterized as different (and more tractable) issues.

## Known post-fix state (not blocking)

6 residual failures on the agronomist smoke, clustered as:

- **5× `confidence_max` breach** — Gemini Flash returns 0.95 across the board; cases cap at 0.88–0.9. Calibration issue inherent to the model, fixed by a separate model-swap PR for agronomist, not by raising caps (the caps guard against EXTREME-confidence sizing behavior).
- **2× `must_not_cite`** — Flash echoes "drought" / "SMA" tokens from the prompt adjacencies. Lane discipline, typically better on gpt-4o-mini / Haiku.
- **1× genuine extrinsic cite** — `minas_frost` claims "Minas Gerais represents 30% of total production" (accurate domain knowledge, not in `grounded_data`). Cheap-tier floor.
- **1× detector false positive** — `zw_drought_hrw` "28%" is `(45.2 − 32.5) / 45.2 ≈ 28.1%`, derived from two grounded numbers. Separate follow-up: tune the regex detector to whitelist arithmetic over grounded facts.

None of these block this PR; they're the *correct* residual after the harness bug is fixed.

## Test plan

- [x] `pytest tests/test_eval_harness.py -q` — 17/17 pass (14 existing + 3 new)
- [x] `python -m evals.runner --lint-only` — 123 cases clean
- [x] `python -m evals.runner --mode mock --compare-to-baseline` — green, no regression
- [x] Cached-mode re-smoke on agronomist — primary failures resolved as tabled above
- [ ] CI `fast` job runs green on this PR

## Follow-ups (separate PRs)

1. `eval:model-swap` — evaluate gpt-4o-mini as agronomist override, re-smoke, likely close 7 of the 8 remaining agronomist signal/calibration issues.
2. `eval:detector-tune` — relax `forbid_hallucination` regex when a percentage is derivable from two grounded numeric facts via basic arithmetic (addresses the `zw_drought_hrw` FP class).


rozavala:
Approving.

**Bug verified on main:** `EvalHarness.__init__` calls `self._load_commodity_profile()` once with `config.get("symbol", "KC")`, producing a single KC profile that `run_case` then hands to `build_specialist_prompt` for every case regardless of `case.commodity`. After #1723 added NG/SB/ZW cases, those non-KC cases were silently rendered against the Coffee (Arabica) prompt — exactly the smoking-gun symptom in the PR description ("no impact on Coffee (Arabica)" on wheat-data cases).

**Fix correctness:**
- Per-case lookup via `self._commodity_profile_for(case.commodity)` keyed on the canonical `EvalCase.commodity` field, which is already required at case-load time (`case.py:108`) — no risk of `None` ticker from malformed data (it'd fail earlier at case loading).
- Cache is keyed on ticker, so a 128-case run does ~4 lookups.
- Unknown ticker / loader exception → `None`, which `build_specialist_prompt` already handles (`specialist_prompt.py:79 if commodity_profile is not None:`). Good failure mode for the eval path (degrades to persona-only prompt, doesn't crash the run).
- Log level bumped from `debug` to `warning` on profile load failure — appropriate, since an unknown ticker in an eval case is noteworthy.

**Scope check:** Zero production code touched. Changed files: `.gitignore`, `evals/runner/harness.py`, `tests/test_eval_harness.py`. No `trading_bot/`, `config.json`, or `orchestrator.py` edits.

**Regression risk on existing KC cases:** None. Pre-fix, KC cases already got the KC profile (it was the only one loaded). Post-fix, KC cases get the KC profile via the per-case path. Identical behavior.

**Test coverage:** The new `test_harness_renders_commodity_specific_prompts` is exactly the regression test the bug deserved — same harness instance, KC + ZW cases, asserts the builder receives two different profile objects. Plus caching test and missing-profile tolerance test.

**Test runs:**
- `pytest tests/test_eval_*.py` → 69/69 pass (incl. 3 new regression tests).
- Full `pytest tests/` → 1497/1497 pass.

**.gitignore nit (informational, not blocking):** Adding `evals/.response_cache/` alongside the existing `evals/cache/` is correct — the harness default at `harness.py:101` is `evals/.response_cache` (hidden dir), so the old entry was masking the wrong path. Good catch.

Residual eval-smoke issues called out in the PR (confidence cap breaches on Flash, lane-token echoes, one extrinsic cite on `minas_frost`) are correctly characterized as separate follow-ups — they are post-fix residuals, not blockers.

Labeling Ready to be merged.

PR #1730, opened 2026-04-23
fix(agronomist): disambiguate SENTIMENT TAG as price direction, not crop outcome

rozavala:
## Summary

- Agronomist persona framed the task as *YIELD IMPACT* with rules like *Rain in 'Harvest' is BAD (Rot)* but never mapped crop-outcome judgments to BULLISH/BEARISH for **price**. The SENTIMENT TAG semantics were model-dependent.
- Add a *SENTIMENT RULE* line parallel to the existing `macro` persona's \"Weaker BRL = BEARISH for Coffee Prices (Supply Push). Stronger BRL = BULLISH.\"
- Explicitly maps supply-damaging events → BULLISH (supply tightens → price rises) and favorable weather → BEARISH (supply expands → price falls).

## Diagnostic evidence

11 agronomist cases, identical grounded data, three model families:

| Model family | Frame | Direction vs. `expect:` |
|---|---|---|
| Gemini 3 Flash | price-directional | 11/11 match |
| OpenAI gpt-4o-mini | crop-outcome | 11/11 inverted |
| Anthropic Haiku 4.5 | crop-outcome | 9/11 inverted |

Two of three model families read the persona as crop-outcome direction. Only Flash infers price direction unaided. Committing a Flash-only baseline would embed a minority interpretation as ground truth — so the fix is in the prompt, not the model selection.

## Test plan

- [ ] JSON validates (`python3 -c \"import json; json.load(open('config.json'))\"`)
- [ ] After merge: re-run agronomist smoke across Flash + gpt-4o-mini + Haiku. All three should agree on direction, and agronomist-case pass rate should be ≥ the Flash baseline on the other two models.
- [ ] No code reads the persona text by substring — confirmed via grep for `PHENOLOGY IS KEY` / `YIELD IMPACT` (only match outside `config.json` is a descriptive reference in `evals/cases/README.md`).

## Scope

- Touches only the `agronomist` entry under `gemini.personas` in `config.json`. No code, no test changes.
- Structural pattern matches the existing `macro` and `volatility` personas, which already encode price-side sentiment rules explicitly.


rozavala:
Closing — edit was in the wrong tier. `build_specialist_prompt` resolves commodity-profile templates (tier 2, `trading_bot/prompts/base_prompts.py::AGRONOMIST_CONTEXT`) ahead of the legacy `config.json` personas (tier 3). The eval harness passes `commodity_profile=profile`, so tier 2 always wins on the eval path — this PR's edit was inert.

Replaced by a tier-2 fix in a new PR that also addresses the cache-key bug discovered during diagnostic (eval-override state not included in `_cache_signature`).

PR #1704, opened 2026-04-22
feat(evals): eval harness for Tier-2 specialist agents

rozavala:
## Summary

Closes the biggest regression surface we had no coverage for: specialist prompts. Unit tests never call the LLM, so a one-line prompt tweak can silently flip direction distribution, kill contrast sensitivity, or unleash fabricated citations. This PR wires that gap shut — end to end, production-ready.

Builds on #1703 (grounded-packet archival).

### Pure prompt builder + golden snapshot (prerequisite refactor)
- **`trading_bot/specialist_prompt.py`** — pure \`build_specialist_prompt()\` extracted from \`TradingCouncil.research_topic\` Phase 2.
- **\`agents.py\`** now delegates to it; \`tests/test_specialist_prompt.py::test_golden_snapshot_legacy\` enforces byte-identical output so the refactor is provably behavior-preserving. All 1434 existing tests pass.

### Three-mode harness
- **\`evals/runner/harness.py\`**: \`mock\` | \`cached\` | \`live\`. Mock derives responses from the case \`expect:\` block — hermetic, no API keys, what CI runs. Cached SHA256-keys prompts for cheap replays. Live hits the real router.
- **\`evals/runner/case.py\`**: YAML + JSON Schema loader, \`check_case()\` assertion engine, regex-based hallucination detector (years / % / \$ tokens not traceable to the grounded packet).

### Quality gates
- **\`evals/runner/lint.py\`** rejects tautological \`direction_in: [all three]\`, empty assertions, inverted confidence bounds, singleton contrast cases, and contrast pairs with identical inputs or directions.
- **\`evals/runner/baseline.py\`** snapshots per-agent metrics and diffs against committed baselines. Severity bands on distribution drift (3pp/10pp) plus absolute thresholds on accuracy (-10pp), contrast (-15pp), and hallucination rate (+5pp). Red severity fails CI.

### CI integration
- **\`.github/workflows/eval-gate.yml\`** runs lint + mock-mode + baseline diff on every PR touching prompts, agents, or the harness. Posts the summary and diff back to the PR. Mock is free and hermetic — no secrets needed.

### Seed cases
- **\`evals/cases/agronomist/\`**: one canonical-bullish case + one contrast pair. Exercises every assertion kind end-to-end and gives downstream agent PRs a working template.
- **\`evals/baselines/agronomist.json\`** committed so CI has something to diff against on the first run.

## Operating the harness

\`\`\`bash
python -m evals.runner --lint-only                    # fast sanity check
[...]

PR #1731, opened 2026-04-23
fix(agronomist): price-direction sentiment rule + cache-key busts on override swap

rozavala:
## Summary

Two coupled fixes for eval-harness correctness on the agronomist persona. Replaces #1730 (wrong tier).

- **Tier-2 fix:** Add an explicit *Sentiment Rule* block to `AGRONOMIST_CONTEXT` in `trading_bot/prompts/base_prompts.py` — the template production and the eval harness actually hit when a commodity profile is present. Parallel structure to the existing `INVENTORY_CONTEXT` price mapping (\"Stocks DECREASING = BULLISH\").
- **Cache-key fix:** Fold `evals/config/eval_models.json` overrides into `_cache_signature`. `apply_eval_overrides` mutates `router.assignments` after the cache key is computed, so without this fix swapping a role's model silently returns the prior model's cached response.

## Why tier 2

`build_specialist_prompt` resolves persona prompts via three tiers:
1. DSPy-optimized (rarely set)
2. **Commodity-profile template** (`AGRONOMIST_CONTEXT` in `base_prompts.py`) — wins whenever a profile is present
3. Legacy `config.json` → `gemini.personas.agronomist` — only reached if `get_agent_prompt` raises

The eval harness (`evals/runner/harness.py:128`) always passes `commodity_profile=profile`, so tier 2 always wins on the eval path. Earlier PR #1730 edited tier 3 — inert on the eval path.

## Diagnostic evidence (motivates tier-2 fix)

11 agronomist cases, identical grounded data, three model families:

| Model family | Frame | Direction vs. `expect:` |
|---|---|---|
| Gemini 3 Flash | price-directional | 11/11 match |
| OpenAI gpt-4o-mini | crop-outcome | 11/11 inverted |
| Anthropic Haiku 4.5 | crop-outcome | 9/11 inverted |

Two of three families read the persona as crop-outcome direction. Committing a Flash baseline would embed a minority interpretation as ground truth.

## Why the cache-key fix ships with it

The cache-key bug was **discovered** during the diagnostic smoke: all 11 \"Haiku\" responses came back byte-identical to the gpt-4o-mini cache. Root cause: `apply_eval_overrides` mutates `router.assignments` after the cache key is computed, so `eval_models.json` swaps don't invalidate cached responses. Bundling this fix here makes the post-merge validation trustworthy — without it, a smoke-test \"all three models agree\" result could just mean all three hit stale cached entries.

## Regression tests

- `test_cache_signature_includes_eval_overrides` — changing overrides must produce a new cache key
- `test_cache_signature_stable_when_overrides_unchanged` — identical inputs produce identical keys
- `test_agronomist_template_emits_price_direction_sentiment_rule` — guards against future edits that drop the mapping

## Test plan

- [ ] `pytest tests/test_eval_harness.py` — 20 tests pass (17 existing + 3 new)
- [ ] `pytest tests/ -k 'specialist_prompt or base_prompts or prompt'` — 61 tests pass, no golden-snapshot breakage
- [ ] Post-merge: re-run agronomist smoke across Flash + gpt-4o-mini + Haiku. All three should agree on direction. Because the cache-key fix ships in the same commit, each model will now fetch fresh responses against the new prompt rather than hitting stale Flash responses.
- [ ] Proceed to 8-agent bootstrap after agronomist smoke passes all three families.

## Out of scope (separate follow-up)

- Pruning the legacy `config.json` → `gemini.personas` block. Tier 3 is a live fallback when `get_agent_prompt` raises — removing it needs its own design pass (telemetry for when tier 3 fires, default-message audit). Filing as a separate issue.


rozavala:
## Approving — ready to merge

Reviewed against PR review skill. Two coupled eval-harness correctness fixes, both clean.

### Scope verification
Only 3 files touched: `evals/runner/harness.py`, `tests/test_eval_harness.py`, `trading_bot/prompts/base_prompts.py`. No safety-critical files (order_manager, ib_interface, compliance, weighted_voting, signal_generator, risk_management, orchestrator) — confirmed via diff grep.

### Tier-2 analysis (sentiment rule)
PR description is accurate. `specialist_prompt._resolve_persona_prompt` applies the documented precedence: tier 3 (legacy `personas` dict) is overwritten by tier 2 (`get_agent_prompt` with profile) when `commodity_profile is not None`. The eval harness (`harness.py:133`) always passes `commodity_profile=profile` when the case has a ticker, so tier 2 is the live path. PR #1730's tier-3 edit was indeed inert. The fix correctly targets `AGRONOMIST_CONTEXT` in `base_prompts.py` with parallel structure to the existing `INVENTORY_CONTEXT` rules block. Directional vocabulary (BULLISH/BEARISH/NEUTRAL) matches CLAUDE.md convention.

### Canonical regression (#1704) — `canonical_bullish.001_minas_frost`
Frost at vulnerable phenology with Cooxupé exposure = "supply-damaging event". New rule maps it to BULLISH. Expected direction preserved — no regression.

### Cache-key fix
Key factor is the overrides dict itself (serialized via `json.dumps(default=str)` with `sort_keys=True`) — stable and serializable. Loader failure is caught with warning + empty dict fallback, so cache signatures remain computable even if the override file is malformed. Lazy init on first call, cached in `_eval_overrides_cached`. Pre-existing cache entries get a one-time miss + re-fill, which is the right behavior.

### Coupling justification
Bundled because the cache bug was discovered during the rule-fix diagnostic (all Haiku responses byte-identical to gpt-4o-mini cache). Shipping together makes the post-merge three-family validation trustworthy. Acceptable coupling.

### Tests
- `pytest tests/test_eval_harness.py tests/test_base_prompt_rules.py tests/test_specialist_prompt.py tests/test_eval_models.py` — 65/65 pass
- `pytest tests/ -k 'agronom or prompt or eval or agent'` — 181/181 pass, 0 failures

Regression tests are well-targeted: override-busts-cache, stable-when-unchanged, template-contains-rule-phrases.

No Brier/weighted-voting/agent-name churn. No CSV schema impact.

PR #1732, opened 2026-04-23
feat(personas): tier-2 price-direction rules for 4 personas

rozavala:
## Summary

Ports the agronomist fix pattern from PR #1731 to four more persona contexts: **sentiment**, **volatility**, **macro**, and **supply_chain** (logistics). Each now includes an explicit Sentiment Rule block that disambiguates the [SENTIMENT TAG] (forward price direction or forward IV direction) from the underlying fundamental/operational variable.

## Motivation

8-agent smoke run (after the agronomist fix merged) surfaced that four other personas suffer the same tag-inversion class of failure:

| Agent | Fails | DIR-inversion | Pattern |
|---|---|---|---|
| sentiment | 13 | 13 (100%) | extreme long tagged BULLISH instead of BEARISH |
| volatility | 14 | 14 (100%) | rich premium tagged BULLISH (cheap to buy directional) |
| macro | 8 | 5 | weaker BRL tagged BULLISH (ignored supply-push) |
| supply_chain | 10 | 5 | model says "prices downward" then tags BULLISH |

The model's [ANALYSIS] text is usually correct; only the [SENTIMENT TAG] line inverts. Same signature as the agronomist failure class, same fix.

Geopolitical (9 fails) is **excluded** — 8/9 are confidence overshoots, not direction inversions. That's a calibration concern with a different surface area; separate PR if 8-agent re-smoke flags it.

## What's in each rule

- **Sentiment**: positioning extreme → price direction (long crowded = BEARISH)
- **Volatility**: premium level → forward IV direction (rich = BEARISH for long premium)
- **Macro**: currency / DXY / growth → commodity price direction (weaker producer FX = BEARISH via supply push)
- **Supply_chain**: physical flow → pricing-point direction (constrained = BULLISH; expanded/debottlenecked = BEARISH)

No worked examples keyed to test-case facts (per prior feedback on not over-fitting persona prompts to specific cases).

## Test plan
- [x] `test_sentiment_template_emits_price_direction_sentiment_rule`
- [x] `test_volatility_template_emits_forward_iv_sentiment_rule`
- [x] `test_macro_template_emits_price_direction_sentiment_rule`
- [x] `test_logistics_template_emits_price_direction_sentiment_rule`
- [x] Full `tests/test_eval_harness.py` — 24 passed
- [ ] Post-merge: 8-agent smoke at default Flash tier; expect ≥90% direction accuracy with no single persona <80%, and no single direction dominating failure set

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1734, opened 2026-04-23
feat(prompts): tighten confidence calibration rubric

rozavala:
## Summary

PR-1732 cleared the direction-inversion failure class. The post-merge 8-agent smoke surfaced the next bottleneck: **~27 of the remaining 54 failures are confidence overshoots**, spread across 7/8 agents. Models pin to 0.9 or 0.95 regardless of evidence characteristics.

## What's changing

1. **`BASE_SYSTEM_INSTRUCTION` gets a `CONFIDENCE CALIBRATION` block** anchoring each band to evidence requirements:
   - EXTREME (~0.95): RESERVED for >=2 independent sources agreeing, zero counter-evidence
   - HIGH (~0.80): default for well-sourced single-source directional call
   - MODERATE (~0.55): offsetting factors / thin base / partial staleness
   - LOW (~0.30): mixed signals / anecdotal / stale
   - Anti-pin language: step DOWN on ambiguity, don't reflexively pick 0.90/0.95

2. **Inline JSON schema in `specialist_prompt.py`** — the last thing the model sees before generating — now references the rubric bands directly instead of the vague "Float based on data quality".

## Why this helps the smoke

Case caps for canonical cases sit at 0.85-0.92. HIGH (0.80) fits under every cap. If models default to HIGH on clean directional cases instead of EXTREME (0.95), the ~27 CONF_HI failures should collapse to zero. Expected post-merge pass rate: ~80% (56% → removing calibration failures, leaving direction residuals + hallucinations).

## Production impact

This ships to live agents (same path as #1731 / #1732). Three effects, all desirable:

- **Brier calibration improves.** Overconfidence is penalized harder than under-confidence; lower confidence on the same evidence tightens the expected error.
- **Weighted-voting input weights adjust proportionally.** Individual specialists influence the Master Strategist slightly less; the council hasn't changed, just the inputs are more honest.
- **Compliance / position sizing may shrink marginally** on the same signal set. Safer, not riskier.

No behavior regression.

## Test plan
- [x] `test_base_system_instruction_emits_confidence_calibration_rubric`
- [x] `test_specialist_prompt_inline_schema_references_calibration`
- [x] `test_golden_snapshot_legacy` updated to match new schema line
- [x] Full suite: 1506/1506 passed
- [ ] Post-merge: 8-agent smoke expects direction-accuracy unchanged (already clean), confidence_max failures near zero, residual = hallucinations + geopolitical NEUTRAL-avoidance


rozavala:
## Approval — clean follow-up to #1731/#1732 calibration push

Reviewed as part of the Brier-safety-critical prompt series. Approved with one forward-looking observation.

### Vocabulary safety — PASS
Confidence labels stay strictly within `{LOW, MODERATE, HIGH, EXTREME}`. No `VERY_HIGH` / `MEDIUM` / `UNCERTAIN` / similar slippage introduced. Regression tests `test_base_system_instruction_emits_confidence_calibration_rubric` and `test_specialist_prompt_inline_schema_references_calibration` lock the rubric in place.

### Scope
Live-production path — `trading_bot/prompts/base_prompts.py` feeds every specialist through `build_specialist_prompt()`, which the eval harness README explicitly calls out as "the same builder production uses". Immediate behavior flip for all agents once merged. That's consistent with how #1731/#1732 landed and aligns with the PR body (explicit acknowledgement of live impact, safe direction).

### Brier continuity — PRESERVED
- `trading_bot/brier_scoring.py` consumes `predicted_confidence: float` from JSON — unchanged schema.
- Agents.py L1201 canonicalizes via `parse_confidence()` before any downstream code (voting, trace, TMS) reads it.
- `weighted_voting._parse_agent_report()` regex `\[CONFIDENCE\]:\s*([0-9.]+)` still matches fine against numeric outputs and falls through cleanly on band strings.
- No reliability-multiplier mapping keyed off confidence bands is touched.

### Baseline drift risk — LOW
`evals/runner/baseline.py` gates on direction-distribution pp drift + absolute drops on `directional_accuracy` / `contrast_consistency` / `hallucination_rate`. There is no `mean_confidence` gate. Rubric tightens calibration magnitude, not direction, so expected drift-alert risk on merge is green. Baselines may still want a refresh post-merge once cached-mode reruns settle, but that's a follow-up not a blocker.

### Test results
- `tests/test_brier_*.py tests/test_eval_*.py tests/test_specialist_prompt.py`: 131/131 pass.
- `tests/test_base_prompt_rules.py`: 15/15 pass.
- Full suite (minus 2 optional files): 1495 passed, 0 failed, in 355s.

### Forward-looking note (not blocking)
Band-anchor mismatch worth tracking in a follow-up: `trading_bot/confidence_utils.CONFIDENCE_BANDS` still maps `LOW=0.55, MODERATE=0.65, HIGH=0.80, EXTREME=0.90`, while this rubric anchors `LOW~0.30, MODERATE~0.55, HIGH~0.80, EXTREME~0.95`. Matters only when an LLM emits a band *string* (instead of a float per the JSON schema) — in that edge case `parse_confidence('LOW')` returns 0.55, which is 25pp above the rubric's intent. The JSON schema keeps this as an edge path rather than the primary one, and the inconsistency predates this PR (#1654 set the Python mapping before the rubric was tightened). Recommend a cleanup PR that either reconciles `CONFIDENCE_BANDS` to the rubric anchors or documents the two as deliberately different mappings.

Shipping.

PR #1737, opened 2026-04-23
chore(evals): narrow lane-guard must_not_cite canaries to phrases

rozavala:
## Summary

- Replaces broad single-word \`must_not_cite\` tokens (e.g. \`Black Sea\`, \`drought\`, \`frost\`, \`Minas\`, \`strike\`, \`harvest\`) with targeted multi-word phrases that capture the actual lane-drift each case is designed to catch.
- Prevents false-positive hallucination flags when the model legitimately references a corpus-present token without pivoting its thesis into the blocked lane.
- 9 case YAMLs touched across macro (3), agronomist (2), sentiment (2), volatility (4), supply_chain (1). No logic changes.

## Motivation

The baseline cached-mode run (#1735) flagged 0.38 hallucination on agronomist and 0.17 on inventory. A chunk of those are not real fabrications — the detector at \`evals/runner/case.py:405-412\` uses bare \`substring.lower() in evidence.lower()\` with no awareness of corpus content, so any single-word canary that also appears in the case's own \`raw_findings\` / \`market_context\` gets cited-and-fired as a hallucination even when the model is correctly paraphrasing the prompt.

Example: \`sentiment.canonical_bearish.001_kc_crowded_long_extreme.yaml\` contains \`"KC has rallied +6.1% over the past 5 sessions on Minas Gerais frost headlines"\` in \`raw_findings\` but blocks bare \`"frost"\` and \`"Minas"\`. Any acknowledgment of the backdrop — even one that correctly stays in the positioning lane — fires the canary.

## Approach

Audit against the full content the model sees (\`search_instruction\` + \`raw_findings\` + \`extracted_facts\` + \`market_context\`, not just \`raw_findings\`). For tokens that appear in corpus but have a clearly-stated lane-drift intent in the case's \`notes\`, replace with phrases that only fire when the model pivots its thesis:

| Old | New |
| --- | --- |
| \`Black Sea\` | \`Black Sea disruption\`, \`Black Sea tension\`, \`Black Sea supply shock\`, \`Russian competition\` |
| \`drought\` | \`drought concern\`, \`drought stress\`, \`drought worsening\` |
| \`frost\` | \`frost damage\`, \`frost risk\`, \`frost bullish\` |
| \`Minas\` | \`Minas drought\`, \`Minas frost\` |
| \`strike\` | \`strike risk\`, \`strike imminent\`, \`strike disruption\` |
| \`force majeure\` | \`force majeure declared\`, \`force majeure risk\` |
[...]

PR #1735, opened 2026-04-23
chore(evals): cached-mode baselines post-calibration (66% / acc ≥88% per agent)

rozavala:
## Summary

Seeds `evals/baselines/cached/` with the first cached-mode baseline snapshot over all 154 cases on main at 29cd62e24 (post-#1732 persona audit + #1734 calibration rubric + #1733 canonical_neutral). The existing `evals/baselines/*.json` (mock-mode corpus-shape baselines consumed by the per-PR mock gate) are intentionally **untouched**.

The cached subdir is reference-only in this PR — wiring `--baselines-dir evals/baselines/cached` into the scheduled cached-mode lane is a separate PR.

> **Note:** The first commit (`64c78de3d`) on this branch overwrote the root baselines with cached-mode metrics and blocked the mock gate (cached-mode distribution vs. case-corpus distribution = structural drift on three agents). Second commit (`418c0d1c2`) restores the root baselines to their state on main and relocates the cached metrics to the subdir.

## Cached-mode headline metrics (reference)

| Agent | Cases | Passed | Dir. Acc | Abstention | Contrast | Mean Conf | Halluc. |
|---|---|---|---|---|---|---|---|
| agronomist | 14 | 6 | 0.88 | 0.14 | 1.00 | 0.79 | **0.38** |
| geopolitical | 20 | 11 | 0.92 | 0.25 | **0.75** | 0.81 | 0.00 |
| inventory | 20 | 17 | 0.92 | 0.15 | 1.00 | 0.75 | 0.17 |
| macro | 20 | 11 | 0.92 | 0.20 | 1.00 | 0.78 | 0.00 |
| sentiment | 20 | 13 | 0.92 | 0.15 | 1.00 | 0.80 | 0.00 |
| supply_chain | 20 | 15 | 0.92 | 0.30 | 1.00 | 0.76 | 0.00 |
| technical | 20 | 18 | 1.00 | 0.30 | **0.75** | 0.74 | 0.00 |
| volatility | 20 | 10 | 0.92 | 0.25 | 1.00 | 0.79 | 0.08 |

**Overall: 101/154 passed (66%). Directional accuracy ≥ 88% on every agent.**

Compared to the pre-#1732 failure set, direction inversions dropped from ~40% down to the single-digit per-agent range; mean confidence dropped across the board (calibration rubric working as intended).

### 1. Tier-resistance hallucinations (agronomist 0.38, inventory 0.17, volatility 0.08)
[...]

PR #1738, opened 2026-04-23
chore(evals): tolerate ±1pp rounding in hallucination percentage check

rozavala:
## Summary

- Hallucination detector currently does exact-substring match on every \`NN%\` token in the agent's output vs the grounded corpus.
- This flags legitimate rounding paraphrases (corpus \`71% good/excellent\` → model writes \`~70%\`) as fabrications.
- Adds a ±1 percentage-point tolerance: if the exact substring isn't in the corpus, parse the value and check whether any corpus-present percentage is within 1pp.
- Still catches drifted values — \`71%\` in corpus → model writes \`85%\` stays flagged (covered by new test).

## Motivation

Cached-mode baselines (#1735) show residual hallucination rates:

| Agent | Rate |
| --- | --- |
| agronomist | 0.375 |
| inventory | 0.17 |
| volatility | 0.08 |

Spot-audit of the agronomist miss set shows the miss pattern is dominated by small rounding drift (e.g., a case with \`71% G/E\` gets the model writing \`around 70%\` in evidence → currently flagged as fabrication). With ±1pp tolerance, these clean-paraphrase cases stop counting against the metric, leaving the headline hallucination_rate to reflect actual invented numbers.

## Approach

- \`evals/runner/case.py\`: added \`_parse_percentages\` helper and a tolerance branch in the percentage loop of \`_detect_hallucination\`. Exact-match fast path is preserved; tolerance is the fallback before flagging.
- \`tests/test_eval_case.py\`: two new tests (\`within_one_point_tolerated\`, \`outside_tolerance_flagged\`) document the contract. All 21 tests pass.

## Test plan

- [x] Full \`tests/test_eval_case.py\` passes locally.
- [ ] Mock gate stays green on this PR (expected — no case YAML changes, detector change is strictly more permissive).
- [ ] After merge, re-snapshot cached-mode baselines to quantify the hallucination_rate drop. If the residual is still meaningful after this + the lane-guard narrowing (#1737), revisit with model-swap experiments per the tier-resistance idea.


rozavala:
Approving — clean, well-scoped change.

**Scope:** Only `evals/runner/case.py` (+27) and `tests/test_eval_case.py` (+34). No safety-critical files touched (order_manager, ib_interface, compliance, weighted_voting, signal_generator, risk_management, orchestrator, brier_scoring).

**Tolerance logic (correct):**
- This is the per-case *hallucination detector*, not the baseline drift comparator. The ±1pp tolerance is symmetric grounding tolerance: an output `70%` is accepted iff the corpus contains any percentage within 1pp. That matches the rounding-paraphrase intent.
- Zero-corpus-percentage case handled correctly: `_parse_percentages(corpus)` returns `[]`, `any(... for c in [])` returns `False`, so the token is flagged. Fail-closed preserved.
- Exact-match fast path preserved before the tolerance branch — no behavior change for already-grounded tokens.
- The baseline drift comparator (`baseline.py:239`) remains direction-aware via `direction_of_concern=\"rise\"` / `HALLUCINATION_RISE_FAIL_PP`. This PR tightens the detector (fewer false positives from rounding) without touching that pathway, so it can't silently mask rising hallucination at the baseline level.

**Test coverage:** Two new tests pin the contract — `within_one_point_tolerated` (corpus 71%, model 70% → accepted, exercises the inclusive `<=1.0` boundary) and `outside_tolerance_flagged` (corpus 71%, model 85% → flagged). All 57 eval tests pass (test_eval_case + test_eval_baseline + test_eval_harness), full `tests/ -k eval` suite green (83 passed).

**Consistency:** Years (20xx) and dollar amounts remain exact-match. That's justified — rounding paraphrase is a percentage-specific phenomenon.

Ready to merge.

Issue #1694, opened 2026-04-21
investigate: IV unavailable during VaR compute — 13 fallbacks today on deep-OTM options


rozavala:
## Context

Audit of 2026-04-21 logs (\`orchestrator_multi.log\`) shows 13 occurrences of \`IV unavailable\` across multiple VaR compute cycles. Each case triggers fallback to the commodity profile's \`fallback_iv\` (0.30 for SB, 0.35 for KC), which is then used in VaR Monte Carlo.

### Top offenders by conId (today only)
| conId | Local symbol | Commodity | Occurrences |
|---|---|---|---|
| 796568540 | SON6 P0.1375 | SB (Sugar) | 3 |
| 822566312 | SON6 P0.1275 | SB (Sugar) | 2 |
| 812787480 | SON6 P0.1325 | SB (Sugar) | 2 |
| 729177412 | KOZ6 P2.625 | KC (Coffee) | 1 |
| 694702605 | SOH7 P0.15 | SB (Sugar) | 1 |
| 675726623 | KOZ6 P2.25 | KC (Coffee) | 1 |

### Frequency
- 4× \`IV fetch timed out after 5.0s\` with \`N/12 tickers still pending\` (N = 3 or 4)
- 13× \`IV unavailable — using fallback IV=0.30/0.35\`

Same 3 SON6 put strikes (0.1275, 0.1325, 0.1375) fail repeatedly across the day.

## Source

\`trading_bot/var_calculator.py:562-572\`:
\`\`\`python
# Use fallback IV from commodity profile if market IV unavailable
if iv_val is None or iv_val <= 0:
    try:
        profile = get_commodity_profile(c.symbol)
        iv_val = profile.fallback_iv
    except (ValueError, KeyError):
        iv_val = 0.35
    logger.warning(
        f"IV unavailable for {c.localSymbol} (conId={con_id}) "
        f"— using fallback IV={iv_val:.2f}"
    )
\`\`\`

Timeout source (\`var_calculator.py:700-718\`):
\`\`\`python
timeout = config.get('compliance', {}).get('iv_fetch_timeout', 5.0)
\`\`\`

## Root-cause hypotheses

1. **Deep-OTM illiquidity**: the failing strikes (SON6 P0.1375, P0.1325, P0.1275) are deep out-of-the-money puts with low quote activity → IB's model IV isn't computed because there's no recent trade/quote.
[...]

PR #1744, opened 2026-04-23
fix(ib-interface): poll for IV up to 5s instead of fixed 2s sleep (#1694)

rozavala:
## Summary

- `get_option_market_data` used a fixed `await asyncio.sleep(2)` before reading `modelOptionImpliedVol` once. Deep-OTM strikes (SON6 P0.1275, P0.1375) routinely need more than 2s for IB to compute model IV → fell back to profile default (35% for SB), skewing theoretical pricing for the pre-trade spread-quality gate.
- PR #1694 (78eaa76a4) already fixed the parallel VaR-compute path via smile-neighbor fallback. This addresses the residual hits on a different call path.
- Replace the fixed sleep with a polling loop that checks every 0.5s for up to 5s, exiting early when IV + bid + ask are all populated. Fallback warning now includes elapsed poll duration for observability.

## Behavior change

- **Liquid options**: return on the first poll (~0.5s) instead of waiting a full 2s. Net latency win.
- **Illiquid strikes** (deep-OTM): get up to 5s before falling back, matching `var_calculator.py`'s `iv_fetch_timeout=5.0`.
- **Worst case** (IV never arrives): +3s per call vs. the old 2s wait. For the ~2 occurrences/day we've been seeing, this is a negligible throughput cost in exchange for receiving real IV when it eventually arrives.

## Test plan

- [x] New `TestGetOptionMarketDataPolling` class with three tests: early-exit, fallback-on-timeout, `None`-return when bid/ask missing.
- [x] Full suite: **1511 passed, 0 failed** (390s).
- [ ] Post-merge: monitor `IV data missing` log rate. Expect further reduction from the 2/day seen on 2026-04-23 once deep-OTM strikes get the full 5s window.

## Context

Root cause investigation in #1694. Today's recurrence comment (2026-04-23) identified the `ib_interface.py:156` path as the source of today's 2 fallback warnings, distinct from the `var_calculator.py` path fixed by the prior PR.


rozavala:
## Approved — clean and well-targeted fix

### Verification
- ✅ `pytest tests/test_ib_interface.py -x -q --timeout=120` → 23 passed
- ✅ Core logic is bounded: `while True` gated by `loop.time() >= deadline` where `deadline = poll_start + IV_POLL_TIMEOUT (5.0s)` — no infinite-loop risk
- ✅ Early-exit correct: exits when `iv is not None AND bid is not None AND ask is not None` (all three)
- ✅ Fail-closed: `bid is None or ask is None` → returns `None` (unpriceable); IV missing after poll → logs WARNING with elapsed duration and falls back to `profile.fallback_iv` (35% for SB, 30% for KC, etc.)
- ✅ `iv is None and ...` guard in the loop prevents stale-IV overwrite after first successful read
- ✅ `loop.time()` is monotonic (wraps `time.monotonic()`), so `elapsed` is wall-clock safe
- ✅ Two code paths genuinely separate: `ib_interface.get_option_market_data` (ib_interface.py:432, called by `create_combo_order_object` for pre-trade theoretical pricing) is distinct from `var_calculator._batch_fetch_iv` (var_calculator.py:555, called by VaR Monte Carlo)

### Behavior-change math
- Liquid options: ~0.5s (vs. 2s prior) — net latency win
- Deep-OTM strikes: up to 5.0s (matches profile-fallback semantics at `var_calculator.py:740`'s `iv_fetch_timeout=8.0` default — close enough; see follow-up suggestion below)
- Worst case: +3s per call when IV genuinely never arrives — small throughput cost for actual IV accuracy

### Minor follow-up (non-blocking)
`IV_POLL_TIMEOUT = 5.0` is hardcoded, while `var_calculator.py` reads `config.compliance.iv_fetch_timeout` (default 8.0). Consider harmonizing in a follow-up so both paths share the config knob. Not a blocker for merge — the current 5.0 is safe and the PR description correctly justifies the tradeoff.

Approving.

PR #1745, opened 2026-04-23
feat(orderref): prefix orderRef with ticker for engine-attributed tagging (#1670 Step 1)

rozavala:
## Summary
- Step 1 of #1670's Phase 2 refactor: replace raw `uuid4()` orderRefs with a ticker-prefixed helper so DISCONNECT_GUARD can filter orders by `f"{ticker}-"` prefix instead of clientId.
- Additive only — no behavior change yet. Step 2 (DISCONNECT_GUARD flip behind feature flag) and Step 3 (slot allocator) land in follow-up PRs.
- Also fixes two sites that were placing orders with no `orderRef` at all.

## Design
Format: `{TICKER}-{20 hex}` (e.g. `KC-a1b2c3d4e5f60718a9b2`, 23 chars).

Chose `hex[:20]` specifically to:
- Stay well under IB Gateway's 40-char `orderRef` cap, leaving headroom for the `CATASTROPHE_` inheritance prefix used by catastrophe-close paths (`CATASTROPHE_KC-…` = 35 chars).
- Preserve the `len(orderRef) > 20` heuristic in `utils._generate_position_id_from_trade` (`utils.py:565`). Dropping below 20 chars would silently break position tracking.
- Keep 80 bits of entropy — more than enough for per-session uniqueness.

## Changes
- `trading_bot/ib_interface.py`: add `_make_order_ref(ticker)` helper; replace 3 `uuid4()` origination sites (lines 781, 813, 849).
- `trading_bot/risk_management.py:194`: add missing `orderRef` on the `MarketOrder` in the position-close path.
- `trading_bot/order_manager.py:4744`: add missing `orderRef` on the retry close path.
- `tests/test_ib_interface.py`: new `TestMakeOrderRef` class with 6 tests — format, uppercase, empty→`UNK`, uniqueness (500 refs), 40-char cap (incl. `CATASTROPHE_` prefix), heuristic-threshold regression guard.

## Test plan
- [x] Isolated: `pytest tests/test_ib_interface.py` — 28 passed
- [x] Full suite: `pytest tests/` — 1516 passed, 11 warnings
- [ ] Staging smoke: confirm orders land on IBKR with ticker-prefixed `orderRef` on a DEV cycle

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1670, opened 2026-04-20
refactor: fixed-pool IB connection architecture with orderRef tagging (Phase 2 of 2)


rozavala:
## Context

Phase 1 (#1669) consolidates 33 per-purpose sockets down to 10 per-engine sockets (2 per engine + 2 master). It ships quickly and solves the 2026-04-20 thundering-herd pattern, but it still scales **linearly** with commodity count:

| Commodities | Phase 1 sockets | Gateway limit headroom |
|---|---|---|
| 4 (today) | 10 | ~22 spare (default `MaxClientsPerAccount=32`) |
| 9 (4+5 new) | 20 | ~12 spare |
| 15 | 32 | **at limit** |

Phase 2 decouples socket count from engine count, so adding the 5th, 10th, or 20th commodity adds **zero** new sockets.

## Problem statement

Per-engine sockets are a convenience, not a requirement. The only load-bearing reasons for physical socket separation:

1. **Order ownership tracking** — `DISCONNECT_GUARD` filters `trade.order.clientId != my_client_id` to know which orders to drain (`connection_pool.py:461`). Comment there explicitly notes: *"attempting to cancel another engine's orders causes Error 10147 and 120s blocking stalls."*
2. **Fault isolation** — if a socket dies, you don't want all engines to stall simultaneously.

Both are solvable without per-engine sockets:

- **Order ownership** → tag every order with `order.orderRef = "{ticker}-{cycle_id}-{leg}"`; DISCONNECT_GUARD filters by `orderRef` prefix instead of `clientId`.
- **Fault isolation** → small fixed pool (e.g., 2 read + 2 order sockets) with round-robin + liveness failover. One dying socket is absorbed by its twin.

## Proposed architecture

```
Read pool:   2 shared sockets (round-robin across all engines)
Order pool:  2 shared sockets (orderRef-tagged, engine-agnostic)
Master:      1 socket (equity_service + monitor)
-----------------------------------------------------------
Total:       5 sockets — independent of commodity count
```

At 4 engines: 5 sockets. At 10 engines: 5 sockets. At 30 engines: 5 sockets.

### 1. orderRef tagging (`trading_bot/order_manager.py`)

Every order gets a unique `orderRef` at placement:

```python
def tag_order(order, ticker: str, cycle_id: str, leg_index: int):
    \"\"\"Attach traceable orderRef for post-hoc attribution.\"\"\"
    order.orderRef = f\"{ticker}-{cycle_id}-{leg_index}\"
[...]

PR #1748, opened 2026-04-23
feat(orderref): DISCONNECT_GUARD orderRef filter behind flag (#1670 Step 2)

rozavala:
## Summary

Step 2 of #1670's Phase 2 refactor. Step 1 (#1745) made every order carry a ticker-prefixed `orderRef`; this PR adds the drain-path switch so `DISCONNECT_GUARD` can filter by that prefix instead of `clientId`.

- Default behavior unchanged: `connection_pool.use_orderref_filtering` defaults to `false`, so production continues on the clientId path.
- Even with the flag off, the guard now computes **both** filters per non-terminal trade and logs any disagreement (`DISCONNECT GUARD [KC_orders]: filter divergence — clientId-only ... orderRef-only ...`). That's the forensic trail the issue asks for during the 3-trading-day validation window before flipping the flag.
- Master-level pool keys (`master_orders`, `dashboard_orders`, `dashboard_close`, `monitor`) have no ticker to anchor on, so they keep using clientId even when the flag is on.

## Prefix shapes recognized

Matches the three orderRef shapes emitted in this codebase today:

| Shape | Example |
|---|---|
| `{TICKER}-{hex}` | `KC-a1b2c3d4e5f60718a9b2` |
| `CATASTROPHE_{TICKER}-{hex}` | `CATASTROPHE_KC-a1b2c3...` |
| `CATASTROPHE_CLOSE_{localSymbol}` | `CATASTROPHE_CLOSE_KCZ6 C400` (localSymbol starts with the ticker) |

## Rollout per issue body

1. **This PR (default off)** — ships the machinery + divergence logging. No behavior change.
2. **DEV flip** — set `use_orderref_filtering: true` in DEV config, validate drain logs match clientId-path behavior for 3 trading days.
3. **Divergence audit** — grep for `filter divergence` warnings in prod logs. Zero disagreements is the gate to flip PROD.
4. **PROD flip** — later change once DEV is clean.

## Test plan

- [x] `pytest tests/test_connection_pool.py` — all 74 tests pass (run locally, 12 min)
- [x] Unit: pool-key ticker extraction for engine keys (`KC_orders` → `KC`) and master keys (→ `None`)
- [x] Unit: orderRef match for raw, `CATASTROPHE_`, `CATASTROPHE_CLOSE_` shapes (incl. negative cases)
- [x] Unit: flag off → clientId behavior preserved exactly; flag on → foreign clientId with matching ticker gets drained, foreign ticker on our clientId is skipped
- [x] Unit: terminal-state orders skipped under both filters
- [x] Unit: divergence log fires when filters disagree, stays quiet when they agree, skipped for master keys
- [x] Unit: config flag read is tolerant to missing key / load errors
- [ ] Post-merge on DEV: flip flag, tail logs for 3 trading days, confirm zero `filter divergence` entries before recommending PROD flip

Closes none (issue #1670 stays open through Step 3/4).


rozavala:
## Review: fix needed before merge

Step 2 logic is well-designed (feature flag default-off, divergence telemetry captured even when flag is off, correct prefix match handling empty/None, master-key fallback, no prefix collisions among current tickers). But **the new tests don't actually pass** under a normal CI timeout — they hang individually for 120 s each.

### Blocker: 6 new tests hang for `DISCONNECT_GUARD_TIMEOUT` seconds

`_drain_pending_orders` (connection_pool.py:776–790) waits up to `DISCONNECT_GUARD_TIMEOUT = 120 s` for any cancelled trade to reach a terminal state. Every new test that enqueues at least one "owned" non-terminal trade uses a plain `MagicMock` whose `orderStatus.status` stays `'Submitted'` forever, so the drain loop burns the full 120 s deadline on every such test:

| Test | Why it hangs |
|---|---|
| `test_flag_off_preserves_clientid_behavior` | `mine` is ours by clientId, stays `'Submitted'` |
| `test_flag_on_uses_orderref_prefix` | `foreign_cid_kc_ref` matches by orderRef, stays `'Submitted'` |
| `test_catastrophe_prefixes_drain_under_orderref` | two CATASTROPHE orders, both stuck |
| `test_master_pool_key_always_uses_clientid` | `mine` matches by clientId, stuck |
| `test_divergence_logged_when_filters_disagree` | `clientid_only` + `both` are drained, stuck |
| `test_no_divergence_log_when_all_agree` | `mine_and_tagged` is drained, stuck |

That's 6 × 120 s = 12 min — matching the PR body's "12 min" run time. The tests only 'pass' locally because the soft deadline expires; `pytest --timeout=120` (a reasonable CI default) fails them:

```
FAILED tests/test_connection_pool.py::TestDrainPendingOrdersOrderRefSwitch::test_flag_off_preserves_clientid_behavior
1 failed, 65 passed, 1 warning in 122.15s
```

For a live-trading safety-critical refactor this is a blocker — the tests need to actually exercise the branches they claim to, deterministically, in seconds.

### Fixes (pick one)

**Option A (preferred) — make `cancelOrder` flip the status to terminal** so the drain loop exits on its first tick:

```python
@staticmethod
def _make_ib(my_client_id: int, trades):
    ib = MagicMock()
    ib.client.clientId = my_client_id
    ib.openTrades.return_value = trades
    def _cancel(order):
        for t in trades:
            if t.order is order:
                t.orderStatus.status = 'Cancelled'
    ib.cancelOrder = MagicMock(side_effect=_cancel)
    return ib
```

**Option B** — patch the timeout on the class for the duration of each test:

```python
with patch.object(IBConnectionPool, 'DISCONNECT_GUARD_TIMEOUT', 0):
    ...
```

(Slightly worse because it bypasses the wait loop entirely, so the "await reaches terminal" branch gets no coverage.)

### Nit (optional)

`_order_ref_matches_ticker` is safe today because `{KC, CC, SB, NG, ZW}` have no shared prefixes, but `CATASTROPHE_CLOSE_KC` would `startswith`-match any future ticker starting with `KC` (e.g., a hypothetical `KCA`). A cheap hardening would be:

```python
if order_ref.startswith(f"CATASTROPHE_CLOSE_{ticker}") and (
    len(order_ref) == len(f"CATASTROPHE_CLOSE_{ticker}") or
    order_ref[len(f"CATASTROPHE_CLOSE_{ticker}")] in " 0123456789"
):
```

— or just use the raw contract's ticker via the ib_interface instead of localSymbol-parsing. Not a blocker for current tickers; worth a comment at minimum.

### What's right

- Flag defaults to `false` (config.json:63–65) — good.
- Flag-off path: divergence log runs, cancel path unchanged — good.
- `_pool_key_ticker` correctly returns `None` for master keys → always clientId fallback.
- `_order_ref_matches_ticker` handles empty/None, is case-sensitive consistent with `_make_order_ref` upper-casing, boundary-correct for the `{TICKER}-` and `CATASTROPHE_{TICKER}-` shapes.
- `_use_orderref_filtering` fails closed to `False` on config exception.
- Rollback story is clean: flip one config key.

### Action

Adding `claude-fix` label — fix the test hangs (Option A above), and ideally the nit.

rozavala:
Thanks for the careful review — both issues fixed in 12d27e88e.

**Blocker (test hang):** Your diagnosis was exactly right. `MagicMock.cancelOrder` did nothing, so `orderStatus.status` stayed `'Submitted'` forever and the drain loop burned all 120 s on every test. Applied Option A — `cancelOrder` side_effect now flips the matching trade to `'Cancelled'`:

```python
def _cancel(order):
    for t in trades:
        if t.order is order:
            t.orderStatus.status = 'Cancelled'
ib.cancelOrder = MagicMock(side_effect=_cancel)
```

Verification: `pytest tests/test_connection_pool.py --timeout=60` → **76 passed in 3.92 s** (was 12 min under the soft timeout). That's a real functional exercise of the drain loop's terminal-state exit path, not a timeout expiry.

**Nit (CATASTROPHE_CLOSE shadowing):** Agreed the latent hazard is worth closing structurally rather than commenting around. Swapped the naive `startswith` for a longest-prefix match against `COMMODITY_ID_OFFSET`:

```python
if order_ref.startswith(IBConnectionPool._CATASTROPHE_CLOSE_PREFIX):
    local_sym = order_ref[len(IBConnectionPool._CATASTROPHE_CLOSE_PREFIX):]
    best = ""
    for t in COMMODITY_ID_OFFSET:
        if local_sym.startswith(t) and len(t) > len(best):
            best = t
    return best == ticker
```

Two new tests cover it: `test_catastrophe_close_longest_prefix_wins` (injects a temporary `KCA` offset and asserts `KCA` wins, `KC` rejects) and `test_catastrophe_close_unknown_symbol_matches_nothing` (orphaned prefix drains nowhere). If a future commodity profile ever introduces overlapping tickers, the disambiguation is structural and tested — no silent collisions.

rozavala:
## Re-review on 12d27e88e — fix verified, ready to merge

Both review items addressed cleanly:

1. **Test hangs fixed (Option A — preferred path).** `_make_ib`'s `cancelOrder` now flips the targeted trade's status to `'Cancelled'` via `side_effect`, so the drain loop exits on its first ~0.5s tick instead of burning the full 120s `DISCONNECT_GUARD_TIMEOUT`. The await-terminal branch is still exercised, just with a fast-terminating mock. `tests/test_connection_pool.py -q --timeout=30` → **76 passed in 3.99s** (was 12 min hanging).

2. **CATASTROPHE_CLOSE prefix-collision hardened.** `_order_ref_matches_ticker` now does a longest-prefix lookup against `COMMODITY_ID_OFFSET` for the `CATASTROPHE_CLOSE_*` shape, so a hypothetical future `KCA` ticker can't be shadowed by `KC`. Two new tests cover the shadowing case (with a temporary `COMMODITY_ID_OFFSET` injection) and an unknown-symbol fallthrough.

**Full suite:** `pytest tests/ -x -q --timeout=120` → **1538 passed in 368.83s**, no regressions.

Default-off flag, divergence telemetry, master-key fallback, and rollback story all unchanged from prior review. LGTM — swapping `claude-fix` → `Ready to be merged`.

PR #1757, opened 2026-04-24
evals: accept derived percentages on unit-qualified values

rozavala:
## Summary

Narrows what the hallucination detector treats as a fabrication. Accepts a derived percentage when the model shows its arithmetic — a relational phrase + two unit-qualified values in the ±120-char window whose ratio matches within ±1pp. Keeps rejecting pure assertions and ungrounded claims.

Adds one detector helper (`_percentage_is_derived`, ~90 lines incl. docstring) and four new unit tests in `tests/test_eval_case.py` (~150 lines).

## The false-positive this fixes

Before this PR, the detector flagged correct analytical derivations. Examples from the 2026-04-24 cached bootstrap:

- `geopolitical/kc_summit_quota_rumor`: model wrote *"3.5M bag limit is approximately 9.3% higher than the historical 3.2M average"*. 9.3% isn't in the packet literally, so flagged — even though it's exactly (3.5−3.2)/3.2.
- `inventory/certified_drawdown`: *"surplus has collapsed from 2.1M to 0.4M bags — a 81% reduction"*. 81% flagged, even though it's (2.1−0.4)/2.1.
- `inventory/ng_draw_pace`: *"from −80 Bcf to −280 Bcf in current week, a 250% acceleration"*. 250% flagged, even though it's (280−80)/80.

These aren't hallucinations. The packet's purpose is raw numbers; computing ratios is the analysis.

## Design: unit-gated, phrase-anchored

Two structural guards prevent the classic whitewash trap:

### 1. Relational marker required in the window

Words like `higher`, `lower`, `above`, `below`, `reduction`, `premium`, `surplus`, `deficit`, `acceleration`, `of`, `from`, `to`, `vs`, etc. If the ±120-char window around the percentage contains none of these, the derivation check doesn't fire. Standalone number assertions stay flagged.

### 2. Candidate values must carry a recognized unit

A number is a derivation input only if followed by a real quantity unit (`%`, `million`, `billion`, `Mt`, `Bcf`, `bpa`, `bushels`, `cents`, `c/lb`, `tonnes`, `days`, `weeks`, `bags`, `hours`, `percent`, or `M`/`K`/`B` with word boundary). Compound-noun descriptor numbers like `5-year average` or `3-month period` fail the unit check — the `5` there isn't quantifying anything.

### Why (2) is load-bearing

Without unit gating, the check whitewashes this real fabrication:

> agronomist/004 output: *"Kansas Wheat Commission's yield estimate — which is 28% below the 5-year average — points to a substantial supply contraction."*

The packet actually says yield is **+14.6% ABOVE** average. The model flipped the sign. But a naive ratio-match finds that the `5` in `5-year` and an unrelated `18` from a nearby G/E-ratings clause (`"ratings (18% G/E vs 52% last year)"`) coincidentally give 5/18 ≈ 27.8%, which matches the fabricated `28%`. That's whitewash of a real error.

Unit gating rejects `5-year` (no unit), leaving only one usable number in the window → no pair → derivation not licensed → claim correctly stays flagged.

## Four load-bearing tests

All four encode the design contract and are committed *with* the detector change. Any regression on these means the design has drifted:

| Test | What it guards |
|---|---|
| `test_hallucination_sign_flip_one_sentence_number_flagged` | Minimal "28% below 5-year" shape stays flagged |
| `test_hallucination_two_number_derivation_allowed` | "3.5M is 9.3% higher than 3.2M" passes |
| `test_hallucination_pure_fabrication_without_packet_numbers_flagged` | Raw "47% loss" assertion stays flagged despite relational marker |
| `test_hallucination_adjacent_clause_numbers_do_not_license_derivation` | **Realistic agronomist/004 output** — 18/52 in adjacent clause must not coincidentally whitewash the 28% yield claim |

The last test specifically encodes the realistic-output version of the whitewash bug. If a future change relaxes unit gating or expands the window, this test catches it.

## Pre-registered prediction vs actual

Pre-registered: **+6 flips to PASS, 0 flips to FAIL, agronomist/004 stays failed**.

Actual (from `/tmp/eval_bootstrap_step3_v2.log` run on this branch):

- ✓ `geopolitical/kc_summit_quota_rumor` (9.3% from 3.5M / 3.2M)
- ✓ `inventory/certified_drawdown` (81% from 2.1M / 0.4M)
- ✓ `inventory/ng_draw_pace_007_b` (250% from 80 / 280 Bcf)
- ✓ `agronomist/004` correctly **stays failed** on `forbid_hallucination`
- ✗ 3 cases predicted to flip but didn't: `inv/sb_build_misses` (0.85 outside window — model didn't repeat the base value in the 41% sentence), `inv/ng_storage` (only 2 of 3 halluc tokens are derivable; case still fails on the 20%), `inv/surplus_build` (model's 6% derives from 2.7M / 44.5M, but 44.5 isn't in the ±120 window)

Honest downgrade: **3 flips, not 6**. These misses are cases where the model didn't fully show its work within the ±120-char window of the percentage. The stricter check prefers false negatives (missed derivations the model didn't explicitly demonstrate) over false positives (whitewashing coincidental matches). Acceptable trade-off.

**Pass rate**: 197/247 (79.8%) → 200/247 (81.0%), +1.2pp from the three flips.

### Side observation (unrelated to this PR)

The verification run also shows `macro.contrast.kc_fed_magnitude_007_a` flipping back from PASS to FAIL on `direction`. This is the stochastic macro cache-churn I flagged in the Step-1 verification — 7 macro cases have hidden cache-signature instability that triggers fresh LLM calls between runs. This case has flipped direction NEUTRAL ⇄ BEARISH between runs with no code changes. Not caused by this PR; worth a separate investigation (probably a dynamic element in macro prompt building like timestamp or FRED current-rate lookup).

## Scope out

- **4 agronomist Pattern A cases** still fail (`Minas = 30% of production` training-data completions, `2027` year extrapolation) — not derivations, real fabrications. Deserve a separate targeted fix (agronomist persona hardening or model swap).
- **3 cases where the model didn't show full work** stay flagged even though the derivation was mathematically correct. Could be addressed by broadening the window or adding a second-order "compute from evidence list" lookup, but that opens the whitewash door wider. Explicitly deferred.

## Test plan

- [x] `pytest tests/test_eval_case.py -v` → 25 passed (21 existing + 4 new)
- [x] `pytest tests/test_eval_case.py tests/test_eval_harness.py tests/test_eval_baseline.py tests/test_eval_lint.py -q` → 69 passed
- [x] Lint: `python -m evals.runner --lint-only` → 247 cases clean
- [x] Pre-registered regression test `test_hallucination_adjacent_clause_numbers_do_not_license_derivation` fails under the interim-buggy implementation and passes after the tightening (design validated end-to-end)
- [ ] Post-merge: re-run weekly cached-mode baseline. No cache invalidation since detector runs on cached responses; re-bootstrap is free.


rozavala:
## Approving — well-designed derivation allowance with robust safety

**Scope:** adds `_percentage_is_derived` to `evals/runner/case.py` as an additional escape hatch inside `_detect_hallucination` for percentages that the model computes from two packet-grounded unit-qualified values. Eval infrastructure only; no production path touches `_detect_hallucination` (grep-confirmed).

**Design review:**

The three layered gates make the check hard to whitewash:

1. **Relational marker** in ±120-char window — loose by design (includes "of", "from", "to"), but isn't load-bearing alone.
2. **Unit-gating** is the real safety — `_VALUE_WITH_UNIT` matches `\d{1,4}(?:,\d{3})*(?:\.\d+)?` immediately followed by `%|percent|pp|million|billion|thousand|mt|bcf|bpa|bushels?|cents?|c/lb|tonnes?|days?|weeks?|bags?|hours?|[MKB](?![a-zA-Z])`. `year` and `month` are intentionally excluded, so `"5-year"` / `"3-month"` cannot be paired as derivation inputs. Verified in-process:
   - `"5-year average"` → no matches
   - `"3.5 million bag"` → `['3.5 million']`
   - `"82 days vs 62 days +32% surplus"` → `['82 days', '62 days', '32%']`
3. **Ratio match ±1pp** across `a/b`, `(a−b)/b`, `(a−b)/a` — covers standard "X% higher/lower than Y" framings.
4. **Corpus presence required for both values** — soft confirmation that the model didn't fabricate the inputs.

The check runs *after* the existing literal-match and ±1pp-neighbor checks, so it strictly relaxes the detector — it cannot produce new false positives, only additional false negatives in narrowly scoped cases.

**Adjacent-clause whitewash guarded:** the load-bearing test `test_hallucination_adjacent_clause_numbers_do_not_license_derivation` encodes the real agronomist/004 trap. `"5-year"` fails unit-gating, so the 5 cannot pair with the 18/52 G/E ratings numbers to coincidentally license "28% below 5-year average". I reproduced the logic manually: `_percentage_is_derived('28%', 28.0, ...)` correctly returns False for this shape.

**Verification I ran:**

1. ✅ **4 new pre-registered tests all pass** (sign-flip flagged, two-number derivation allowed, pure fabrication flagged, adjacent-clause trap rejected).
2. ✅ **Full eval case tests**: 25/25 passed (+4 from baseline).
3. ✅ **Full test suite**: `pytest tests/ -x -q` → 1542 passed in 6m13s.
4. ✅ **Eval gate**: 247/247 mock cases pass, 0.0pp drift on every metric for every agent — hallucination_rate stays 0.0%. Since this is pure relaxation, zero impact in mock mode is expected; value delivers in real-LLM runs.
5. ✅ **Grep-confirmed** `_detect_hallucination` is eval-only, not used in any production trading code path.

**Minor observations (not blocking):**

- The relational-marker list includes common words like `of` / `from` / `to`, so it's near-universal. That's fine because the unit-pair count is the real gate; the marker is just a cheap filter.
- `_number_in_corpus` uses substring match, so `"82"` would match into `"182"`. As the docstring notes, this is acceptable because the unit-gating + ratio-match already restricts false-pass shapes; the corpus check is soft confirmation, not the primary guard.
- The `\d{1,4}(?:,\d{3})*` number capture won't match 5+ digit numbers without comma separators — an intentional conservatism that errs toward false negatives on the allowance, which is the safe direction.

The 4 load-bearing tests explicitly encode the design contract and the commit message names them as pre-registered safety tests. If the detector ever drifts, these will break first — good defense in depth.

Ready to merge.

PR #1761, opened 2026-04-24
fix(emergency-close): preserve original position_id + trend_alignment in ledger (#1760)

rozavala:
Fixes #1760.

## Problem

`emergency_hard_close` was writing each closing leg to the trade ledger under a **new synthetic position_id** of the form \`EMERGENCY_{localSymbol}_{date}_{ts}\` instead of reusing the position_id the spread was opened under. Every emergency close since this code shipped has produced orphan ledger entries — the original spread appears phantom-open (no close matches), and the close entries appear as legs with no open. The \`trend_alignment\` column on close entries was also left blank.

Visible in today's KC emergency close (2026-04-24 17:25 UTC):

- **2026-04-23 open**: \`pos_id=7c3397b2-2360-49ff-be1b-4e6478e0a0c8\`, both legs, \`trend_alignment=WITH_TREND\`.
- **2026-04-24 close**: \`pos_id=EMERGENCY_KON6 P2.9_2026-04-24_1777051505\` and \`pos_id=EMERGENCY_KON6 P2.675_2026-04-24_1777051505\`, \`trend_alignment=(blank)\`.

Downstream consumers that join open/close by position_id (P&L reporting, phantom-stop detection via \`order_manager._find_phantom_stops\`, thesis lifecycle via TMS) all silently miscompute or miss these positions.

## Fix

Two small helpers in \`orchestrator.py\`:

- \`_lookup_original_position_info(open_positions) → {localSymbol: (position_id, trend_alignment)}\` — reads the trade ledger and, for each live position, finds the most recent opening entry whose \`action\` direction matches the position's sign (long → BUY, short → SELL). Returns \`{}\` on ledger load failure, empty ledger, or missing columns so callers can fall back cleanly.
- \`_resolve_emergency_position_id(sym, info, date)\` — returns \`(position_id, trend_alignment)\` when the lookup hit, otherwise falls back to the legacy \`EMERGENCY_{sym}_{date}_{ts}\` with a **WARNING** log so the continuity loss is visible rather than silent.

Wired into all three \`log_trade_to_ledger\` call sites:

- \`orchestrator.py:6130\` — primary emergency close
- \`orchestrator.py:6224\` — margin-retry (Error 201)
- \`orchestrator.py:6256\` — reduced-qty retry

Each now passes \`trend_alignment=\` through to the ledger write too.

## Compatibility

- Ledger schema unchanged. Already-written \`EMERGENCY_*\` rows in archive ledgers stay as-is (forward-only fix).
[...]

Issue #1760, opened 2026-04-24
bug: emergency_hard_close logs each leg under a new per-leg EMERGENCY_ position_id, orphaning the original spread in the ledger


rozavala:
## Problem

`emergency_hard_close` in `orchestrator.py` builds a fresh position_id for **each closing leg** instead of reusing the **original position_id** the spread was opened under. Result: the original spread's position_id has no close entry in the ledger (appears phantom-open), and the closing legs appear as orphan closes with no matching open.

Source sites:

- `orchestrator.py:6122` — `position_id=f"EMERGENCY_{contract.localSymbol}_{today_date}_{int(_time.time())}"`
- `orchestrator.py:6216` — same pattern with `sym`
- `orchestrator.py:6248` — same pattern with `sym`

The pattern has existed since at least March 2026 (found `EMERGENCY_*` entries in `/data/KC/archive_ledger/trade_ledger_20260305_184116.csv`, `/data/NG/archive_ledger/trade_ledger_20260305_*.csv`, etc.), but is newly visible today because today's emergency_hard_close actually fired for KC.

## Evidence — 2026-04-24 KC

**Yesterday's open** (one position_id for both legs — correct):

```
2026-04-23 15:08, pos_id=7c3397b2-2360-49ff-be1b-4e6478e0a0c8, KON6 P2.675, SELL 1, WITH_TREND
2026-04-23 15:08, pos_id=7c3397b2-2360-49ff-be1b-4e6478e0a0c8, KON6 P2.9,   BUY  1, WITH_TREND
```

**Today's emergency close** (per-leg position_id, trend_alignment lost):

```
2026-04-24 17:25, pos_id=EMERGENCY_KON6 P2.9_2026-04-24_1777051505,   SELL 1, (trend_alignment blank)
2026-04-24 17:25, pos_id=EMERGENCY_KON6 P2.675_2026-04-24_1777051505, BUY  1, (trend_alignment blank)
```

Downstream effect: any reporter that joins open/close by position_id will see:

- `7c3397b2-...-0a0c8` → still "open" (no close entry matches)
- `EMERGENCY_KON6 P2.9_...` → orphan close (no open entry matches)
- `EMERGENCY_KON6 P2.675_...` → orphan close (no open entry matches)

IBKR actually filled both closes at 17:25:05 (per `Orchestrator - INFO - Emergency fill: ...`), so the **real** position is flat. The misalignment is purely in local ledger continuity.

## Why this matters

[...]

Issue #1758, opened 2026-04-24
feature: auto-restart IB Gateway on sustained handshake TimeoutError (wedge watchdog)


rozavala:
## Problem

IB Gateway has now wedged **twice in 3 days** with an identical signature, each requiring manual restart to recover:

| Date | Issue | Outage duration | Recovery trigger |
|---|---|---|---|
| 2026-04-22 | #1708 (closed) | ~45 min across 5 notifications | manual Gateway restart |
| 2026-04-24 | #1751 (today) | ~108 min, 8 FORCE_RESETs, 58 backoff escalations | manual Gateway restart at 14:03:55 UTC |

Both incidents have the same fingerprint in logs:

- `ib_insync.client - ERROR - API connection failed: TimeoutError()` (handshake times out at 15s)
- TCP port stays open — no `ConnectionRefusedError`
- Every reconnect attempt fails the same way, so exponential backoff climbs to 600s but never recovers
- FORCE_RESET triggers when `consecutive_failures >= FORCE_RESET_THRESHOLD = 5` but only re-instantiates the client-side `IB` object — it **can't unwedge the Gateway process**
- The moment the user kills and restarts Gateway, the error shape flips to `ConnectionRefusedError(111, ...)` (port now actively refused), then clean reconnects resume within seconds

Current mitigations (`connection_pool.py`) all operate on the client side: `_force_reset_connection`, exponential backoff, clean-disconnect hold-off, portfolio outage detector. None of them **actually restart the Gateway process**, so a wedged Gateway stays wedged until manual intervention. The Pushover notification goes out within ~10 min, but nobody is guaranteed to be awake or near a laptop.

## Proposal

Add a **Gateway wedge watchdog** that:

1. **Detects the wedge signature** distinctively:
   - Sustained `TimeoutError()` on handshake, across **multiple pool keys**, for **N consecutive minutes** (suggest 15 min — long enough that a routine IBKR-side blip has recovered, short enough to beat the 108-min worst case).
   - Absence of `ConnectionRefusedError` — that would indicate Gateway is already down for a legitimate reason (restart in progress, systemd stopped it), so piling more restarts on top makes things worse.
   - Portfolio outage detector has already fired (piggyback on `_check_portfolio_outage` as the gating signal rather than re-deriving the heuristic).

2. **Restarts the Gateway process**:
   - DEV (remote droplet): SSH + `systemctl --user restart ibgateway` or equivalent.
[...]

Issue #1751, opened 2026-04-24
[CRITICAL] uncategorized: PORTFOLIO OUTAGE DETECTED: 5 IB connection failures across all purposes in th...


rozavala:
## Error Report

| Field | Value |
|-------|-------|
| **Category** | `uncategorized` |
| **Severity** | CRITICAL |
| **Environment** | DEV |
| **First seen** | 2026-04-24 12:54:02,285 |
| **Last seen** | 2026-04-24 12:54:02,285 |
| **Occurrences** | 2 |

### Sample Log Entry

```
PORTFOLIO OUTAGE DETECTED: 5 IB connection failures across all purposes in the last 5 min. IB Gateway may be fully unreachable.
```

### Impact

Investigate to determine impact on trading operations.

> The `claude-fix` label will be added after triage — Claude will automatically attempt a fix.

---
*Automated report by `scripts/error_reporter.py` — fingerprint: `188caa20515a`*

rozavala:
## Root cause analysis (from DEV logs on Droplet, 2026-04-24)

**Pattern: IB Gateway internal deadlock.** TCP port stayed open (accepted connections) but the API handshake never completed — so every reconnect attempt hit `ib_insync.client - ERROR - API connection failed: TimeoutError()` after the 15s handshake deadline. Not a network issue, not a port closure — the Gateway process was alive but wedged.

### Timeline (UTC)

| Time | Event |
|---|---|
| 04:28:22 | Earlier warning: Error 1100 (connectivity lost) on NG_reader / ZW_reader; recovered by 04:30:10 (Error 1102). Gateway was already flaky. |
| 12:16:14 | First symptom: `SB_reader` disconnect (routine post-task release triggered by `Risk Agent timed out after 30s` — LLM-side, unrelated to Gateway). |
| 12:17:40 | First `SB_reader` reconnect attempt — **TimeoutError on handshake**. Gateway has become unresponsive. |
| 12:17–12:27 | SB_reader backoff climbs: 10s → 20s → 40s → 80s → 160s → 320s. |
| 12:23:59 | First `FORCE RESETTING connection: SB_reader` (FORCE_RESET_THRESHOLD = 5 consecutive failures). |
| 12:26:00 | First CRITICAL: `IB outage notification sent after 10 minutes`. |
| 12:46:57 | NG_reader also starts failing — cascade spreads as other engines' reconnect timers fire. |
[...]

PR #1762, opened 2026-04-24
fix(startup): release without holdoff in cleanup_orphaned_theses + process_deferred_triggers (#1759)

### LINKED ISSUE #1759: bug: startup bootstrap disconnects use holdoff=True, blocking post-restart recovery within 60s
## Problem

When the bot restarts (CI auto-deploy, manual restart, crash recovery), the engine startup sequence briefly **acquires and releases** order-bearing connections (`{ticker}_orders`, `{ticker}_reader`) as part of its bootstrap — e.g. for audit/sanity checks. Those releases go through `release_connection` with its default `holdoff=True`, which stamps `_last_clean_disconnect[pool_key] = now` and triggers a 60-second `RECONNECT_HOLDOFF_AFTER_CLEAN_DISCONNECT` window (#1711 Q8).

If the **RECOVERY harness** then tries to re-run a missed task (like `close_stale_positions`) within that 60s window, it's blocked — the per-caller 3-attempt retry loop (5s + 10s = 15s total) gives up long before the hold-off expires.

## Evidence — 2026-04-24 18:08 restart

CI auto-deploy after PR #1757 merge restarted the bot at 18:08:41 UTC. Sequence for NG:

```
18:09:22 - Connecting IB (NG_orders) ID: 1237 [REMOTE/PAPER]...
18:09:22 - IB (NG_orders) connected successfully with ID 1237
18:09:22 - DISCONNECT GUARD [NG_orders]: No pending orders. Safe to disconnect.
18:09:22 - Disconnecting IB (NG_orders)...           ← holdoff=True stamped here
18:09:47 - 🔄 RECOVERY: Running missed close_stale_primary (was scheduled 14:05 ET)
18:09:47 - close_stale_positions connection attempt 1/3 failed: Connection NG_orders in post-disconnect hold-off. Retrying in 5.0s...
18:09:52 - close_stale_positions connection attempt 2/3 failed: Connection NG_orders in backoff period. Retrying in 10.0s...
18:10:02 - CRITICAL - close_stale_positions FAILED after 3 attempts. Positions may remain open!
```

Same shape for ZW_orders: bootstrap disconnect at 18:09:27 → recovery retry at 18:09:52 → failure after 3 attempts at 18:10:07.

KC and SB weren't affected because their close tasks had already completed in the prior process window (KC 17:05-17:25, SB 16:35-16:45), so the task_tracker (or the "market closed" skip path) correctly short-circuited the recovery.

## Why this matters

- `close_stale_positions` is a **safety-net** job — it's meant to catch positions the scheduled signal didn't close. Having it silently fail post-restart degrades that safety net.
- The larger the rate of CI auto-deploys on DEV, the more exposure we have to this race. Today alone: 3 restarts (15:56, 16:54, 18:08) from unrelated PR merges.
- Not a regression: `holdoff=True` default comes from #1711 Q8 and is correct for its original use case (cycle-transition tearing down + immediately reacquiring from the same caller). The bug is that startup bootstrap code uses the same default when it shouldn't.

## Proposed fixes (pick one, or combine)

**Option A — startup uses holdoff=False (narrow, preferred).** Audit every place the engine bootstrap path calls `release_connection` and pass `holdoff=False` explicitly. This preserves #1711 Q8's semantics for the cycle-transition case it was built for, and unblocks recovery.

**Option B — recovery retry window ≥ 60s.** Make `close_stale_positions` (and other recovery-harness entry points) retry for at least `RECONNECT_HOLDOFF_AFTER_CLEAN_DISCONNECT` duration. More defensive but wastes time on the happy path.

**Option C — recovery skip if pool key is in hold-off from <30s ago.** Treat recent clean disconnect as "already done" — if the pool was just released cleanly, assume the in-flight task handled its own state and don't re-run. Riskier semantic change.

Option A is the most surgical and the smallest blast radius. Option B would also fix it but trades clarity for resilience.

## Acceptance criteria

- [ ] After bot restart (simulated or real CI deploy), `close_stale_positions` recovery run acquires pool connections without hitting `post-disconnect hold-off` OR retries long enough to clear it.
- [ ] No `CRITICAL: close_stale_positions FAILED after 3 attempts` messages appearing within 60s of process start.
- [ ] Integration test: simulate a `release_connection(holdoff=True)` followed by a recovery-triggered `get_connection` within 30s; assert success (with Option A or C) or delayed success (with Option B).

## Related

- #1751 — today's IB outage (the context in which this bug surfaced)
- #1711 — Q8 introduced `RECONNECT_HOLDOFF_AFTER_CLEAN_DISCONNECT`
- #1758 — Gateway wedge watchdog (orthogonal; addresses Gateway-side issue, not client-side)

rozavala:
Fixes #1759.

## Problem

When the orchestrator restarts (CI auto-deploy, crash recovery), the recovery harness runs missed tasks in sequence. Two of the earliest runners — \`cleanup_orphaned_theses\` and \`process_deferred_triggers\` — released their IB connections with the default \`holdoff=True\`, stamping the 60-second post-disconnect hold-off (from #1711 Q8) on the \`{ticker}_reader\` and \`{ticker}_orders\` pool keys.

Later recovery tasks (\`close_stale_primary\`, \`audit_*\`, etc.) that needed the same pool keys within that 60s window hit \`Connection X in post-disconnect hold-off\` and gave up after their 3-attempt retry (5s + 10s = 15s, much less than 60s).

## Evidence — 2026-04-24 18:08 restart

CI auto-deploy after an unrelated PR merge restarted the bot at 18:08:41 UTC.

\`\`\`
18:09:22 - Connecting IB (NG_orders) ID: 1237 [REMOTE/PAPER]...
18:09:22 - DISCONNECT GUARD [NG_orders]: No pending orders. Safe to disconnect.
18:09:22 - Disconnecting IB (NG_orders)...          ← holdoff=True stamped here
18:09:47 - 🔄 RECOVERY: Running missed close_stale_primary (was scheduled 14:05 ET)
18:09:47 - close_stale_positions connection attempt 1/3 failed: Connection NG_orders in post-disconnect hold-off. Retrying in 5.0s...
18:09:52 - close_stale_positions connection attempt 2/3 failed: Connection NG_orders in backoff period. Retrying in 10.0s...
18:10:02 - CRITICAL - close_stale_positions FAILED after 3 attempts. Positions may remain open!
\`\`\`

Same shape for ZW.

## Fix

Two one-kwarg changes:

- \`orchestrator.py:1386\` (\`cleanup_orphaned_theses\`) → \`release_connection(\"cleanup\", holdoff=False)\`
- \`orchestrator.py:4153\` (\`process_deferred_triggers\`) → \`release_connection(\"deferred\", holdoff=False)\`

Both release paths already live in \`finally\` blocks — only the \`holdoff\` kwarg changes. Comments cite #1759 and the rationale (next recovery task needs the same pool key immediately).

Matches the precedent at \`order_manager.py:1936\` (\`generate_and_queue_orders\`), which already opts out of hold-off for the same \"sequential reacquire within one caller chain\" reason. Both of these also fit that pattern — routine scheduled runs are hours apart, so the hold-off was never load-bearing for the happy path; it only mattered as a back-pressure guard for shutdown and cycle transitions, neither of which applies to these two functions.

## Why narrow over harness-wide

Considered a harness-level fix — clear \`_last_clean_disconnect\` at the start of each recovery task run. Rejected because:

- Bypasses a safety mechanism globally during recovery instead of letting each caller declare what it needs.
- The harness doesn't know which tasks will actually acquire pool keys or which pool keys they'll use.
- Two sites is audit-complete for today's observed failure; other paths already use \`holdoff=False\` (order_manager.py:1936) or don't run during recovery.

If a third site surfaces, the fix pattern is equally small.

## Test plan

- [x] \`pytest tests/test_startup_bootstrap_holdoff.py\` — **5/5 pass** in 17.8s. Covers:
  - Happy path: \`cleanup_orphaned_theses\` calls release with \`holdoff=False\`
  - Exception path: \`cleanup_orphaned_theses\` still releases with \`holdoff=False\` when body raises
  - Happy path: \`process_deferred_triggers\` with an active trigger releases with \`holdoff=False\`
  - Exception path: \`process_deferred_triggers\` whose downstream raises still releases with \`holdoff=False\`
  - No-op path: \`process_deferred_triggers\` with empty queue doesn't acquire a connection (so no release either)
- [x] \`pytest tests/test_deferred_triggers.py tests/test_connection_pool.py\` — **93/93 pass** no regressions
- [ ] Post-merge on DEV: simulate restart, verify no \`post-disconnect hold-off\` warnings from recovery harness within the first 90s of startup

## Not fixed here

- Hold-off default value — leaving at \`True\`. The Q8 semantics for cycle-transition churn are correct; only these two startup-path sites were wrong.
- Gateway wedge watchdog (#1758) — orthogonal, addresses Gateway-side wedging, not client-side hold-off.

## Related

- #1751 — today's IB outage that surfaced this
- #1711 — Q8 introduced \`RECONNECT_HOLDOFF_AFTER_CLEAN_DISCONNECT\`
- #1760 — separate ledger-continuity bug from the same incident; PR #1761


rozavala:
## Approving — narrow, well-justified fix that mirrors an existing precedent

**Scope:** two single-kwarg changes (`holdoff=False`) on `release_connection` calls in `cleanup_orphaned_theses` and `process_deferred_triggers`, plus a new test file with 5 tests.

**Verification I ran:**

1. ✅ **API exists** — `release_connection(cls, purpose: str, holdoff: bool = True)` at `connection_pool.py:833`. The signature explicitly documents the `holdoff=False` opt-out: "Pass `holdoff=False` when the caller intends to immediately re-acquire" (`connection_pool.py:848`).
2. ✅ **Holdoff window is real** — `RECONNECT_HOLDOFF_AFTER_CLEAN_DISCONNECT = 60.0` at `connection_pool.py:160`, with the documented purpose "suppress reconnects on the same pool key for this window so a caller that tears down and immediately reacquires does not pound Gateway with open/close churn during shutdown or cycle transitions" (`connection_pool.py:155-159`).
3. ✅ **Precedent is real** — `order_manager.py:1936` uses `holdoff=False` already, with the same "sequential reacquire within one caller chain" rationale. The two new sites fit the same pattern: routine scheduled runs are hours apart, so the hold-off was never load-bearing for the happy path; it only mattered as back-pressure for shutdown/cycle transitions.
4. ✅ **Bug confirmed** — checked the current orchestrator on `feat/1643-...` working tree: `release_connection("cleanup")` and `release_connection("deferred")` both still use the default `holdoff=True`. Default is `True` per signature, so the 60s window is stamped on `{ticker}_reader` and `{ticker}_orders` after each runs, blocking subsequent recovery tasks within the next 60s.
5. ✅ **Recovery-task retry math holds** — close_stale_positions' 3-attempt retry (5s + 10s = 15s) gives up well before the 60s hold-off expires, exactly as the PR description claims.
6. ✅ **Targeted tests**: 5/5 pass — happy + exception paths for both functions, plus the no-op (empty queue) sanity check.
7. ✅ **Full test suite**: 1547 passed in 8m13s (+5 from baseline). No regressions.

**Design assessment:**

The narrow fix is the right call over the harness-wide alternative the PR's §"Why narrow over harness-wide" considered. Bypassing the safety mechanism globally during recovery would invert the responsibility model — the harness shouldn't have to know which downstream tasks will reacquire which pool keys. Two-site audit is complete for today's observed failure shape; if a third surfaces, the same one-line fix applies.

**Both finally-blocks tested for the exception path.** This matters because the original failure was triggered by an unrelated PR merge (CI auto-deploy → restart). The exception-path tests confirm that the fix holds even when the cleanup body raises mid-run — the finally still releases with `holdoff=False`. Critical for outage-resilience.

**Ledger continuity untouched** — this PR doesn't interact with #1760/#1761's emergency-close fix; they're parallel symptoms of the same incident but address different code paths.

Ready to merge.

Issue #1643, opened 2026-04-17
feature: intentional scale-in with parent/child thesis lineage


rozavala:
## Context

PR #1642 blocks silent cross-cycle duplication at the thesis-coherence gate — the correct *default*. But it's a deliberate floor, not a ceiling: there are legitimate reasons to add to a position across cycles (growing conviction, thesis playing out favorably, laddered entries). Today there's no first-class way to express that intent; adds are either blocked by #1642 or leak through as drift (#1613 SOV6 accumulation).

This issue scopes a proper scale-in mechanism so operators can opt in without re-opening the dedup hole.

## Design principle: "add to winners only"

Time-alone triggers ("N cycles in a row") are a Martingale trap — they fire just as readily when the market is flat and the council is repeating itself as when the thesis is actually playing out. The council being *repeatedly right* and the council being *repeatedly stuck* look identical at the signal layer.

Any scale-in must gate on the **market confirming the thesis**, not the council repeating itself:
- underlying has moved in the thesis direction
- existing position is unrealized-profitable
- new signal confidence has materially increased over the original

## Proposed config (opt-in, default off)

\`\`\`yaml
strategy_tuning.scale_in:
  enabled: false                    # opt-in per engine
  max_adds_per_thesis: 2            # hard cap on children
  min_cooldown_hours: 24            # no same-day stacking
  min_underlying_move_pct: 0.5      # price confirmed direction
  min_unrealized_pnl_pct: 2.0       # position is already winning
  min_confidence_delta: 0.10        # new signal stronger than original
\`\`\`

All predicates must be true. Any one false → \`BLOCK_ACTIVE_DUPE\` as today (#1642).

## Gate placement

Inside \`_check_thesis_coherence\` (\`trading_bot/order_manager.py\`), at the point where \`BLOCK_ACTIVE_DUPE\` is returned. Before returning, check scale-in eligibility; if eligible, attach \`scale_in_of_thesis_id\` + \`parent_confidence\` to the signal and return \`PROCEED_SCALE_IN\` (or just \`PROCEED\` with the extra fields).

## TMS lineage (schema change)

Add \`parent_thesis_id\` field to entry thesis records.
- Children inherit exit gates from parent (if parent is marked \`pending_close\`, children cascade-close).
- \`reconcile_orphaned_theses\` must walk the lineage when deciding orphan status.
- \`get_active_theses_by_contract\` returns parents + all non-closed children; dedup gate sees them all.

[...]

Issue #1613, opened 2026-04-16
bug: SB trade quantity mismatch (1 lot in, 4 lots out) causing PnL distortion


rozavala:
### Description
The Financials page shows an incorrectly large positive PnL for SB positions. Ledger analysis reveals that the system is entering 1 lot but attempting to exit 4 lots for the same position_id.

### Example (Thesis cbac2939)
- **Entry (2026-04-13):** 1.0 lot SOV6 Bear Put Spread @ 0.47 debit.
- **Exit (2026-04-15):** 4.0 lots SOV6 Bear Put Spread @ 0.43 credit.
- **Net Result:** Dashboard shows +400 PnL because it sums the total_value_usd across all 5 ledger entries (1 in, 4 out).

### Impact
Massive distortion of PnL metrics and potential over-exposure/execution errors if the bot is actually attempting to sell more than it owns.

rozavala:
Fixed in PR #1612. Root cause confirmed via archive ledger: 4 separate 1-lot entries for SOV6 P0.1225/P0.1375 (2/day on April 13-14) accumulated due to missing duplicate-spread gate. Fix: added same-direction accumulation check to `_check_contract_overlap` — if ALL legs of a proposed spread match an active thesis in the same direction, block with 'BLOCKED - Duplicate Spread'. The 4-lot close on April 15 at 12:21 was from `run_position_audit_cycle` triggered by a PriceSentinel ≥1.5% move, not a reconnect or ledger bug.

PR #1764, opened 2026-04-24
feat(tms): Phase 1 lineage plumbing — parent_thesis_id + entry_price (#1643)

rozavala:
Phase 1 of #1643 (scale-in with parent/child thesis lineage). Behavior-neutral schema + migration — Phase 2/3 will add the lineage-aware exit cascade and the scale-in predicate gate respectively.

Follows the 5-phase plan from [#1643's prior-session design comment](https://github.com/rozavala/real_options/issues/1643#issuecomment-4265421198) and the three product decisions confirmed this session:

1. **Brier attribution** → Option B (adds are extensions of parent, not independent predictions)
2. **Predicates** → per-commodity overrides + ordinal confidence delta (not float)
3. **Child exits** → Hybrid (thesis-level cascades, price-level per-entry)

None of those decisions gate Phase 1, but they'll drive Phase 3's predicate implementation.

### \`trading_bot/tms.py\`
\`record_trade_thesis\` now writes two new ChromaDB metadata fields:

- \`parent_thesis_id\`: \`\"\"\` for top-level (ChromaDB forbids \`None\`), \`\"thesis_{parent_id}\"\` for children.
- \`entry_price\`: prefers \`thesis_data['entry_price']\` (top-level canonical), falls back to \`supporting_data.entry_price\` (the effective breach-calc price used today for IC / straddle strategies), 0.0 sentinel if neither present. Non-numeric input coerces to 0.0 without crashing.

### \`trading_bot/order_manager.py\`
\`record_entry_thesis_for_trade\` gains an optional \`parent_thesis_id: str = None\` kwarg that threads into \`thesis_data\` before the TMS write. All existing call sites work unchanged (default stays empty). Phase 3's scale-in gate will be the first producer to pass it.

### \`scripts/migrations/backfill_tms_lineage.py\`
Idempotent backfill for legacy records. For each engine's \`data/{TICKER}/tms\` dir:

- Reads existing \`entry_thesis\` records from the \`agent_insights\` collection.
- Fills missing \`parent_thesis_id\` with \`\"\"\`, missing \`entry_price\` with the doc-body's \`supporting_data.entry_price\` (or 0.0).
- Mixed-state records (one field present, one missing) only get the missing field filled.
- Second run reports 0 updated; safe to run repeatedly.
- \`--dry-run\` flag reports what would change without writing.
- Default scans all \`data/*/tms\` dirs; pass explicit dirs to override.

### \`tests/test_tms_lineage_phase1.py\` (new)

| Cluster | Tests |
|---|---|
| \`record_trade_thesis\` contract | 5 — top-level default, child parent id, supporting_data fallback, missing→0.0, non-numeric coercion |
| Migration idempotency | 6 — fresh backfill, two-run idempotency (1→0), dry-run no-op, empty-doc fallback, mixed-state partial fill, missing-dir safe skip |
| Wiring through order_manager | 2 — parent_thesis_id threading, default-empty behavior |

## Test plan

- [x] \`pytest tests/test_tms_lineage_phase1.py\` — **13/13 pass** in 16.8s.
- [x] \`pytest tests/test_order_manager.py\` — **60/60 pass**, no regressions.
- [ ] Post-merge: run migration on DEV, verify per-engine counts match expectation (most records backfill with 0.0 for entry_price since they predate the canonical top-level field).
- [ ] Phase 2 PR picks up lineage-aware exits from here.

## Rollout

1. Merge this (schema + migration + writer plumbing; no read sites yet).
2. On DEV, run \`python scripts/migrations/backfill_tms_lineage.py data/KC data/NG data/SB data/ZW\`. Single pass; idempotent if rerun.
3. Let one DEV cycle run to confirm new entries carry the fields (\`grep 'parent_thesis_id' ...\` on the TMS log line is a quick check).
4. Phase 2 PR starts once this has baked one DEV cycle cleanly.

## Related

- #1643 — scale-in feature (this is Phase 1 of 5)
- #1642 — cross-cycle dedup floor this feature sits on top of
- #1613 — the SOV6 accumulation incident that motivated the hardening


rozavala:
## Approving — clean Phase 1 plumbing with strong forward-looking discipline

**Scope:** behavior-neutral schema additions + idempotent backfill migration. New ChromaDB metadata fields (`parent_thesis_id`, `entry_price`) on entry-thesis records; nothing reads them yet (Phase 3's scale-in gate will be the first reader).

**Verification I ran:**

1. ✅ **`effective_entry_price` mirrors live breach-calc semantics** — `order_manager.py:4980` computes `effective_entry_price = underlying_price if underlying_price else entry_price`, then writes it to both `supporting_data.entry_price` (existing) and the new top-level `entry_price` field. So the new metadata reads exactly the value the breach-calc code already uses — Iron Condor uses underlying price; directional spreads use spread credit/debit. No semantic split between what the predicate will read and what the closer-side code already operates on.
2. ✅ **TMS public API exists** — `set_data_dir` at `tms.py:50` and `TransactiveMemory.__init__(persist_path: str = None)` at `tms.py:121`. Tests use both correctly.
3. ✅ **ChromaDB None constraint sentinel** — empty string for `parent_thesis_id`, 0.0 for `entry_price`. The `_safe_float` and `or ''` patterns coerce defensively; non-numeric input (e.g. `'not-a-number'`) coerces to 0.0 without raising.
4. ✅ **Migration idempotency** — `if not (needs_parent or needs_entry): already_ok += 1; continue` — a record with both fields present is left untouched. Two-run test asserts `(updated, already) = (1, 0)` then `(0, 1)`.
5. ✅ **Migration mixed-state handling** — `needs_parent` and `needs_entry` are independent; only the missing field gets filled. Test `test_mixed_state_record_only_backfills_missing_field` confirms preserved-field behavior.
6. ✅ **Migration entry_price resolution** — `_resolve_entry_price_from_doc` mirrors the live write-path preference order (top-level → `supporting_data.entry_price` → 0.0). Correctly returns 0.0 for malformed JSON, non-dict docs, and missing fields. Edge case "top-level is genuinely 0.0" falls through to `supporting_data` — the right behavior since 0.0 is the unknown sentinel; if `supporting_data` has a real value, prefer that.
7. ✅ **`record_entry_thesis_for_trade` keyword-defaulted** — `parent_thesis_id: str = None` with all existing call sites unchanged. Backward-compatible plumbing as claimed.
8. ✅ **13 targeted tests pass**: 5 contract tests on `record_trade_thesis`, 6 migration-script tests (idempotency, dry-run, fallback, mixed-state, missing dir), 2 wiring tests on `record_entry_thesis_for_trade`.
9. ✅ **Full test suite**: 1555 passed in 8m19s (+13 from baseline). No regressions.

**Migration-script design notes (good):**

- `--dry-run` flag works correctly (test `test_dry_run_does_not_write` confirms metadata still lacks new fields after dry-run).
- Default scans `data/*/tms` automatically (no need to enumerate engines), but accepts explicit dirs to override.
- Batch-update size 1000 is safely within ChromaDB limits.
- Lazy `import chromadb` inside `backfill_tms_dir` so `--help` works without the heavy import.
- Missing directories print SKIP rather than crashing — important for the multi-commodity loop where some engines may not have a `tms/` dir yet.

**Safety assessment for the production-touched files:**

- `trading_bot/order_manager.py` — single new optional kwarg threaded through to `thesis_data` before TMS write. No call-site changes; default keeps existing behavior. The `effective_entry_price` semantic is preserved (line 4980 unchanged), and the new top-level `entry_price` in `thesis_data` mirrors that same value (lines 5005-5006).
- `trading_bot/tms.py` — `record_trade_thesis` now writes two extra metadata keys. The `try/except (TypeError, ValueError)` around the float coercion prevents legacy callers passing garbage from crashing the write. The existing `symbol`/`contract_month` reads were refactored to use a hoisted `supporting = thesis_data.get('supporting_data', {}) or {}` local — same semantics, less repetition.

**Phase boundary discipline:**

This PR is genuinely behavior-neutral — no read site uses the new fields yet, so even if the backfill produced surprising values for legacy records, nothing in production would act on them. Phase 2/3 land on a fully populated metadata floor, which is exactly the right shape for an additive migration.

**Rollout plan checks out:** merge → run migration on DEV → let one DEV cycle bake → start Phase 2. Single-pass migration with second-run-zero-updates idempotency means re-running on PROD if needed is safe.

Ready to merge.

PR #1767, opened 2026-04-24
evals: refresh cached-mode baselines after Steps 1–5A cleanup (activates cached gate)

rozavala:
## Summary

Regenerate `evals/baselines/cached/*.json` from the current **247-case** library. **This activates the cached-mode gate and the weekly drift cron** — both have been running against stale baselines from a 154-case library captured on 2026-04-23 07:45Z (before canonical_neutral, adversarial.006_*, contrast _007, and the Step 1–5A cleanup PRs).

8 files changed, +75 / −75 lines. No code changes.

## Why now

The cached-mode per-PR gate (`eval-gate.yml` `full` job on production-targeting PRs) and the weekly drift cron (`eval-drift-weekly.yml`, Mondays 14:00 UTC) both compare against `evals/baselines/cached/*.json`. Running them against the 154-case baselines on a 247-case library produces either size-mismatch errors or false-positive alarms that don't reflect real regressions.

All the Step 1–5A cleanup work was preparation for this moment. The case library is in a state where baselining it is defensible:
- Hallucination rates are down materially (Step 2 + 3 + 5A)
- Case-construction defects (substring FPs, packet echoes, derivation detector gaps) are resolved
- The remaining 40 failures are documented-known — not bugs to fix first

## Before → after per-agent

| agent | total | pass rate | halluc | contrast_consistency |
|---|---|---|---|---|
| agronomist | 14→23 | 43%→57% | 0.38→0.29 | 1.00→0.67 * |
| geopolitical | 20→32 | 55%→81% | 0.00→0.04 | 0.75→1.00 |
| inventory | 20→32 | 85%→97% | 0.25→0.04 | 1.00→1.00 |
| macro | 20→32 | 50%→88% | 0.00→0.00 | 1.00→0.88 |
| sentiment | 20→32 | 65%→75% | 0.00→0.00 | 1.00→0.88 |
| supply_chain | 20→32 | 75%→88% | 0.00→0.00 | 1.00→1.00 |
| technical | 20→32 | 90%→94% | 0.00→0.00 | 0.75→0.88 |
| volatility | 20→32 | 50%→84% | 0.08→0.00 | 1.00→0.88 |

* Agronomist contrast drop is the new `phenology_007` pair (added in #1750) that the cheap eval model fails on. **Genuine behavior capture, not regression** — the drop from 1.00 → 0.67 just means the expanded library includes harder contrast cases that the model can't flip. This is what the baseline should record.

## What's baked into the baseline

40 documented failures, categorized:
[...]

PR #1768, opened 2026-04-24
evals: scope cache signature to the current role's override

rozavala:
## Summary

Change `_cache_signature` in `evals/runner/harness.py` to fold only *this role's* entry from `eval_models.json` into the signature, not the whole overrides dict. Prerequisite for cheap per-agent model-swap experiments.

## The scoping bug

Before:
```python
sig = json.dumps({
    "p": prompt, "m": registry, "t": temperature,
    "o": self._eval_overrides_cached,  # full dict — ANY change busts ALL signatures
}, ...)
```

Swapping agronomist's model in `eval_models.json` (e.g. `gemini-flash` → `claude-haiku`) changes the full dict → invalidates the signature for every case in the library, not just the ~23 agronomist cases that actually route through the new model. A targeted model-swap experiment on one agent costs a full 247-case re-bootstrap (~$1.24).

## The fix

```python
def _cache_signature(self, prompt, role=None):
    ...
    override_payload = (
        self._eval_overrides_cached.get(role.value) if role is not None
        else self._eval_overrides_cached
    )
    sig = json.dumps({"p": prompt, "m": registry, "t": temperature, "o": override_payload}, ...)
```

- `_cached_call` computes `role = _role_for_agent(case.agent)` and passes it through.
- Swapping agronomist's override now only changes agronomist's signatures. The 200+ other-agent cases keep their cached responses.
- Per-agent model-swap experiments become **~\$0.12 instead of ~\$1.24**.
- Defensive fallback: when `role is None` (any call site that doesn't know the role), the full dict is still hashed — **over-invalidation preferred over silent cache collisions**.

## Four load-bearing tests

Any regression breaks one of these. Don't relax them without considered re-design.

| Test | What it guards |
|---|---|
| `test_cache_signature_includes_eval_overrides` | Swapping *this role's* override produces a new key (the original regression guard, verified against scoped form) |
| `test_cache_signature_stable_when_overrides_unchanged` | Determinism under identical inputs |
| **`test_cache_signature_scoped_to_role_unrelated_override_does_not_bust`** | **THE NEW GUARD.** Changing macro's override must NOT invalidate agronomist's signature. Also sanity-checks: changing macro's override DOES invalidate macro's signature. |
| `test_cache_signature_fallback_when_role_is_none_folds_in_full_dict` | `role=None` path preserves safe-over-sorry behavior |

## Rollout cost

The signature formula itself changed, so **the first bootstrap run post-merge invalidates all 247 cached responses**. One-time cost: ~\$1.24. After that, per-agent experiments are cheap.

Post-merge, a baseline refresh (like #1767) may be warranted if the fresh responses shift aggregate metrics enough to trip drift thresholds. Suggest triaging after the first weekly cron run. Worst case we capture a small second baseline in the new regime; best case nothing shifts and existing baselines hold.

## Downstream value

This is the prerequisite for:
- **6B: agronomist model swap** (swap Gemini Flash → Claude Haiku for agronomist only) — now ~\$0.12 per experiment, not ~\$1.24.
- **Macro cache-churn diagnosis** (flagged in #1767 follow-up) — if churn survives scoped signatures, the instability is not in `eval_overrides` and we can narrow the investigation.

## Test plan

- [x] `pytest tests/test_eval_harness.py -v` → 28 passed, including 4 signature-behavior tests
- [x] `pytest tests/test_eval_case.py tests/test_eval_harness.py tests/test_eval_baseline.py tests/test_eval_lint.py tests/test_specialist_prompt.py tests/test_eval_models.py -q` → 102 passed
- [x] `python -m evals.runner --lint-only` → 247 cases clean
- [ ] Post-merge: monitor next `eval-drift-weekly` run — if drift alarms fire on fresh responses, land a follow-up baseline refresh.


rozavala:
## Approving — surgical scoping fix that unblocks cheap per-agent experiments

**Scope:** changes `_cache_signature(prompt)` → `_cache_signature(prompt, role=None)` in `evals/runner/harness.py`. When `role` is provided, only that role's entry from `_eval_overrides_cached` is folded into the signature; full-dict fallback when `role is None`.

**Verification I ran:**

1. ✅ **`_role_for_agent` exists at `harness.py:393`** with the full agent → AgentRole mapping (8 specialists). Unknown agents raise `ValueError` — exactly what `_cached_call`'s `try/except ValueError → role=None` catches.

2. ✅ **`AgentRole.value` matches `eval_models.json` keys** — `AgentRole.AGRONOMIST.value = "agronomist"`, `AgentRole.MACRO_ANALYST.value = "macro"` (per `heterogeneous_router.py:44-55`). And `apply_eval_overrides` in `evals/runner/eval_models.py:54` keys overrides by these same `role_value` strings. So `self._eval_overrides_cached.get(role.value)` correctly extracts THIS role's override entry.

3. ✅ **The 4 load-bearing tests cover the contract:**
   - `test_cache_signature_includes_eval_overrides` — swap THIS role's override → signature changes (existing regression guard, now scoped)
   - `test_cache_signature_stable_when_overrides_unchanged` — determinism
   - **`test_cache_signature_scoped_to_role_unrelated_override_does_not_bust`** — THE NEW GUARD: macro override change must NOT invalidate agronomist signature, but MUST invalidate macro signature
   - `test_cache_signature_fallback_when_role_is_none_folds_in_full_dict` — `role=None` falls back to full-dict hashing (safe-over-sorry)

4. ✅ **Defensive fallback is genuinely safer** — when `role` is unknown or not provided, the full overrides dict is hashed. This over-invalidates (any change to ANY override busts the cache) rather than risking silent cache collisions across role swaps. Right call for a cache-key change.

5. ✅ **Targeted eval suite**: `pytest tests/test_eval_harness.py tests/test_eval_case.py tests/test_eval_baseline.py tests/test_eval_lint.py tests/test_specialist_prompt.py tests/test_eval_models.py` → **102 passed in 8.52s**, including all 4 signature behavior tests.

6. ✅ **Full test suite**: 1576 passed in 8m13s. +2 from main baseline (the 2 new signature tests). No regressions.

7. ✅ **Branch is based on current `origin/main`** (parent commit `c117190b2` = #1767 merge HEAD).

**One-time cost is correctly disclosed:** the signature formula itself changed, so the first post-merge bootstrap will invalidate all 247 cached responses (~$1.24). After that, per-agent experiments are ~$0.12 each — a 10× cost reduction for the kind of model-swap experiment the upcoming Step 6B (agronomist Gemini Flash → Claude Haiku) will run.

**Useful diagnostic byproduct:** if the macro cache-churn cases (`kc_fed_*`, `sb_fed_on_hold`) flagged in #1767's follow-up section continue to churn after this lands, the instability is definitively NOT in `eval_overrides` — that narrows the investigation to the prompt itself, the registry, or the temperature path. If they stop churning, the per-agent scoping was the fix. Either way it's clearer signal than the current state.

**Post-merge baseline-refresh consideration is correct:** since fresh responses on a one-time refill of the cache may shift aggregate metrics enough to trip drift thresholds, suggesting "triage after the first weekly cron run" is the right cadence — best case nothing shifts, worst case capture a small second baseline. No upfront refresh forced.

Ready to merge.

PR #1770, opened 2026-04-24
🎨 Palette: Professionalize Agent Names on LLM Monitor

cristobalmoena:
💡 **What:** Professionalized the representation of AI agents on the LLM Monitor page.
🎯 **Why:** Previously, the "Cost by Agent Role" chart and "Fallback Chains" table displayed raw internal strings (e.g., `meteorologist_sentiment`, `technical_sentiment`). This change maps them to their canonical display names with emojis (e.g., `🌦️ Meteorologist`, `📉 Technical`), improving readability and aligning with the visual language used on the Cockpit and Council pages.
📸 **Visuals:** The "Cost by Agent Role" bar chart and the "Fallback Chains" data table now feature human-readable labels and icons.
♿ **Accessibility:** Improved clarity for all users by replacing technical identifiers with semantic, descriptive labels.

---
*PR created automatically by Jules for task [14036516912247483602](https://jules.google.com/task/14036516912247483602) started by @cristobalmoena*

PR #1769, opened 2026-04-24
refactor(debate-validation): extract check_quote_authenticity helper — Tier-3 evals PR-B (RFC Step 1)

rozavala:
RFC: \`docs/rfcs/tier3_evals.md\` Step 1 (PR-B — production-side refactor).

## Summary

Factor the quote-authenticity check out of \`TradingCouncil._validate_debate_output\` into a module-level helper at \`trading_bot/debate_validation.py\` so the Tier-3 eval harness (Step 2 / PR-A) can run the same check without re-implementing it. Per the RFC's \"two copies drift silently\" risk: one shared helper keeps the wrapper and the harness aligned.

**Zero behavior change** — all 15 existing \`_validate_debate_output\` tests pass unchanged. Blast radius is \`TradingCouncil\`'s debate-validation path only; the helper is additive.

### New: \`trading_bot/debate_validation.py\`

- \`check_quote_authenticity(quote, reports_text, source_agent=None, threshold=0.3) -> bool\`: two-tier scoping (named agent's section → fallback to full blob) mirroring \`agents.py:1476-1504\`. Returns False when fuzzy word overlap is below \`threshold\`; returns True for empty/no-word quotes (upstream handles missing-quote cases).
- \`parse_agent_sections(reports_text) -> dict[str, str]\`: public helper split out so the Tier-3 harness can surface sections for number-check scoping without duplicating the \`--- AGENT_NAME REPORT ---\` header regex.

### Refactored: \`TradingCouncil._validate_debate_output\` (\`trading_bot/agents.py\`)

- Quote check delegates to \`check_quote_authenticity\`. The two-tier scoping logic is gone from the method; replaced by a single call whose result maps to \"strip this arg\".
- Number-hallucination logic, JSON plumbing (\`_clean_json_text\`, \`json.loads\`, \`isinstance(dict)\`), logging, and \`weakest_point\` warnings all stay at this call site per the RFC's **Extraction boundary** rule (helper takes string inputs; JSON parsing stays at the call site).
- Section parsing for the *number* check still lives in the wrapper — they're needed locally for \`scope_numbers\`.

### Docs: \`docs/rfcs/tier3_evals.md\` status flip

- \`Status: Draft v3.2 → Accepted (v3.2)\` with 2026-04-24 approval date. RFC Step 0 complete; this PR is Step 1.

## Compatibility

- Method signature of \`_validate_debate_output\` unchanged.
- External callers (\`agents.py:1619\`, \`:1623\`) untouched.
[...]

PR #1773, opened 2026-04-24
evals: sort country set in render_macro for deterministic macro prompt

rozavala:
## Summary

Root-cause fix for the macro cache-churn behavior flagged in #1767 and #1768. **One-line fix** plus a regression test. 2 files, +59 / −1 lines.

```diff
- countries = set(r.country for r in profile.primary_regions)
+ countries = sorted({r.country for r in profile.primary_regions})
```

`set` iteration order is randomized per Python process via PYTHONHASHSEED. The rendered macro prompt converts `countries` into a comma-joined `currency_pairs` string. Different process → different order → different prompt text → different cache signature → fresh LLM call every bootstrap run.

## Diagnostic

Built the same case prompt (`macro/contrast.kc_fed_magnitude_007_a`) in five fresh subprocesses on the buggy code:

```
sha=f0a9fe24b82ba312 len=6867   # runs 1, 2
sha=f0a9fe24b82ba312 len=6867
sha=145162c50dac7c83 len=6867   # run 3 — DIFFERENT content, same length
sha=145162c50dac7c83 len=6867
sha=f0a9fe24b82ba312 len=6867
```

Diff narrowed to exactly one line:

```diff
< USD/ETB, USD/BRL, USD/COP, USD/VND
> USD/ETB, USD/VND, USD/BRL, USD/COP
```

Same currencies, different sequence. Set-iteration randomization.

After the fix, all five subprocesses produce `sha=f1f9e24efb7b345a`. Stable.

## Symptom in the cache

Pre-fix `evals/.response_cache/`:

| agent | cache files | cases | ratio |
|---|---|---|---|
| **macro** | **120** | **32** | **3.75x** |
| volatility | 96 | 32 | 3.0x (legitimate — Step 4 persona attempt invalidated all once) |
| inventory | 65 | 32 | 2.0x (Step 5A packet augmentations + Haiku JSON failures) |
| geopolitical | 64 | 32 | 2.0x |
| agronomist | 44 | 23 | 1.91x (Step 6B model swap attempt invalidated all once) |
| sentiment | 36 | 32 | 1.13x |
| **technical** | **32** | **32** | **1.0x** ✓ |
| **supply_chain** | **32** | **32** | **1.0x** ✓ |

Macro stands out as the agent with **chronic per-run churn** (the 6B run alone added 32 fresh macro signatures in <10 minutes despite scoped-signature signature scoping landing in #1768 — proving the variance was upstream of `_cache_signature`, in the prompt itself). Technical and supply_chain at exactly 1:1 are the healthy reference; their personas don't traffic in country sets.

## Production prompt impact

The production prompt now consistently lists currencies alphabetically:
- KC: `USD/BRL, USD/COP, USD/ETB, USD/VND`
- SB: same set, alphabetical
- ZW: alphabetical

Semantically identical — the model sees the same set of pairs — just stable. No behavior change for the trading system, just a one-time prompt-content alignment.

## Regression test

`test_render_macro_currency_pairs_are_deterministic_across_calls` in `tests/test_eval_harness.py` asserts the rendered `currency_pairs` is alphabetical regardless of `primary_regions` input order. Catches any future reintroduction of unsorted set iteration in the prompt-build path.

## Test plan

- [x] `pytest tests/test_eval_harness.py -q` → 29 passed (28 existing + 1 new regression test)
- [x] `pytest tests/test_eval_case.py tests/test_eval_harness.py tests/test_eval_baseline.py tests/test_eval_lint.py tests/test_specialist_prompt.py tests/test_eval_models.py -q` → 103 passed
- [x] `python -m evals.runner --lint-only` → 247 cases clean
- [x] **Empirical:** 5 subprocess invocations all produce identical SHA on the new code; same setup on the old code produced 2 distinct SHAs across the 5 runs.
- [ ] Post-merge: next bootstrap run should NOT generate any new macro cache files for cases that already have a cached response under the new (sorted-currencies) prompt. Fresh signatures will be written once for each case as the new prompt is hit, then stable thereafter.

## Follow-up (not this PR)

Existing 88 orphan macro cache files (120 − 32 = 88 stale signatures from random orderings) will linger until eviction. They don't affect correctness — current bootstrap will hit the new stable signature and ignore the old files. Worth a one-time housekeeping script if the cache dir gets unwieldy, but not blocking.


rozavala:
## Approving — beautifully diagnosed root-cause fix with empirical verification

**Scope:** one-line code fix (`set(...)` → `sorted({...})`) in `trading_bot/prompts/base_prompts.py:336`, plus a 54-line regression test. Closes the macro cache-churn behavior flagged in #1767/#1768 follow-ups.

**Verification I ran:**

1. ✅ **Bug class is real** — Python set iteration order is randomized per process via `PYTHONHASHSEED` (default behavior since Python 3.3+). The PR's empirical evidence — same `render_macro` call producing 2 distinct SHAs across 5 fresh subprocess runs on the buggy code, single SHA on the fixed code — pins this down precisely. Diff narrows to a single line:
   ```
   < USD/ETB, USD/BRL, USD/COP, USD/VND
   > USD/ETB, USD/VND, USD/BRL, USD/COP
   ```
   Same currencies, different sequence — exactly what set-iteration randomization produces.

2. ✅ **Symptom-side data confirms diagnosis:**
   - macro: 120 cache files / 32 cases = **3.75x inflation** ← worst, exactly the agent whose persona traffics in country sets
   - technical / supply_chain: 32/32 = **1.0x** ← reference baseline, no country sets in their personas
   The macro outlier with no other plausible explanation (scoped signatures landed in #1768, eliminating overrides as the variance source) → the prompt itself was the variance source. The PR proves it with subprocess SHA experiments.

3. ✅ **Production reach is the right shape** — `render_macro` is registered in `get_agent_prompt`'s renderer table (`base_prompts.py:515`), called from `agents.py:70` and `specialist_prompt.py:81`. So this DOES affect production prompts, but in the right direction:
   - **Before**: production prompt had random-order currency pair list (one of 4! = 24 possible orderings per Python process)
   - **After**: production prompt has alphabetical currency pair list (one specific ordering)
   - Net effect: production trading was *already* stochastic in this dimension across restarts; making it deterministic is strictly an improvement for reproducibility (backtest replay, forensic comparison, eval cache stability).

4. ✅ **Set contents unchanged** — `sorted({...})` produces the same elements as `set(...)`, just in alphabetical order. The model sees the identical set of currency pairs; only the listing order changes. Zero semantic shift in what's being asked of the LLM.

5. ✅ **Regression test pins the contract** — `test_render_macro_currency_pairs_are_deterministic_across_calls` constructs a profile with intentionally non-alphabetical input order (`Vietnam, Brazil, Colombia, Ethiopia`), renders the macro prompt, and asserts the rendered string contains `"USD/BRL, USD/COP, USD/ETB, USD/VND"` (alphabetical). If anyone ever reintroduces unsorted set iteration here, this test fails immediately with a clear diagnostic message.

6. ✅ **Targeted eval suite**: `pytest tests/test_eval_harness.py tests/test_eval_case.py tests/test_eval_baseline.py tests/test_eval_lint.py tests/test_specialist_prompt.py tests/test_eval_models.py` → **103 passed in 8.34s**.

7. ✅ **Full test suite**: 1594 passed (+1 new regression test) in 5m55s. No regressions. Same `test_jitter_seconds_stable_across_process_restarts` flake deselected.

8. ✅ **Branch on current `origin/main`** with #1768 + #1769 + #1770 integrated.

**Why this isn't a persona change in the cautioned sense:** the cached memory note about persona/model swaps being fragile is about altering *what the persona says* (vocabulary, instructions, decision frames) — Step 4 and 6B were both attempts to change persona behavior. This PR changes nothing about what the macro persona says or does; it only stabilizes the iteration order of an already-present set. The macro analyst sees the same currency pairs in the same prompt structure, just consistently. That's plumbing, not tuning.

**On the 88 orphan macro cache files (120 − 32):** correctly scoped out as housekeeping. They don't affect correctness — the new bootstrap will hit the new stable signature for each case once, then stay stable thereafter. A one-time eviction script could land separately if the cache dir gets unwieldy.

**Cross-PR diagnostic value:** this fix proves PR-#1768's scoped-signature change wasn't the macro-churn source (signatures were stable per-call after #1768; the variance was upstream in the prompt itself). Useful confirmation that #1768 did its job and macro churn was a separate, specific bug.

Ready to merge.

PR #1781, opened 2026-04-25
evals: DSPy direction-trainer infrastructure (no training run yet)

rozavala:
## Summary

Adds the trainer scaffold for eval-case-supervised prompt optimization. **No actual training run, no prompt files emitted, no production behavior change.** This PR ships the plumbing; the per-agent training PRs come after with pre-registered hyperparameters.

3 files, +1,358 / −14 lines:

| file | size | purpose |
|---|---|---|
| `evals/dspy_trainer.py` | ~640 lines | trainer + 3 CLI modes |
| `trading_bot/agents.py` | +50 lines | loader fallback to eval-trained namespace (no-op until first prompt ships) |
| `tests/test_eval_dspy_trainer.py` | ~370 lines | 25 invariant tests |

## What this trainer does differently from `trading_bot/dspy_optimizer.py`

| | Existing `dspy_optimizer.py` | New `evals/dspy_trainer.py` |
|---|---|---|
| Signal | Brier-weighted historical direction (price outcomes) | Direct match against case YAML `expect.direction` |
| Dataset | `enhanced_brier.json` + `council_history.csv` | `evals/splits/train.json` → resolved EvalCase objects |
| Metric | `_compute_metric_score` (contribution-weighted) | `direction_only_metric` via `check_case` filter |
| Module | `dspy.Predict` with `persona / market_context / task` synthetic signature | Custom `dspy.Module` calling `build_specialist_prompt` (matches production byte-for-byte) |
| Output namespace | `data/{TICKER}/dspy_optimized/{agent}/prompt.json` | `data/_eval_optimized/{agent}/prompt.json` |
| Granularity | per-ticker | agent-scoped |

The two pipelines share only ~30 lines: atomic write helper, agent name validation, optimized-prompt JSON output schema. They don't share a metric or a dataset class.

## Three CLI modes (mutually exclusive at argparse)

```
python evals/dspy_trainer.py --agent supply_chain --train
python evals/dspy_trainer.py --agent supply_chain --validate
python evals/dspy_trainer.py --agent supply_chain --smoke
```

- **`--train`** runs `BootstrapFewShot`, writes prompt + report. Reads `train.json` only.
- **`--validate`** read-only val report. Reads `val.json`, runs cached-mode eval, scores currently-deployed prompt. **No writes, no optimizer.**
- **`--smoke`** single-case plumbing check. No optimizer. Verifies LM dispatch + JSON parse before a $1+ training run.

Mutual exclusion is enforced at parse time as defense against muscle memory like `--train --validate` (an attempt to "validate during training") which is the exact discipline violation the train/val split was designed to prevent.

## Direction-only metric (single source of truth)

```python
from evals.runner.case import check_case

def direction_only_metric(case, response):
    if case.expect.direction is None and not case.expect.direction_in:
        return 1.0  # no direction signal → nothing to optimize against
    _, failures = check_case(case, response)
    direction_failures = [f for f in failures if f.check in ("direction", "direction_in")]
    return 0.0 if direction_failures else 1.0
```

Reuses the harness's `check_case` so direction matching, `direction_in` lists, BULLISH/BEARISH/NEUTRAL string normalization, and edge cases are not reimplemented in two places. Confidence/citation/hallucination checks stay in the val report (per-case + aggregate `secondary_check_pass_rates`) so calibration regressions are visible even when direction is fine — but they don't drive training. Direction is what we want DSPy to actually optimize.

## Custom `dspy.Module`

`EvalAgentModule.forward(case)` — direct port of the harness dispatch path, NOT a parameter swap on the existing `AgentPredictor`:

1. `build_specialist_prompt(...)` — same call the production / eval harness uses
2. `with dspy.context(lm=eval_tier_lm): raw = self.lm(prompt)` — per-call LM config (cleaner than mutating the global)
3. `_parse_json_response(raw)` — same parser the harness uses
4. Returns `dspy.Prediction(direction, confidence, evidence, analysis, _raw_response)` so the metric can call `check_case`

Per-agent eval-tier LM is resolved from `evals/config/eval_models.json`. Different agents with different providers (Flash, Haiku, gpt-4o-mini, grok-fast) all work — `dspy.context` localizes the LM config so sequential per-agent training doesn't require global reconfigures.

## Loader fallback (`trading_bot/agents.py`)

`TradingCouncil._load_optimized_prompt` now tries:

1. `data/dspy_optimized/{ticker}/{persona_key}/prompt.json` — per-ticker (existing, price-outcome trained)
2. `data/_eval_optimized/{persona_key}/prompt.json` — agent-scoped (new, eval-case trained)

Per-ticker wins when both exist. Documented as **PROVISIONAL precedence** in the docstring — the right policy depends on which signal proves stronger, which is exactly the question this trainer is built to answer. Logs at INFO which source was picked. The `use_optimized_prompts.{ticker}` toggle remains the kill switch — when False, neither namespace is consulted.

The fallback is a **no-op until eval-trained prompts exist on disk**, so this PR can't change production behavior. Tested both paths (per-ticker present, fallback to eval-trained, neither exists, malformed per-ticker).

## Failure-mode taxonomy + `do_not_ship` flag

Every training run writes a result JSON to disk for audit. Non-zero exit codes per failure mode:

| code | constant | meaning |
|---|---|---|
| 10 | `EXIT_TOO_FEW_DEMOS` | bootstrap couldn't find viable demos |
| 11 | `EXIT_DISPATCH_ERROR` | LM call failed (network, rate limit, quota) |
| 12 | `EXIT_ABSTENTION_HIGH` | val abstention exceeded ceiling |
| 13 | `EXIT_NO_IMPROVEMENT` | pre-registered ship threshold not met |
| 14 | `EXIT_REGRESSION` | val score worse than baseline |
| 0  | `EXIT_OK` | ship-eligible result |

`do_not_ship: bool` in the result JSON is the source of truth. Exit code is for shell pipelines.

## Pre-registered hyperparameters (for the upcoming first training PR)

For supply_chain (the cleanest agent at 28/32 baseline — chosen first per the design memo):

| flag | value | rationale |
|---|---|---|
| `--max-bootstrapped-demos` | 4 | DSPy default; few-shot saturates around 4-6 |
| `--max-labeled-demos` | 4 | matches above |
| `--ship-threshold` | 0.10 | 10pp val lift over baseline |
| `--abstention-ceiling` | 0.50 | don't ship if >50% NEUTRAL on val |
| `--min-train-cases` | 10 | bootstrap won't find signal below this |
| seed | 42 | matches the train/val split seed |

For agronomist (last per design memo), `--ship-threshold 0.15` instead of 0.10 — agronomist's val set is ~10 cases; on a binary metric the standard error is ±15pp; a 5pp improvement is indistinguishable from sampling noise. The pre-registered threshold makes "DSPy didn't beat Pattern A" a perfectly fine outcome (which is what the memory says we should expect).

## Reproducibility note

"Same seed produces same output" is aspirational on a stochastic LM. Even at temperature=0, providers generally aren't bit-identical run-to-run. The reproducibility we test is: same seed → same case ordering, same shuffle, same selected demo *indices*, modulo LM stochasticity in the bootstrap-generated outputs. There's no test that fails when a provider has nondeterminism — that would be a flake factory.

## What this PR does NOT do

- **No training run.** No `dspy.compile()` invocation; no LLM spend.
- **No prompt files.** `data/_eval_optimized/` will be empty until the first per-agent training PR ships.
- **No CI integration.** The trainer is a script, not a workflow step.
- **No production behavior change.** Loader fallback only activates when eval-trained prompts exist, which they don't yet.

## Test plan

- [x] `pytest tests/test_eval_dspy_trainer.py -v` → **25 passed**
- [x] `pytest` (full eval suite: `tests/test_eval_*` + `test_specialist_prompt`) → **137 passed**
- [x] `python -m evals.runner --lint-only` → **248 cases clean**
- [x] CLI argparse behaviors:
  - `--train --validate` → `SystemExit` (mutually exclusive)
  - `--train --smoke` → `SystemExit` (mutually exclusive)
  - no mode flag → `SystemExit` + usage line
  - no `--agent` → `SystemExit` + usage line
- [x] Loader fallback: per-ticker present (wins), only fallback present (returns it), neither (returns None), per-ticker malformed (falls back gracefully)

## Rollout plan

1. **This PR** — trainer infra. Loader fallback no-op until first prompt ships.
2. **Smoke PR** — `python evals/dspy_trainer.py --agent supply_chain --smoke`, paste output into PR. Confirms LM dispatch + JSON parse work. Cost ~\$0.01.
3. **First training PR** — supply_chain `--train` + `--validate`. Pre-registered hyperparameters from this PR's table. Eval gate runs in cached mode and catches any regression.
4. **Per-agent PRs** — in increasing-difficulty order, ending with agronomist. Each labeled `eval:prompt-optimization:{agent}`.


rozavala:
## Approving — clean trainer scaffold with strong discipline against the failure modes the memory predicts

**Scope:** trainer infrastructure only. **No training run, no prompts emitted, no production behavior change** until follow-up PRs ship eval-trained prompts. The author has explicitly internalized that "DSPy didn't beat Pattern A" is a likely outcome for some agents (and pre-registered the threshold to prove it cleanly).

**Verification I ran on the production-touched file (`trading_bot/agents.py`):**

The `_load_optimized_prompt` change is genuinely backward-compatible:

1. ✅ **Refactor extracts** a single-source loader into a two-source loader via the new `_read_optimized_prompt_file` helper. The helper is `@staticmethod` and has no side effects beyond reading + JSON-parsing one file.
2. ✅ **Per-ticker source preserved** — same path computation (`{prompts_dir}/{ticker}/{persona_key}/prompt.json`), same precedence (checked first), same return shape. Existing per-ticker prompts continue to load identically.
3. ✅ **Eval-trained source is genuinely no-op** — `data/_eval_optimized/{persona_key}/prompt.json` doesn't exist anywhere in the repo today, and the trainer that would write it isn't run by this PR. Confirmed by grep — no other writer to that namespace.
4. ✅ **Kill switch preserved** — docstring explicitly notes `use_optimized_prompts.{ticker}` toggle is checked upstream in `research_topic` (verified at `agents.py:1155-1158`); when False, `_load_optimized_prompt` isn't called at all.
5. ✅ **Cache key unchanged** — `f"{ticker}/{persona_key}"` continues to key the in-memory cache, so loading order and cache-hit shape are preserved.
6. ✅ **PROVISIONAL precedence policy documented** — the docstring is honest about the design uncertainty: "the question of which should win when both produce a prompt is genuinely contested. For now the per-ticker pipeline wins (existing behavior). Flip if the eval-trained signal proves consistently stronger." That's the right framing — don't pretend the policy is settled.
7. ✅ **INFO-level logs name the source** — when something is loaded, the log says whether it came from per-ticker or eval-fallback. Production observability for debugging which prompt is in flight.

**Verification I ran on the trainer module:**

1. ✅ **CLI mutual exclusion enforced at argparse** — `--train`, `--validate`, `--smoke` are mutually exclusive at parse time. Catches `--train --validate` muscle memory before it can violate the train/val discipline.
2. ✅ **`--validate` is read-only** — explicitly documented as "no writes, no optimizer." That's the discipline guard.
3. ✅ **`--smoke` is single-case plumbing-only** — verifies LM dispatch + JSON parse before a $1+ training run. Useful pre-flight check.
4. ✅ **Direction-only metric reuses `check_case`** — `evals/runner/case.py::check_case` is the harness's source of truth for direction matching; the trainer reuses it via a small filter for `direction`/`direction_in` failures only. Avoids a parallel implementation that would silently drift.
5. ✅ **Pre-registered hyperparameters** with rationale per agent — including the explicit "agronomist threshold = 0.15 because val set is ~10 cases and standard error is ±15pp; a 5pp lift is sampling noise." That sentence alone shows the author has ingested the memory's warning.
6. ✅ **Failure-mode taxonomy** — exit codes 10-14 + `do_not_ship` flag in result JSON. Shell-pipeline + persistent-record dual signal. Right level of structure for an experiment that's expected to often fail to ship.
7. ✅ **Custom `dspy.Module` matches production byte-for-byte** — calls `build_specialist_prompt` (the same function production uses), uses `dspy.context(lm=...)` for per-call LM config (cleaner than mutating global state, supports sequential per-agent training without reconfigures).

**Verification I ran on tests:**

1. ✅ **Targeted suite** (`pytest tests/test_eval_dspy_trainer.py`): 25 passed.
2. ✅ **Eval-adjacent suite** (per PR description): 137 passed.
3. ✅ **Full test suite**: 1772 passed in 6m24s. No regressions. Same `test_jitter_seconds_stable_across_process_restarts` flake deselected.
4. ✅ **CLI argparse behaviors** verified by tests: mutual-exclusion, no-mode-flag, no-`--agent`.
5. ✅ **Loader fallback paths** verified: per-ticker present (wins), only fallback present (returns it), neither (returns None), per-ticker malformed (falls back gracefully).
6. ✅ **Branch on current `origin/main`** (parent `0ca45219f` = #1779 merge).

**Why this is safe even if the training experiments fail:**

- The infrastructure PR can't ship a bad prompt — the prompt files don't exist yet, and writing them requires a separate `--train` invocation with a `--ship-threshold` that the author won't lower.
- Each per-agent training PR is a separate review surface where the verification report (val pass rate, abstention rate, do_not_ship flag) lands as part of the PR description.
- If a training run produces a regression, it's caught by the val report (exit 14 = `EXIT_REGRESSION`) and the prompt isn't shipped.
- Even if a bad prompt somehow shipped, the `use_optimized_prompts.{ticker}` kill switch flips it off without redeploy.

**Reproducibility note is honest:** "same seed produces same output" is aspirational on a stochastic LM. The author calls this out explicitly and tests only the deterministic dimensions (case ordering, shuffle, demo indices) rather than fabricating a reproducibility test that would be a flake factory.

Ready to merge.

PR #1780, opened 2026-04-25
feat(evals): Tier-3 canaries + CLI integration + Tier-2 reference validation — PR-A.3.4.a of RFC Step 2

rozavala:
Seventh slice of Tier-3 eval infrastructure. RFC: \`docs/rfcs/tier3_evals.md\` Step 2 (PR-A).

**Lands the Step 2 exit criterion.** Per the RFC: \"schema + loader + dispatch + one canary case per subtype all green in mock mode.\" After this, Step 3 (case authoring) is unblocked.

## End-to-end smoke

\`\`\`
$ python -m evals.runner --mode mock
Loaded 248 eval cases from evals/cases
Loaded 6 Tier-3 eval cases from evals/cases
Lint passed: 248 Tier-2 case(s) clean; 6 Tier-3 case(s) loaded
Wrote report: evals/runs/.../summary.md
Run summary: 254 cases, 0 failed
\`\`\`

### 6 canary YAMLs

One per subtype + the DA short-circuit canary:

- \`evals/cases/master/canary.yaml\` — \`canonical_consensus_bullish\`
- \`evals/cases/permabear/canary.yaml\` — \`canonical_bear_ammo\`
- \`evals/cases/permabull/canary.yaml\` — \`canonical_bull_ammo\`
- \`evals/cases/compliance/canary.yaml\` — \`canonical_approve\`
- \`evals/cases/devils_advocate/canary.yaml\` — \`canonical_substantive_objection\`
- \`evals/cases/devils_advocate/canary_neutral_skip.yaml\` — \`canonical_neutral_skip\`, exercises the no-LLM short-circuit

Adversarial Tier-3 cases ship in Step 3 onwards — RFC notes mock mode has near-zero gate value for adversarial Tier-3 (synthesized-from-expect responses trivially pass "did the LLM take bait" tests).

### Tier-2 loader skips Tier-3 files

\`evals/runner/case.py::load_cases\` peeks each YAML and skips files with a \`subtype\` field. Before this, a Tier-3 YAML under \`evals/cases/<subtype>/\` would crash the Tier-2 load with a \`ValueError\` on missing required fields. Per-PR cost: one cheap YAML load per file, only on the Tier-2 path.

### CLI integrates Tier-3

\`evals/runner/__main__.py::_run\` now:

1. Loads Tier-2 cases via \`load_cases\`
2. Loads Tier-3 cases via \`tier3_case.load_cases\` (filtered by \`--agent\` reused as subtype, plus \`--commodity\`)
[...]

Issue #1740, opened 2026-04-23
bug: xAI rate-limiter slot timeouts (30s) on supply_chain / permabear during 14:29-14:30


rozavala:
## Context

New failure class not previously tracked. During the 14:29-14:30 UTC window on 2026-04-23, the \`trading_bot.rate_limiter\` emitted **\`Rate limit timeout for xai after 30.0s\`** twice, each followed by \`heterogeneous_router\` skipping to fallback.

## Evidence

\`/home/rodrigo/real_options/logs/orchestrator_multi.log\`:

\`\`\`
3175: 2026-04-23 14:29:57,606 - trading_bot.rate_limiter - WARNING - Rate limit timeout for xai after 30.0s
3176: 2026-04-23 14:29:57,606 - trading_bot.heterogeneous_router - WARNING - Rate limit slot timeout for xai on role supply_chain. Skipping to fallback.
3177: 2026-04-23 14:30:01,885 - trading_bot.rate_limiter - WARNING - Rate limit timeout for xai after 30.0s
3178: 2026-04-23 14:30:01,886 - trading_bot.heterogeneous_router - WARNING - Rate limit slot timeout for xai on role permabear. Skipping to fallback.
3190: 2026-04-23 14:30:28,550 - trading_bot.router_metrics - WARNING - Fallback event: permabear xai -> openai. Reason: Rate limit timeout for xai
3208: 2026-04-23 14:30:47,357 - trading_bot.router_metrics - WARNING - Fallback event: supply_chain xai -> gemini. Reason: Rate limit timeout for xai
\`\`\`

Both events are **our local rate-limiter failing to acquire a slot within 30s**, not xAI returning a 429. That means our own concurrency accounting is over-committed — the limiter is queueing requests beyond what the slot pool can drain in 30s.

## Behaviour

- Role \`supply_chain\` fell back \`xai → gemini\`.
- Role \`permabear\` fell back \`xai → openai\`.
- The temporal correlation with the Perplexity timeout burst (#1739) at exactly the same minute window suggests a **shared trigger** — probably a KC + ZW scheduled-cycle overlap flooding the limiter and the grounded-data path simultaneously.

## Impact

- **Heterogeneity preserved**: diversity maintained since fallbacks route to different providers.
- **Signal quality risk**: permabear going from xAI (reasoning model) → OpenAI changes the adversarial-attack character of the debate in a way that may not be captured in Brier calibration.
[...]

PR #1772, opened 2026-04-24
feat(evals): extract shared hallucination detector — PR-A.2 of RFC Step 2

rozavala:
Second slice of Tier-3 eval infrastructure. RFC: \`docs/rfcs/tier3_evals.md\` Step 2 (PR-A).

Independent of PR-A.1 (#1771) — doesn't touch the Tier-3 schema or loader, just reshapes Tier-2's existing hallucination detector so both tiers can call the same code.

## Summary

Consolidate year / percentage / money hallucination detection into a shared helper at \`evals/runner/hallucination.py\`, exposing the signature the RFC specifies:

\`\`\`python
def detect_hallucinations(response_text: str, source_text: str) -> List[str]
\`\`\`

Tier-2's \`_detect_hallucination\` in \`evals/runner/case.py\` becomes a thin wrapper that extracts the packet corpus (\`raw_findings\` + \`extracted_facts\` + \`source_urls\`) and delegates to the helper. Tier-3's harness (PR-A.3) will call the helper with its own assembled \`reports_text\` — no re-implementation, no drift.

**Zero behavior change** — 53 existing Tier-2 hallucination / eval tests pass unchanged.

## Design

- **Strings only.** The RFC calls for \`response_text\` + \`source_text\` strings; callers own their corpus assembly because the source shapes differ (packet for Tier-2, assembled reports for Tier-3). Keeps the helper's surface tiny.
- **All four support helpers move.** \`_parse_percentages\`, \`_number_in_corpus\`, \`_percentage_is_derived\`, \`_DERIVATION_MARKERS\`, \`_VALUE_WITH_UNIT\` — all private to the helper module now. \`case.py\` no longer imports them.
- **Corpus lower-casing moves inside the helper.** Eliminates a class of bug where a caller feeds mixed-case source and silently loses matches.
- **Derivation escape hatch preserved exactly.** The \"both inputs must appear in corpus\" rule still closes the \`a=fabricated, b=fabricated, ratio=derived\` loophole — shown-work protects the *arithmetic*, not the numbers themselves.

## Changes

- **\`evals/runner/hallucination.py\`** (new, 189 lines): public \`detect_hallucinations\`, four private helpers, full docstrings mirroring the Tier-2 originals.
- **\`evals/runner/case.py\`**: -189 / +12. \`_detect_hallucination\` shrinks from 68 lines to 33 (now a pure adapter); all four supporting helpers removed.
[...]

PR #1783, opened 2026-04-25
fix(evals): trainer skips Tier-3 cases via load_cases() (smoke regression)

rozavala:
## Summary
- Smoke run `python -m evals.dspy_trainer --agent supply_chain --smoke` crashed at case-resolve time after PR #1780 (Tier-3 canaries) and PR #1781 (DSPy trainer) merged together. Root cause: the trainer's `_resolve_case_ids_to_objects` did a manual `glob` + `load_case()` walk and tripped on `evals/cases/compliance/canary.yaml` (Tier-3 schema, no Tier-2 required fields).
- Defer to `evals.runner.case.load_cases`, which already has the peek-and-skip for Tier-3 files (see `evals/runner/case.py:239`). One source of truth for the loader contract.
- Pinned with `test_resolve_case_ids_skips_tier3_cases` so a future regression fails loudly instead of $1+ into a training run.
- Documented the `python -m evals.dspy_trainer` invocation form in the module docstring — the direct-script form (`python evals/dspy_trainer.py`) puts `evals/` at `sys.path[0]` and shadows the package, breaking every `from evals.runner.case import …` lazy import.

## Context
This was caught by the pre-registered smoke step in #1781's rollout plan — exactly what smoke is for. The fix lands as a small standalone PR (per project convention) before the supply_chain training PR.

After this lands, smoke reaches actual LM dispatch end-to-end. Today's dispatch returned an xAI monthly-spending-limit `RateLimitError` (separate from #1740 slot timeouts), which is environmental, not a code defect — flagging here so you can resolve billing before the actual training run.

## Test plan
- [x] `pytest tests/test_eval_dspy_trainer.py -q` → 26 passed
- [x] `python -m evals.dspy_trainer --agent supply_chain --smoke` reaches LM dispatch (no more case-load crash)
- [ ] Re-run smoke once xAI credits are restored


rozavala:
## Approving — exemplary discipline pattern: smoke-caught, cheaply fixed, regression pinned

**Scope:** 62/8 across 2 files. Tiny, surgical fix for a real interaction bug between PR #1780 (Tier-3 canaries) and PR #1781 (DSPy trainer).

**The bug shape (verified):**

The trainer's `_resolve_case_ids_to_objects` did:
```python
for path in sorted(CASES_ROOT.glob("*/*.yaml"), key=str):
    c = load_case(path)  # ValueError on Tier-3 schema
```

After PR #1780 added Tier-3 canary YAMLs at `evals/cases/<subtype>/*.yaml`, the glob picks them up and `load_case()` raises `ValueError` because Tier-3 cases lack the Tier-2 required fields (`agent`, `category`, `search_instruction`).

**The fix is right:**

Defer to `evals.runner.case.load_cases()`, which already has the peek-and-skip for Tier-3 files (added by PR #1780 at `evals/runner/case.py:239`). This consolidates filter logic into one place — exactly the "single source of truth" pattern PR-B (#1769) and PR-A.2 (#1772) established for the helper extractions. Avoids the same Tier-3 detection logic existing in two places.

**Verification I ran:**

1. ✅ **Diff is clean** — replaces 6-line manual walk with 3-line `load_cases()` call + filter. Net -8 / +19 (most of the +19 is updated docstring explaining why).

2. ✅ **Module docstring update is genuinely useful** — documents the `python -m evals.dspy_trainer` invocation form and explicitly calls out that `python evals/dspy_trainer.py` puts `evals/` at `sys.path[0]` and shadows the package, breaking every `from evals.runner.case import …` lazy import. This is a non-obvious shell trap that's easy to hit and silently breaks the trainer.

3. ✅ **Regression test pins the contract** — `test_resolve_case_ids_skips_tier3_cases` asserts:
   - Resolution against the real case library doesn't raise (was the bug shape)
   - No Tier-3 agent ID (`compliance`/`devils_advocate`/`master`/`permabear`/`permabull`) leaks into the resolved Tier-2 set
   - Length sanity check (each train ID resolves to exactly one case)
   The diagnostic message points future debuggers at the right fix: "use load_cases() (which filters Tier-3 via the subtype peek), not a direct case file walk."

4. ✅ **Targeted suite**: `pytest tests/test_eval_dspy_trainer.py` → 26/26 pass.

5. ✅ **Full test suite**: 1780 passed in 7m41s. No regressions.

6. ✅ **Branch on current `origin/main`** (parent `35b79cfa5` = #1781 merge).

**Process discipline observation — this is exactly what smoke is for:**

The PR description correctly notes this was caught by the pre-registered smoke step in #1781's rollout plan (Step 2: "`python evals/dspy_trainer.py --agent supply_chain --smoke`"). The bug surfaced *before* a $1+ training run, which is the whole point of having a smoke step. Lands as a small standalone PR before the actual supply_chain training PR — that's the right shape for a smoke-caught regression.

**Environmental note appropriately disclosed:** the PR mentions today's smoke also surfaced an xAI monthly-spending-limit `RateLimitError` (separate from #1740 slot timeouts) — flagged as environmental, not a code defect, with a checkbox to re-run smoke once xAI credits are restored. Right level of disclosure.

Ready to merge.

PR #1784, opened 2026-04-25
refactor(master-prompt): extract build_master_prompt helper — production-side prep for Tier-3 live mode (PR-A.3.4.c)

rozavala:
Production-side refactor that unblocks Tier-3 live mode in the next eval-harness PR. RFC: \`docs/rfcs/tier3_evals.md\` Step 2 (PR-A).

Mirrors the PR-B (#1769) precedent — that one extracted \`check_quote_authenticity\` from \`TradingCouncil._validate_debate_output\` so the eval harness and production could share the same fuzzy-match code without two copies drifting. Same pattern here for the Master prompt template.

## Why now

The next eval-harness PR (A.3.4.d) wires Tier-3 live mode for the Master subtype. To do that without re-implementing the prompt template (drift risk), production needs to expose the template as a callable helper. This PR is that extraction. Zero behavior change.

### \`trading_bot/master_prompt.py\` (new)

\`build_master_prompt(...)\` renders the Master Strategist prompt. Takes string inputs only — JSON serialization of debate outputs, market-data formatting, and persona loading all stay at the call site per the RFC's **Extraction boundary** rule. Function signature:

\`\`\`python
def build_master_prompt(
    *,
    master_persona: str,
    contract_name: str,
    market_context: str,
    market_data_str: str,
    reports_text: str,
    bear_json: str,
    bull_json: str,
    urgent_context: str = "",
    sentinel_briefing: str = "",
) -> str:
\`\`\`

Body is byte-equivalent to the inline f-string at \`agents.py:1918-1945\` pre-extraction.

### \`trading_bot/agents.py\` (modified)

\`TradingCouncil.decide\` replaces the 27-line inline f-string with a call to \`build_master_prompt(...)\`. All inputs assembled the same way — production still loads persona / formats market_data / runs the debate before calling. The only thing that moved is the template body.

## What this PR does NOT change

- Master prompt **content** — it's byte-equivalent, locked by test
- Production call sites for \`decide\` — same signature, same outputs
- Debate / Compliance / DA prompt-builders — those land separately when their respective subtypes get live-mode wiring

## Test plan

- [x] \`pytest tests/test_master_prompt_byte_equivalence.py\` — **3/3 pass** in 17s.
  - Helper output character-equal to an independent inline reproduction (non-empty fields and empty-optional-fields versions).
  - Decide's source imports \`build_master_prompt\` (catches the failure mode where the extraction is reverted but the helper stays).
[...]

PR #1789, opened 2026-04-25
refactor(evals): trainer = deterministic top-K demo selector (drops DSPy)

rozavala:
## Summary

Replaces PR #1781's BootstrapFewShot trainer with a deterministic top-K demo selector. Honest about what the pipeline actually does (picks K exemplars), removes the DSPy dependency, ships demos in production's expected schema. Validated end-to-end on agronomist.

## Why

PR #1781 hit three blockers when actually run end-to-end (caught by the smoke step in its own rollout plan):

1. **`EvalAgentModule` had no `dspy.Predict` node.** The deliberate "byte-equivalent to production prompts" framing left BootstrapFewShot with nowhere to attach demos. The "Bootstrapped 4 traces" log line was a silent no-op.
2. **BootstrapFewShot doesn't update instructions** — it only generates demos. So the only signal we'd have been optimizing was demo selection.
3. **Demo schema mismatch.** Trainer serialized demos as `{"input_case_id": ..., "expected_direction": ...}`. Production's `format_demos` expects `{"input": {"market_context": ...}, "output": {"direction": ..., "confidence": ..., "analysis": ...}}` — would have rendered every field as "N/A".

(1) + (3) were broken; (2) reduces BootstrapFewShot to demo selection anyway. So we collapse to that and skip the DSPy ceremony that buys nothing here.

## How it works now

```
1. Load train cases for agent
2. Run each through cached EvalHarness (stays ~$0 in steady state)
3. Keep score==1.0 (direction-correct) cases
4. Sort by LM-reported confidence DESC; tiebreak by case id ASC
5. Take top K (default 4)
6. Serialize in production format; write empty instruction so the
   commodity-aware persona stays and demos append
```

If demo signal proves useful in production, MIPROv2 (which actually optimizes instructions) is a real next step — but no point spending on it before verifying demos help.

## Production loader change

`trading_bot/specialist_prompt.py`: demos now work without `instruction`. Previously gated; now the resolved (commodity-aware or legacy) persona is preserved and demos are appended. Provenance source becomes `dspy_optimized_demos_only` when this branch fires.

## End-to-end agronomist run

```
14 train cases → 12 direction-correct (85.71%)
4 demos written (top by LM confidence)
Baseline (cached, full set): 75%
Lift on train: 10.7pp
Pre-registered ship threshold for agronomist: 15pp
Result: do_not_ship: true, exit_code 13
```

The two cases that scored 0.0: `agronomist.contrast.kc_phenology_007_b` and `agronomist.contrast.sb_region_weight_007_b` — both flipped contrast `_b` cases where the model expected NEUTRAL but predicted directional. Honest signal about where the agronomist persona is weakest.

## Known gap (tracked for follow-up)

`--validate` mode doesn't actually load the trained prompt — `EvalHarness._run_tier2` always passes `optimized_prompt=None` to `build_specialist_prompt`. So today, validate re-measures the baseline-on-val. This wasn't introduced by this PR (it's pre-existing), but it does mean we can't yet measure trained-vs-baseline on val. Out of scope here; will need a separate PR that threads optimized_prompt through the harness for the validate path.

## Test plan
- [x] `pytest tests/test_eval_dspy_trainer.py tests/test_eval_splits.py tests/test_specialist_prompt.py tests/test_prompt_trace.py tests/test_master_prompt_byte_equivalence.py tests/test_base_prompt_rules.py -q` → 96 passed
- [x] `python -m evals.dspy_trainer --agent agronomist --train --max-demos 4 --ship-threshold 0.15 --abstention-ceiling 0.50 --min-train-cases 10` ran end-to-end, wrote `data/_eval_optimized/agronomist/prompt.json` with valid demos, returned exit 13 with the expected `do_not_ship` reason
- [ ] After merge: re-run for other agents with direction headroom (geopolitical, macro, volatility — each ~8pp baseline gap, will likely also `do_not_ship` at the default 10pp threshold; that's expected)
- [ ] Follow-up PR: thread `optimized_prompt` through harness so `--validate` can measure trained-prompt val performance


rozavala:
## Approving — honest correction following exactly the discipline pattern the memory predicted

This is the right outcome. The smoke-step in #1781's rollout plan caught three real defects in the DSPy trainer; rather than paper over them, this PR collapses to the only thing the trainer was actually doing (demo selection) and validates end-to-end against a real agent. The agronomist run hitting `do_not_ship` at exit code 13 — because the train-set lift was 10.7pp vs the 15pp pre-registered ship threshold — is the discipline gate firing exactly as designed.

**Verification I ran on the production code change:**

The `trading_bot/specialist_prompt.py` diff is small and load-bearing — let me trace each compatibility shape:

| Payload shape | OLD behavior | NEW behavior | Notes |
|---|---|---|---|
| `{"instruction": "x", "demos": [{...}]}` | instruction loaded, demos appended, source=`dspy_optimized` | identical | existing dspy_optimizer.py output |
| `{"instruction": "x"}` (no demos) | instruction loaded, no demo branch | identical | existing instruction-only path |
| `{"demos": [{...}]}` (no instruction) | **demos silently dropped** (gated under `if instruction`) | demos appended to resolved persona, source=`dspy_optimized_demos_only` | **new path** — what this PR's trainer writes |
| `{}` / `{"instruction": "", "demos": []}` | falls through, defaults | identical | edge case |

The only behavior change is enabling the demos-only path that didn't exist before. No existing payload shape changes its rendering. Confirmed by reading the new `_resolve_persona_prompt` flow:

```python
if instruction:
    persona_prompt = instruction
    prompt_source = "dspy_optimized"
demos = optimized_prompt.get("demos") or []
if isinstance(demos, list) and demos:
    demo_count = len(demos)
    if not instruction:
        prompt_source = "dspy_optimized_demos_only"  # only fires for new shape
    demo_text = format_demos(demos)
    persona_prompt = f"{persona_prompt}\n\n…"
```

Provenance tag (`dspy_optimized_demos_only`) makes the source auditable in `prompt_trace.py` rather than silently bucketed under `dspy_optimized`. Right call.

**Verification I ran on the trainer rewrite:**

The three blockers the PR identifies are real:

1. **`EvalAgentModule` had no `dspy.Predict`** — confirmed by reading PR #1781's `EvalAgentModule.forward` (it called `build_specialist_prompt(...)` then `self.lm(prompt)` directly, with no DSPy primitive in the path). BootstrapFewShot needs a `dspy.Predict` to attach demos to. Without one, "Bootstrapped 4 traces" is silent no-op truth.
2. **BootstrapFewShot only generates demos** — that's correct per DSPy docs. Instruction optimization needs MIPROv2.
3. **Demo schema mismatch** — verified against `format_demos` at `trading_bot/specialist_prompt.py:40-62`. Production reads `inp.get('market_context', 'N/A')`, `out.get('direction', 'N/A')`, `out.get('confidence', 'N/A')`, `out.get('analysis', 'N/A')`. PR #1781's `{"input_case_id": ..., "expected_direction": ...}` shape would have rendered every field as "N/A". This was a real bug.

So collapsing to "deterministic top-K" — load train cases → cached eval → keep score==1.0 → sort by LM confidence DESC, tiebreak by case id ASC → take K → write in production schema — is the honest minimum. If MIPROv2 (which actually optimizes instructions) ever justifies the spend, that's a separable next step; no point paying for it without first proving demos help.

**Verification I ran on tests:**

1. ✅ **Targeted suite**: `pytest tests/test_eval_dspy_trainer.py tests/test_eval_splits.py tests/test_specialist_prompt.py tests/test_prompt_trace.py tests/test_master_prompt_byte_equivalence.py tests/test_base_prompt_rules.py` → 96/96 passed.
2. ✅ **Full test suite**: 1806 passed in 6m06s. No regressions. Same `test_jitter_seconds_stable_across_process_restarts` flake deselected.
3. ✅ **End-to-end agronomist** quoted in PR: 14 train cases → 12 direction-correct (85.71%) → 4 demos written → train lift 10.7pp vs 15pp threshold → `do_not_ship: true, exit_code 13`. The fail-mode is the correct one and the system stayed safe.
4. ✅ **Branch on current `origin/main`** (parent `c1e7c2d12` = #1784 merge).

**On the matched memory note:**

The cached memory ("Eval-harness persona tuning + model swaps are fragile — Step 4 and 6B both rejected; treat persona/model changes as costly experiments expected to fail") predicted exactly this outcome. The agronomist experiment "failed" in the controlled sense — train lift was real but below threshold, so nothing shipped, and we now know agronomist's two contrast `_b` failure modes (`kc_phenology_007_b`, `sb_region_weight_007_b`) are persona-level rather than demo-fixable. That's useful diagnostic signal for the next investigation lane (Pattern A persona content), and exactly what the discipline gate is meant to surface.

**Honest disclosure of the validate-mode gap:** the PR explicitly calls out that `--validate` doesn't yet load the trained prompt (`EvalHarness._run_tier2` always passes `optimized_prompt=None`), so today's validate measures baseline-on-val. This is pre-existing, not introduced by this PR, and tracked as a follow-up. Right level of disclosure rather than silently shipping with a known measurement gap.

Ready to merge.

PR #1794, opened 2026-04-25
feat(evals): wire live_resolver for Tier-3 path-4 use_tier2_case — PR-A.3.4.k of RFC Step 2

rozavala:
## Summary

Closes the last harness loose end from RFC Step 2 — deferred since PR-A.3.1. A Tier-3 case author can now write ``specialist_inputs.{agent}.use_tier2_case: agronomist.canonical_bullish.001`` and the harness runs that referenced Tier-2 case through the same harness, feeding the response into the Tier-3 prompt as the agent's report.

- New ``EvalHarness(cases_root=...)`` parameter, threaded from the CLI's existing ``--cases-root`` flag.
- New ``_resolve_path4_inputs(case)`` async pre-fetch — walks ``specialist_inputs``, dispatches each ``use_tier2_case`` ref via ``_run_tier2`` (same llm_mode as the parent), and de-duplicates refs within a single Tier-3 case.
- ``_run_tier3`` pre-fetches first, then passes a sync ``lambda ref: prefetched[ref]`` to ``resolve_specialist_inputs`` so ``tier3_assemble.py`` stays pure / synchronous.
- Adversarial cases short-circuit the pre-fetch (the adversarial guardrail in ``resolve_specialist_inputs`` rejects path-4 anyway; running the Tier-2 case would burn an LLM call for nothing).
- Failure modes (missing cases_root, missing Tier-2 case, failed Tier-2 dispatch) all surface as ``specialist_inputs_resolution`` structured failures, not crashes.

## What's left after this

This closes Step 2's harness deliverables. Remaining for Step 2 to fully exit:

- **Step 2 calibration pass** — 5-case Master live run on ``claude-sonnet-4-5`` and ``gemini-3-pro-non-preview``, write ``calibration.json``, validate $0.12/case ceiling. This unblocks Step 3 (canonical case authoring).

## Test plan

- [x] ``pytest tests/test_harness_tier3_live_resolver.py`` — 6 new tests cover happy path, dedup, adversarial short-circuit, missing cases_root, missing Tier-2 case, and override-only cases (regression guard for path-1/2/3/5 — no cases_root needed).
- [x] ``pytest tests/ -k 'tier3 or harness'`` — 211 passed, no regressions.
- [x] Existing ``test_use_tier2_case_without_resolver_yields_structured_failure`` updated and renamed to match the new ``cases_root`` error path.


rozavala:
**Approved — closes Step 2's last harness deliverable. Live path-4 resolution wired correctly.**

**Implementation correctness:**
- ✅ `cases_root` parameter optional (defaults to `None`); path-1/2/3/5 only runs don't require it. Backward-compat preserved.
- ✅ `_resolve_path4_inputs` walks `specialist_inputs` looking only for path-4 refs — skips entries with `override` (path 1), `snapshot: frozen` (path 2), `stub_neutral` (path 3). Matches the precedence order from `resolve_specialist_inputs`.
- ✅ Adversarial detection: `case.case_kind.startswith("adversarial_")` short-circuits before any Tier-2 dispatch. Correctly handles the RFC §Adversarial case requirement that path-4 is forbidden on adversarial cases — burning an LLM call only to have `resolve_specialist_inputs` reject the case downstream would be the wrong behavior, and the `spy.await_count == 0` test pins this.
- ✅ De-dup via `if ref in prefetched: continue` — two agents pointing at the same Tier-2 ID share one dispatch. Material in cached/live mode where each Tier-2 dispatch is a real LLM call.
- ✅ Failure paths all surface as structured `specialist_inputs_resolution` CaseFailure: missing `cases_root`, missing Tier-2 case (FileNotFoundError → ValueError), failed Tier-2 dispatch (no direction in response). None crash the run.
- ✅ Resolved dict shape `{sentiment, confidence, evidence, analysis}` matches the production report format that `resolve_specialist_inputs`' validator expects. `CaseResult.direction` → `sentiment`, `CaseResult.confidence` → `confidence` (defaults to 0.0 if None), evidence/analysis flow verbatim.
- ✅ Sync lookup-only resolver (`lambda ref: prefetched[ref]`) keeps `tier3_assemble.py` pure/synchronous — pre-fetch happens once in async land, then the assembly stays sync. Clean separation.

**Test coverage:**
- Happy path: one path-4 ref + one override resolves both (the common Phase 2 case authoring shape).
- De-dup pinned via `spy.await_count == 1` — the duplicate-ref-share-one-dispatch contract.
- Missing `cases_root` → structured failure with `cases_root` mentioned in message (regression of PR-A.3.1's deferred-stub behavior).
- Missing Tier-2 case → structured failure with "not found" or "lookup failed" — same hard-error contract as `snapshot: frozen` deletion.
- Adversarial short-circuit pinned via `spy.await_count == 0`. Critical to keep — without this, adversarial cases burn LLM calls before being rejected.
- Override-only case runs without `cases_root` — most Phase 2 canonical authoring will hit this path (override is the recommended default).

**Updated dispatch test:** `test_use_tier2_case_without_resolver_yields_structured_failure` correctly renamed to `test_use_tier2_case_without_cases_root_yields_structured_failure` to reflect the new failure mode (was about a stubbed `NotImplementedError`, now about missing `cases_root`).

**RFC Step 2 status after this lands:** PR-A's harness deliverables complete. Only the Step 2 calibration pass remains — the 5-case Master live run on `claude-sonnet-4-5` and `gemini-3-pro-non-preview` against `calibration.json` to validate the $0.12/case ceiling. That's the gate to Step 3a (canonical case authoring with apr13-replay).

**CI green** (mock pass, cached skipped per path filter — correct, no production-prompt changes).

PR #1787, opened 2026-04-25
refactor(compliance-prompt): extract build_compliance_prompt helper — production-side prep for Tier-3 Compliance live mode (PR-A.3.4.f)

rozavala:
Production-side refactor that unblocks Tier-3 Compliance live mode in a follow-up eval-harness PR. RFC: \`docs/rfcs/tier3_evals.md\` Step 2 (PR-A).

Same shape as **PR-A.3.4.c** (#1784, Master) and **PR-A.3.4.e** (#1786, debate). Independent of both — touches a different production code path (\`compliance.py\`).

### \`trading_bot/compliance_prompt.py\` (new)

\`build_compliance_prompt(...)\` renders the Compliance Officer audit prompt. String inputs only — reports assembly, the AGENT AVAILABILITY note, the optional debate-section block, and the decision fields all assembled at the call site per RFC §Extraction boundary.

\`\`\`python
def build_compliance_prompt(
    *,
    reports_text: str,
    market_context: str,
    availability_note: str,
    debate_section: str,
    decision_direction: str,
    decision_reasoning: str,
) -> str:
\`\`\`

Body byte-equivalent to \`compliance.py:851-883\` pre-extraction. The hardcoded persona \"You are the Compliance Officer...\" stays inside the prompt body — production never loaded a persona for this role from config (\`master_persona\` was in \`audit_decision\`'s signature but unused).

### \`trading_bot/compliance.py\` (modified)

\`ComplianceGuardian.audit_decision\` replaces a 33-line inline f-string with a single \`build_compliance_prompt(...)\` call. All input assembly (reports_text, availability_note, debate_section) stays where it was.

## Test plan

- [x] \`pytest tests/test_compliance_prompt_byte_equivalence.py\` — **4/4 pass** in 2.8s. Covers:
  - Helper output matches an independent inline rendering character-for-character
  - Empty optional blocks (availability_note + debate_section) produce a clean prompt with no whitespace artifacts (catches \"Source 3\" header leaks on cycles without debate)
  - Decision direction + reasoning interpolated verbatim into the DECISION section
  - \`audit_decision\` source imports \`build_compliance_prompt\` (catches reverted-extraction drift)
- [x] **58/58 adjacent regression** across compliance_prompt / master_prompt / test_compliance / test_compliance_risk / debate_validation / quote_authenticity_helper. No behavior change.

## What this PR does NOT change

- Compliance prompt **content** — byte-equivalent, locked by tests
- Production call sites — \`audit_decision\` keeps the same signature
[...]

PR #1798, opened 2026-04-25
feat(ci): Tier-3 CI workflow skeleton — Step 7a of RFC tier3_evals.md

rozavala:
## Summary

Step 7a of RFC `docs/rfcs/tier3_evals.md` — workflow scaffolding for Tier-3 ahead of real gating (Step 7b). Code-only, no spend, parallel with Step 2 calibration per RFC.

- **New `tier3-adversarial` job in `eval-gate.yml`** — runs Tier-3 cases in mock mode with `continue-on-error: true`. A Tier-3 failure surfaces in a PR comment (banner reflects step status) but does NOT fail the build. Step 7b flips `continue-on-error` off + adds cost alerting once Step 3a baselines exist.
- **New `--tier3-only` runner flag** — scopes the job's dispatch independently of Tier-2. Lets Step 7b change the job's exit policy without affecting the Tier-2 fast gate.
- **Path filters + cache key bumped to `eval-cache-v3-`** in both `eval-gate.yml` and `eval-drift-weekly.yml` to include the four Tier-3 prompt builders (`master_prompt.py`, `debate_prompt.py`, `compliance_prompt.py`, `devils_advocate_prompt.py`). Without these, a prompt-builder-only PR could skip the eval gate silently.

## Why dry-run today

There are no adversarial Tier-3 cases yet (those land in Step 6) and no Tier-3 baselines yet (those land in Step 3a). The workflow exists today to:
- Surface YAML / permissions / comment-posting bugs before they block real gating.
- Give Tier-3 case authors an early signal that their cases are accepted by the harness.

When Step 3a baselines exist, Step 7b is a small flip: `continue-on-error: false`, switch to cached mode, add cost alerting.

## What's left in the eval thread

- **Live Step 2 calibration** (operator decision, ~$5). Unblocks Step 3 case authoring.
- **Step 3a — Master canonical case set** (~25 cases, blocked by calibration).
- **Step 7b** (after Step 3a): flip the new job to required + cached-mode + cost alerting.

## Test plan

- [x] `python -m evals.runner --tier3-only --mode mock` runs 10 Tier-3 cases locally, all pass.
- [x] New `TestTier3OnlyFlag` test confirms the flag skips Tier-2 load and that the canonical Tier-3 metrics section still renders.
- [x] `pytest tests/ -k 'tier3 or harness or canary or aggregation'` — 232 passed, no regressions.
- [x] `python3 -c 'import yaml; yaml.safe_load(open("..."))` — both workflow files parse cleanly.
- [ ] Workflow's first run on this PR will validate YAML + permissions on real GitHub Actions infrastructure.


rozavala:
**Approved — clean Step 7a CI scaffolding, addresses the path-filter gap I flagged in PR #1787.**

**Path filter expansion is the right fix.** I flagged in #1787 that `eval-gate.yml` watched `agents.py` and `prompts/**` but not `compliance.py` or the new `*_prompt.py` files — so #1787 ran zero CI checks. This PR closes that gap by adding all four prompt builders to the path filter:
- `trading_bot/master_prompt.py`
- `trading_bot/debate_prompt.py`
- `trading_bot/compliance_prompt.py`
- `trading_bot/devils_advocate_prompt.py`

Matches the user's "cheap per-PR gates + scheduled full runs" preference.

**Cache key bump v2 → v3** is necessary and correctly scoped. The hash inputs gain the four prompt builders, so v2 caches from before the Tier-3 PRs cannot be safely reused — bumping the version invalidates them cleanly. Both `eval-gate.yml` and `eval-drift-weekly.yml` updated with identical key+hash inputs (consistency between the two workflows is what makes "free rides on recent warmups" actually work).

**`tier3-adversarial` job** correctly implements RFC §Step 7a:
- `continue-on-error: true` at the step level — failure surfaces in PR comment but doesn't fail the build.
- `needs: fast` — runs after the fast mock-mode gate, parallel-safe.
- Banner reflects step outcome ("non-blocking under Step 7a (RFC dry-run)") so a future engineer reading a PR's CI output understands why a red Tier-3 result didn't block.
- 15-minute timeout — plenty for mock-mode dispatch.
- Artifacts uploaded with 14-day retention.

**`--tier3-only` flag** keeps the Tier-3 dispatch independently scoped from Tier-2. The PR's rationale is correct: "Step 7b will flip the job's exit policy from non-blocking to blocking; keeping Tier-3 isolated means flipping that policy can't accidentally break the Tier-2 contract." Test pins both the success path (exit 0, summary renders) and the lint output ("0 Tier-2 case(s) clean").

**The new check is already green:** `Eval gate (Tier-3 — dry-run, non-blocking)` SUCCESS in this PR's CI. Workflow YAML/permissions/comment-posting all validated on real GHA infrastructure as the PR author noted in the test plan.

**Step 7b is now a small flip:** `continue-on-error: false`, switch to cached mode, add cost alerting once Step 3a baselines exist. That clean separation is the value of landing the skeleton early.

PR #1799, opened 2026-04-25
feat(schema): v11 council_history.csv adds optimized_prompts column + migration

rozavala:
## Summary

Phase A enabler #1 (task #58). Adds the load-bearing forensic field for tracking which prompt source was actually used per decision — required for the no-regression check on PR #1797's eval-trained-prompt deploy to KC.

## What ships

**Schema (v11):**
- New column `optimized_prompts` at the end of `COUNCIL_HISTORY_FIELDNAMES` — JSON dict mapping agent name to `{"source": str, "demo_count": int}`. Empty `{}` when no agent used an optimized prompt.
- Backfill default: `"{}"` (operationally indistinguishable from pre-v11 unknown).
- Format follows `vote_breakdown` / `debate_models` precedent (JSON-in-CSV; dashboard parses via `json.loads`).

**Wiring:**
- `trading_bot/agents.py::research_topic` now returns `prompt_source` + `demo_count` in its result dict.
- Reflexion path preserves the initial pass's source (the persona that shaped the analysis is what's load-bearing for attribution).
- Defensive guards (non-dict wrapper, error paths) get `'legacy' / 0` defaults so the field is always present.
- `orchestrator.py` collects per-agent sources from `cached_meta`, builds the JSON map, includes only agents with a non-default source (keeps the JSON small and truthiness check cheap).

**Migration:**
- New `scripts/migrations/backfill_optimized_prompts.py` follows the `backfill_trend_alignment.py` pattern.
- **Why a dedicated migration:** the generic `migrate_council_schema.py` was already marked applied in earlier deploys (`data/.migrations_applied`). Without a new dedicated migration, the column wouldn't be added on the next deploy and `DictWriter` would silently drop the value when orchestrator writes new rows.
- Idempotent — safe to run multiple times. Auto-discovered by `scripts/run_migrations.py` which `deploy.sh` invokes at Step 6.
- Smoke-tested on a fixture: adds the column with default `"{}"`, second invocation correctly no-ops.

**Tests (3 new):**
- Column + default are registered in canonical schema
- Default round-trips through `json.loads` as `{}`
- Append-only invariant — column is at the end of `COUNCIL_HISTORY_FIELDNAMES`

## Why now

Without this column, we can't reliably filter post-deploy decisions by which prompt source was used. Date-as-proxy gets noisy if anything else changes mid-window. PR #1797 deployed eval-trained agronomist + geopolitical to KC; the no-regression check (Brier score with-vs-without optimized prompt) needs this filter to work.

## What's next (task list)

After this lands:
- task #59 — Document monitoring plan + pre-registered expansion criteria
- task #60 — Add `prompt_source` filter/split to `pages/7_Brier_Analysis.py` (depends on this PR)
- (parallel track) task #61 — Design append-vs-replace instruction semantics for MIPROv2

## Test plan
- [x] `pytest tests/test_debate_schema.py tests/test_agents.py tests/test_orchestrator.py tests/test_specialist_prompt.py -q` → 28 passed
- [x] Migration smoke test: created fixture CSV, ran migration, verified column added with `"{}"` default; second run correctly no-ops
- [ ] After deploy: confirm `data/KC/council_history.csv` has the new `optimized_prompts` column (`head -1`); confirm new rows from KC cycles populate it with `{"agronomist": {...}, "geopolitical": {...}}` for cycles where the loader fired the eval-fallback


rozavala:
**Approved — load-bearing forensic enabler for the no-regression check on PR #1797's deploy. Append-only schema, defensive defaults, idempotent migration.**

**Schema change is correctly append-only.** New `optimized_prompts` column lands at the END of `COUNCIL_HISTORY_FIELDNAMES` (test pins this). Per CLAUDE.md, CSV schema continuity matters for Brier scoring continuity — appending at the end preserves all existing column positions. `test_v11_field_position_is_at_end` is the right guardrail.

**Default value is the right null:** `"{}"` (JSON empty dict literal). Round-trips through `json.loads` to `{}`. Operationally indistinguishable from "no optimized prompt was used" vs "pre-v11 row where we didn't track this" — that's correct because both are non-events from the no-regression check's perspective.

**Format precedent followed:** JSON-in-CSV like `vote_breakdown` and `debate_models`. Dashboard parses via `json.loads`. Consistent with the existing pattern.

**Wiring verification:**
- `_prompt_source` and `_demo_count` already exist as locals in `research_topic` (line 1173 of agents.py at HEAD: `full_prompt, _prompt_source, _demo_count = build_specialist_prompt(...)`). This PR just adds them to the result dict so the orchestrator can persist them — no new computation, just exposure.
- Defensive guards: error path and reflexion-init-failed path both default to `'legacy' / 0` so the field is always present. Won't crash the orchestrator on a malformed agent return.
- `research_topic_with_reflexion` correctly preserves the **initial pass's** prompt_source through reflexion. The author's reasoning is sound: the reflexion uses a different (critique) prompt, but the analysis it critiques came from the persona we want to track. Crediting the initial source is the right attribution for the no-regression check.

**Orchestrator filter is well-scoped.** Persists only agents whose source is something other than `'legacy'` / `'commodity_profile'` (the two default sources). Keeps the JSON dict small and the truthy check cheap — most decisions won't have any optimized-prompt agents, so empty `{}` is the common case. One subtle observation: the filter is a hardcoded denylist of default sources. If a future change introduces another "default" source (e.g. `'persona_only'`), it'd be persisted incorrectly. Worth a follow-up to centralize the "is this a default or optimized source?" check, but not blocking.

**Migration script is sound:**
- Idempotent (rows with non-empty values left unchanged).
- Atomic write via `tmp` file + `os.replace`.
- Backup created (`.bak`).
- Dry-run support.
- Per-version pattern matches `backfill_trend_alignment.py` (v10) — the right precedent.
- Auto-discovered by `scripts/run_migrations.py` which `deploy.sh` invokes.
- The PR body explains why a dedicated migration is needed: the generic `migrate_council_schema.py` was already marked applied in earlier deploys, so without a new entry the column wouldn't be added on the next deploy and `DictWriter` would silently drop the value when orchestrator writes new rows. That's exactly the failure mode dedicated per-version migrations prevent.

**Tests:**
- `test_v11_optimized_prompts_column_in_schema`
- `test_v11_backfill_default_is_empty_json_object`
- `test_v11_field_position_is_at_end`
- 28 passed across `test_debate_schema.py tests/test_agents.py tests/test_orchestrator.py tests/test_specialist_prompt.py`

CI green. Migration smoke-tested per PR body.

**Operational checklist** in the PR body is right: after deploy, confirm `data/KC/council_history.csv` has the new column (`head -1`), and confirm KC cycle rows populate it for cycles where the loader fired the eval-fallback. If the column is missing or always empty, the migration didn't run or the loader instrumentation broke — investigate before relying on the column for the no-regression analysis.

PR #1800, opened 2026-04-25
rfc: MIPROv2 instruction optimization — append vs replace decision

rozavala:
## Summary

Phase B prerequisite (task #61). Decides one load-bearing question before any code: **when MIPROv2 generates new instruction text, should the production loader REPLACE the resolved persona with it, or APPEND it as a supplement?**

**Decision: APPEND semantics.**

Three reasons in order:

1. **Preserves load-bearing per-commodity work.** The `commodity_profile` system encodes per-ticker engineering. A single agent-scoped replacement instruction loses that.
2. **Smaller blast radius.** Replace ships an entire persona swap to four tickers at once with one PR. Append ships an additive perturbation. If MIPROv2 generates something subtly wrong, the persona still sets the floor.
3. **Failure-mode targeting matches the diagnosis.** Macro's over-direction-priming and volatility's expensive-premium under-recognition are SPECIFIC failure modes the persona has. A supplement targeting those is conceptually right; a replacement that has to also handle every other case is a bigger, harder generation problem.

## Why now

PR #1795's 4-agent eval established that demo selection alone has structural limits — macro regressed −18pp even with the stratified selector, volatility churned at 0pp net. The diagnosis was clear: those agents need *instruction-level* changes, not exemplar-level. MIPROv2 is the existing DSPy machinery for instruction optimization; the only blocking question is loader semantics.

This RFC unblocks tasks #62 (loader change), #63 (`--mipro` trainer mode), #64 (research spike on macro).

## Scope

**In scope:**
- Decision: append vs replace.
- Implementation sketch for loader + trainer.
- Cost estimate (~$5/agent per MIPROv2 run).
- Out-of-scope items (per-commodity runs, replacing the demo selector, auto-rollback).
- Decision points to revisit after the spike.

**NOT in scope** (RFC explicitly notes):
- Per-commodity MIPROv2 runs — too few cases per commodity per agent for signal.
- Replacing the demo selector — both can coexist in one `prompt.json`.
- Re-optimizing already-shipping agents (agronomist, geopolitical) — spike targets ship-blocked agents first.
- Auto-rollback on production data — separate, larger discussion.

## Risks acknowledged

- Spike may not show macro lift. Acceptable downside: ~$5 cost, learn that demo selection is the documented limit.
- Meta-LM (gpt-5 or claude-opus) cost variance — locks us to a vendor's pricing.
- Generated instructions may hallucinate constraints — mitigated by candidate-scoring step + manual review of chosen supplement before ship.
- Train/val leakage discipline still applies — same as the demo selector.

## Test plan
- [x] RFC reviewed against existing trainer / loader code paths
- [ ] Approval from reviewer (you) before any code lands for #62 / #63
- [ ] After #62 + #63 land: run #64 spike on macro
- [ ] After spike: re-evaluate per "Decision points to revisit" — proceed to volatility or park


rozavala:
**Approved — well-reasoned RFC. Decision (append semantics) is correct and the rationale order matches the right priority.**

**The three reasons hold up:**

1. **Preserves load-bearing per-commodity work.** The `commodity_profile` system is real engineering. Replacing it with a single agent-scoped instruction averaged across 4 commodities is a regression in expected value — quantitatively, you're trading a system that knows KC ≠ NG ≠ SB ≠ ZW for one that has to learn that distinction from a small number of cases per commodity. Append keeps tier-2's work intact.

2. **Smaller blast radius.** Replace ships an entire persona swap to four tickers with one PR. Append ships an additive perturbation. The asymmetry matters: a subtly-wrong replacement persona breaks the floor; a subtly-wrong supplement is noise on top of a still-valid floor. The toggle (`use_optimized_prompts.{ticker}`) is the kill switch in both cases, but the recovery cost differs.

3. **Failure-mode targeting matches the diagnosis.** Macro's "FX-concentrated demos break adversarial NEUTRAL handling" and volatility's "expensive-premium → BEARISH pattern recognition" are SPECIFIC failure modes. A supplement targeting those is conceptually right; a replacement that has to also handle every other case is a bigger, harder generation problem.

**Out-of-scope items are correctly scoped:**
- Per-commodity MIPROv2 runs — case counts too low for signal. Right call.
- Replacing the demo selector — coexistence with demos is the right model.
- Re-optimizing already-shipped agents — concentrate spike on ship-blocked agents first; revisit if both lift.
- Auto-rollback — separate, larger discussion. Toggle remains kill switch.

**Cost estimate is honest:** $1.50-$4 per agent per MIPROv2 run, ~$10 spike total (macro + volatility). Acceptable downside if it doesn't lift macro.

**Decision points to revisit** are concrete and falsifiable. The cost-vs-lift threshold ($5/agent for +1pp = poor) is exactly the right honest framing — at that point persona-text engineering by hand competes on cost, and the trainer machinery isn't earning its keep.

**One thing worth pinning at implementation time (not blocking RFC approval):**

The Implementation sketch shows the loader's precedence as: `instruction` (replace) wins over `instruction_supplement` (append) if both are set. The RFC says "this is enforced in the trainer, not the loader." Reasonable, but the loader's precedence path will silently use the wrong field if a malformed `prompt.json` slips through (e.g. trainer bug, manual edit). Suggest task #62 add a lint/validation step that hard-errors when a `prompt.json` has both fields set, surfacing the bug at load time rather than letting the loader's precedence rule silently choose. The `prompt_source` value will distinguish but only after the fact.

**Stack alignment:**
- Builds on PR #1799's v11 forensic tracking — the `optimized_prompts` source string can take a new value (`dspy_optimized_supplement`) per the RFC's plan.
- Tasks #62 (loader), #63 (trainer `--mipro` mode), #64 (research spike on macro) unblock once this lands.

PR #1796, opened 2026-04-25
feat(evals): Step 2 calibration scaffolding — PR-A.3.4.l of RFC Step 2

rozavala:
## Summary

Scaffolding for the Step 2 calibration pass — the RFC's explicit Step 2 exit criterion gating Step 3 case authoring. Builds the runner + 4 hand-built Master stress cases so the canonical 5-case × 2-model calibration can run on a single command.

**Does not execute live LLM calls in this PR.** Live calibration costs ~$5; the operator decides when to spend.

- `evals/runner/calibrate.py`: CLI runner. Resolves specialist_inputs via the harness machinery (mock-mode resolution; calibration cases use override exclusively), assembles prompts via `build_master_prompt`, then routes per (case, model) with `HeterogeneousRouter` assignment pinning. Writes `evals/runs/<ts>_<commit>_calibration/calibration.json` with per-call telemetry, per-model summary stats, and a gate verdict (`PASS` / `FAIL` / `DRY_RUN`) against the RFC's $0.12/case ceiling.
- `evals/cases/master/calibration/`: 4 hand-built Master stress cases — `consensus_bearish` (KC), `consensus_neutral` (KC), `override_consensus` (SB), `split` (ZW). All 8 Tier-2 specialists provide full-length override text to land prompt sizes in the production range.
- `config/api_costs.json`: rate added for `claude-sonnet-4-6` (matching `model_registry.anthropic.pro`); `gemini-3.1-pro-preview` already in the table. The RFC named `claude-sonnet-4-5` and `gemini-3-pro-non-preview` but production has rolled forward — calibrating against names production actually routes through is more representative.
- `tests/test_calibrate.py`: 34 tests covering cost calculation, percentile math, summary aggregation, gate verdict, dry-run end-to-end, JSON round-trip, and structural validation of the 4 calibration cases.

## To run the calibration when ready (operator decision, ~$5 spend)

```bash
python -m evals.runner.calibrate \
    --cases evals/cases/master/calibration evals/cases/master/canary.yaml \
    --models claude-sonnet-4-6 gemini-3.1-pro-preview
```

Exit code 1 if any model's median exceeds $0.12/case. Follow-up commit lands the resulting `calibration.json` to satisfy the RFC's exit criterion.

## What's left after this

- **Live calibration run + commit `calibration.json`** (your call). If pass, Step 3 unblocks.
- **Step 7a CI skeleton** (parallel work, code-only, no spend).

## Test plan

[...]

PR #1801, opened 2026-04-25
fix(calibrate): unblock live-mode config load before Step 2 calibration run

rozavala:
## Summary

Two bugs that would have crashed the live Step 2 calibration before any LLM call:

1. `calibrate.py` called `load_config(args.config)` but `config_loader.load_config()` takes no arguments.
2. Even with the signature fixed, `config_loader.load_config()` validates the full production stack (IB connection, strategy, risk_management, notifications credentials, data-provider keys). Validation raises on a worktree without every production secret in `.env`.

Fix: new `_load_config_for_calibration` helper reads `config.json` directly + `load_dotenv()` for API keys, skipping production validations the calibration runner doesn't need (no orders, no notifications, no IB). Also drops the unused `--config` CLI flag.

Caught while sanity-checking the runbook before live spend. No live LLM call has happened yet.

## Why a separate helper instead of patching load_config

Production's `load_config` validation is correct — it should fail loudly when the trading bot is missing notification credentials. Loosening it would weaken production. The calibration runner is a different operating context (no orders, no notifications) so it gets its own loader.

## Test plan

- [x] `_load_config_for_calibration()` returns a config with the expected `model_registry` on this worktree (which doesn't have full prod `.env`).
- [x] Dry-run unchanged: `python -m evals.runner.calibrate --cases evals/cases/master/canary.yaml --models claude-sonnet-4-6 --dry-run` → DRY_RUN verdict, $0.0187/case as before.
- [x] `pytest tests/test_calibrate.py` — 34 passed.

After merge, the live calibration command is::

    python -m evals.runner.calibrate \
        --cases evals/cases/master/calibration evals/cases/master/canary.yaml \
        --models claude-sonnet-4-6 gemini-3.1-pro-preview

Run from a worktree with `.env` containing `GEMINI_API_KEY` and `ANTHROPIC_API_KEY`. ~$5 spend.


rozavala:
**Approved — clean unblocker for the live Step 2 calibration run. Caught two real bugs before any spend.**

**Both bugs are real and the fix is correctly scoped:**

1. `load_config(args.config)` was called with an argument that `config_loader.load_config()` doesn't accept — would have crashed at the first live-mode invocation regardless of secrets state.

2. Even with the signature fix, `config_loader.load_config()` validates the full production stack (IB connection, strategy quantities, risk_management, notifications credentials, data-provider keys) — and would raise on a worktree that lacks every prod secret. A calibration host doesn't need IB or Pushover credentials; requiring them just to measure LLM cost would be friction without safety.

**Architectural choice is correct.** "Production's `load_config` validation is correct — it should fail loudly when the trading bot is missing notification credentials. Loosening it would weaken production. The calibration runner is a different operating context (no orders, no notifications) so it gets its own loader." Right call. Two operating contexts, two loaders, no weakening of production's strict validation.

**Implementation is minimal and defensive:**
- Reads `config.json` directly via `Path.read_text` + `json.loads`. No magic.
- `load_dotenv(".env")` so `GEMINI_API_KEY` / `ANTHROPIC_API_KEY` land in `os.environ` where the LLM clients pick them up.
- Tolerates missing `python-dotenv` (graceful import fallback) — keeps the test path clean.
- Drops the unused `--config` flag rather than wiring a path argument that would never be exercised.

**Error message is operator-facing:** "Live calibration requires `config.json` (project root) + API keys in `.env`. Pass `--dry-run` to validate the runner without LLM calls." Tells the operator exactly what to check before retrying.

**Three minor non-blocking observations:**
- Path `Path(".env")` relies on cwd being the project root. Not a regression (previous code had the same dependency), but worth noting if the runbook ever mentions running from a different cwd.
- The helper doesn't validate that the API keys are present after `load_dotenv`. The author's docstring acknowledges this: "The router will surface a clearer error if API keys are actually missing at route time." Reasonable — adding an explicit check here would duplicate logic the router already has.
- Tested on the dry-run path (per PR body: $0.0187/case unchanged). The actual live-load path will only be exercised once when the operator runs the live calibration, so any latent issue surfaces there. Acceptable risk for a calibration runner that the operator runs interactively.

**Caught before any LLM spend** — exactly the kind of pre-spend sanity check that justifies "scaffolding lands separately from execution" pattern from PR #1796.

**CI green** including the new Tier-3 dry-run check from PR #1798.

PR #1802, opened 2026-04-25
feat(prompts): instruction_supplement (append) loader semantics for MIPROv2

rozavala:
## Summary

Implements task #62, follows the design decided in [`docs/rfcs/miprov2_instruction_optimization.md`](https://github.com/rozavala/real_options/blob/main/docs/rfcs/miprov2_instruction_optimization.md) (PR #1800). Production loader supports a new optional field `instruction_supplement` that's appended under the resolved persona instead of replacing it.

## Behavior

```
[commodity-aware persona]            ← tier 2 resolves
\n\n
OPTIMIZED FRAMING:                   ← instruction_supplement (this PR)
[supplement text]
\n\n
EXAMPLES OF GOOD ANALYSIS:           ← demos (existing — PR #1791)
[demo block]
```

**Precedence** (when both `instruction` and `instruction_supplement` are set in the same payload — should not happen, but loader handles defensively):
- `instruction` (REPLACE) wins over `instruction_supplement` (APPEND)
- Demos compose with either

## New `prompt_source` values for v11 forensic column

| value | meaning |
|---|---|
| `dspy_optimized` | `instruction` (replace) was used |
| `dspy_optimized_supplement` | `instruction_supplement` (append) was used. May also include demos; `demo_count` tracks them separately. |
| `dspy_optimized_demos_only` | Only `demos` — no instruction or supplement |

## Why this and not just replace semantics

Per the RFC's three reasons:
1. **Preserves load-bearing per-commodity engineering** — commodity_profile system stays intact
2. **Smaller blast radius** — additive perturbation, not wholesale replacement
3. **Failure-mode targeting matches the diagnosis** — macro/volatility have specific failure modes that supplements target naturally

## Scope

- `trading_bot/specialist_prompt.py::_resolve_persona_prompt` — adds the new branch (~15 lines)
- Docstring update on `build_specialist_prompt`'s `optimized_prompt` param
- 5 new tests in `tests/test_specialist_prompt.py` covering all combinations
- `evals/_eval_optimized/README.md` updated with the schema field table

## What's next (task list)

After this lands:
- task #63 — Implement `--mipro` trainer mode (emits `instruction_supplement` field)
- task #64 — Run MIPROv2 spike on macro, decide based on RFC §"Decision points to revisit"

## Test plan
- [x] `pytest tests/test_specialist_prompt.py tests/test_eval_dspy_trainer.py tests/test_eval_harness.py tests/test_debate_schema.py tests/test_agents.py -q` → 103 passed
- [x] 5 new tests for supplement semantics (legacy persona + supplement, commodity-aware + supplement, supplement+demos compose, replace-wins-over-supplement, empty-supplement-falsy)
- [ ] After merge: implement task #63 so trainer can actually emit `instruction_supplement`
- [ ] After #63: run task #64 spike and confirm the new loader path fires correctly with a real MIPROv2-generated supplement


rozavala:
**Approved — clean append-semantics implementation matching the RFC's sketch.**

**Implementation matches RFC #1800's design:** precedence is `instruction` (replace) > `instruction_supplement` (append), demos compose with either, and `prompt_source` correctly distinguishes the three states (`dspy_optimized` / `dspy_optimized_supplement` / `dspy_optimized_demos_only`). The 15-line loader change is structurally correct.

**Test coverage is exactly what I'd want:**
1. **`test_dspy_supplement_appends_to_legacy_persona`** — supplement under legacy persona, header present, persona ordering pinned (`prompt.index("LEGACY")` < `prompt.index("OPTIMIZED FRAMING:")`).
2. **`test_dspy_supplement_appends_to_commodity_aware_persona`** — the load-bearing case the RFC's append decision protects. Pins that commodity-aware persona is preserved when supplement is set, and explicitly checks legacy doesn't leak through (`assert "LEGACY MACRO PROMPT" not in prompt`).
3. **`test_dspy_supplement_with_demos_appends_both`** — supplement + demos compose, source reflects supplement, demo ordering pinned.
4. **`test_dspy_instruction_replace_wins_over_supplement`** — precedence on conflict. Both `OPTIMIZED FRAMING:` header and `LEGACY MACRO PROMPT` are explicitly absent when replace fires.
5. **`test_dspy_empty_supplement_falls_back_to_resolved_persona`** — defensive against trainer accidentally writing an empty string instead of omitting the field. Source stays `legacy`.

**Two notes for the follow-up trainer PR (task #63):**

1. **My #1800 review suggested hard-erroring at load time when both `instruction` and `instruction_supplement` are set.** The PR chose defensive precedence at the loader instead, with the trainer expected to never emit both (README: "The trainer should never emit both; loader enforces the precedence defensively"). That's a reasonable choice — graceful degradation over fail-loud — but it means a trainer bug producing both fields would be silent. Suggest task #63 adds an explicit validation in the trainer write path that hard-errors when both fields are non-empty, so the failure surfaces at trainer time rather than at production load time.

2. **`prompt_source` value space differs from RFC by design.** RFC §"v11 forensic tracking" listed `dspy_optimized_supplement_with_demos` as a possible source value. This PR consolidates: source is `dspy_optimized_supplement` regardless of whether demos are also present, and `demo_count` captures the demo dimension separately. Test docstring justifies it: "source still reflects the primary override (supplement). demo_count tracks demos independently." This is the right simplification — avoids cartesian explosion of source values, the analyst can split on `(source, demo_count > 0)`. Worth updating the RFC's v11 forensic tracking section in a follow-up to drop the `_with_demos` suffix value, or explicitly noting in `evals/_eval_optimized/README.md` that the source value is intentionally orthogonal to demo presence.

**README updates are good:** schema field table with producer attribution (price-outcome trainer → `instruction`, MIPROv2 → `instruction_supplement`, demo selector → `demos`), explicit precedence statement, v11 source value mapping. Future engineers reading the README will understand the three-tier composition without having to read the loader code.

**CI green** including the new Tier-3 dry-run check.

**Stack alignment:**
- Builds on PR #1800 (RFC).
- Builds on PR #1791 / #1797 / #1799 (demo selector + production deploy + v11 forensic column).
- Unblocks task #63 (`--mipro` trainer mode) and task #64 (research spike on macro).

PR #1804, opened 2026-04-25
feat(evals): MIPROv2 trainer mode (--mipro) — writes instruction_supplement

rozavala:
## Summary

Implements task #63. Adds `--mipro` mode to the eval trainer for instruction optimization (the research path that should help agents demo selection couldn't fix). Follows the design decided in [`docs/rfcs/miprov2_instruction_optimization.md`](https://github.com/rozavala/real_options/blob/main/docs/rfcs/miprov2_instruction_optimization.md) (PR #1800), uses the loader path implemented in PR #1802.

## What it does

`python -m evals.dspy_trainer --agent macro --mipro --meta-model anthropic/claude-opus-4-7 --mipro-auto light`

Runs `dspy.MIPROv2` with:
- **prompt_model** = meta-LM (default `claude-opus-4-7`) — generates candidate instructions
- **task_model** = per-agent eval-tier LM (from `eval_models.json`) — scores candidates
- **metric** = `direction_only_metric` (existing)
- **trainset** = filtered to `--agent`'s train cases

Persists the best instruction to `instruction_supplement` (NOT `instruction`) per the RFC's append semantics. The loader (PR #1802) prepends it under `OPTIMIZED FRAMING:` below the resolved (commodity-aware) persona.

## Cost

Per RFC §Cost: ~\$1.50-\$4 per agent per run with `--mipro-auto light` and default candidate count. Spike budget (task #64): ~\$5 on macro alone. If macro lifts, ~\$5 more on volatility.

## Failure modes (all return audit artifact, no crash)

- `EXIT_TOO_FEW_DEMOS (10)` — not enough train cases
- `EXIT_DISPATCH_ERROR (11)` — `MIPROv2.compile` raised (rate limit, network, quota)
- `EXIT_NO_IMPROVEMENT (13)` — optimizer ran but produced empty instructions (defensive against shipping a useless empty supplement)

Train mode (both `--train` and `--mipro`) does NOT gate ship — `do_not_ship: False` unless hard failure. The real ship decision is `--validate`, which works unchanged.

## What this does NOT do (per RFC)

- Per-commodity MIPROv2 runs — too few cases per commodity for signal
- Combine MIPROv2 supplements with demo selector demos in one artifact — `--mipro` and `--train` overwrite each other's `prompt.json`
- Re-optimize already-shipping agents

## Test plan
- [x] `pytest tests/test_eval_dspy_trainer.py tests/test_specialist_prompt.py tests/test_eval_harness.py tests/test_debate_schema.py -q` → 106 passed
- [x] 8 new tests: argparse mutual exclusion, meta-model spec parsing, too-few-cases path, instruction_supplement write semantics, empty-instruction defensive guard, dispatch-failure capture
- [x] All 8 new tests run offline (mocked `dspy.MIPROv2` + `dspy.Example`) — no LLM cost in CI
- [ ] After merge: task #64 — run `--mipro` on macro, validate, decide based on RFC §"Decision points to revisit"

## What's next (task list)

- task #64 — Run MIPROv2 spike on macro (~\$5 cost). Decision: does instruction-level optimization fix what demo selection couldn't?


rozavala:
**Approved — clean MIPROv2 trainer mode, fully offline-tested. Task #63 from the RFC chain.**

**Implementation hits all the RFC's design points:**
- Writes `instruction_supplement` (NOT `instruction`) — load-bearing append-semantics contract, asserted in `test_mipro_writes_instruction_supplement_not_instruction`.
- Empty `instruction` and `demos` fields explicitly cleared so the artifact is unambiguous about which optimization mode produced it.
- `ship_decision_authority: "validate"` consistent with the train-mode fix from PR #1797's follow-up. `do_not_ship: False / exit 0` unless hard failure; the real ship gate is `--validate`.
- Provenance captures `meta_model`, `mipro_auto`, `mipro_num_candidates` so a future re-train can reproduce the same configuration.

**Mutual exclusion correctly enforced.** `mode = p.add_mutually_exclusive_group(required=True)` makes `--mipro` exclusive with `--train` / `--validate` / `--smoke`. Test pins this:
```python
def test_argparse_mipro_and_train_mutually_exclusive():
    parser = _build_parser()
    with pytest.raises(SystemExit):
        parser.parse_args(["--agent", "macro", "--train", "--mipro"])
```
Defends against the artifact-conflict failure mode from RFC §"Out of scope" (combining demos + supplement is out of scope; mutex makes it impossible at CLI).

**Failure modes have specific exit codes:**
- `EXIT_TOO_FEW_DEMOS (10)` — early bail before any LLM cost. Test pins that the optimizer is **never invoked** (`assert "instruction_supplement" not in result`) so the no-spend property is guaranteed.
- `EXIT_DISPATCH_ERROR (11)` — `MIPROv2.compile` raised. Captures exception type in skip_reason. Audit artifact still written.
- `EXIT_NO_IMPROVEMENT (13)` — defensive against shipping empty supplement (whitespace-only `instructions` from MIPROv2). Whitespace check uses `.strip()` so `"   "` is correctly classified as no improvement.

**Lazy dspy import** is the right call — `_build_dspy_lm` and the dspy.Module construction live inside `mipro_agent`, not at module top-level. `--train` / `--validate` / `--smoke` don't pay the dspy startup cost.

**`temperature=0.0`** for both eval-LM and meta-LM keeps generation deterministic given the seed. Cache layer (if dspy uses one) gets predictable hits across re-runs.

**One subtle design observation worth flagging for the spike (task #64):**

The `_build_mipro_eval_module` constructs a minimal `dspy.Predict` with a `market_context -> response_json` signature. At MIPROv2 optimization time, `signature.instructions` is the PRIMARY instruction for that predict node. At production load time, the same string gets extracted and APPENDED under the resolved persona via PR #1802's `instruction_supplement` path. So:
- **At optimization time:** the instruction is "the whole brief"
- **At deploy time:** the instruction is "extra framing under the existing persona"

These are different semantic contexts. The RFC's hope (correct, IMO) is that MIPROv2 will generate instructions that focus on the *specific failure mode* the agent has (e.g., for macro: "When FX is one of multiple offsetting factors, call NEUTRAL"), and that focus remains useful even when appended after a fully-formed persona. But it's a research bet, not a foregone conclusion. Task #64's spike outcome will tell us whether append-semantics + MIPROv2 actually lifts macro.

If macro's val lift is near-zero post-spike, one decision-point worth raising in the RFC §"Decision points to revisit" follow-up: whether the eval-tier signature should also include the existing persona text (so the meta-LM optimizes a supplement that knows what it's supplementing), rather than the bare-bones `market_context -> response_json` shape this PR uses. That's a v2 refinement, not a blocker for the spike.

**8 new tests, all offline.** Mocks `dspy.MIPROv2` and `dspy.Example` so no network/LLM calls in CI. 106 tests passed in the broader regression. CI green.

**Unblocks task #64** — research spike on macro. Decision criteria documented in RFC §"Decision points to revisit": did MIPROv2 lift macro out of regression? what's the supplement quality? cost vs lift ratio.

PR #1809, opened 2026-04-25
fix(evals): MIPROv2 auto + num_candidates mutex (surfaced by macro spike)

rozavala:
## Summary

Surfaced by task #64's macro spike. Two small fixes to the MIPROv2 trainer mode shipped in PR #1804.

## Issue 1: ValueError from auto + num_candidates conflict

DSPy 3.1.3 raises:

> "If auto is not None, num_candidates and num_trials cannot be set, since they would be overridden by the auto settings."

The trainer always passed both because they were independent CLI flags. The first spike attempt crashed with `EXIT_DISPATCH_ERROR` before any LLM cost was incurred (no LLM calls happened — failed at optimizer init).

**Fix:** only pass `num_candidates` when `auto is None` or `"off"`. Added `"off"` as a valid `--mipro-auto` choice so users can manually control candidate count if needed.

## Issue 2: deprecation warning on requires_permission_to_run

DSPy 3.1.3 deprecated this kwarg (still honored, scheduled for removal). Wrapped the `compile()` call in `try/except TypeError` so a future API change doesn't break the headless behavior.

## End-to-end confirmation

Post-fix, `python -m evals.dspy_trainer --agent macro --mipro --mipro-auto light` runs successfully through bootstrap + trial loop and produces a result JSON. The actual spike outcome (macro regressed -9.1pp on val, `EXIT_REGRESSION 14`) is a separate decision per RFC #1800 §"Decision points to revisit"; that's not a code issue with this fix.

## Test plan
- [x] `pytest tests/test_eval_dspy_trainer.py -q` → 51 passed
- [x] Live `--mipro` run on macro reaches bootstrap + trial loop without crashing (best score 93.75% on train minibatch)
- [x] Validate ran successfully against the produced artifact (regression result was the spike's finding, separately reported)


rozavala:
**Approved — clean DSPy 3.1.3 compatibility fix surfaced by the macro spike.**

**Two real bugs caught by running the spike:**

1. **`auto` vs `num_candidates` mutex.** DSPy 3.1.3 enforces that `MIPROv2(auto=..., num_candidates=...)` raises ValueError when both are set. The trainer always passed both because they were independent CLI flags. First spike attempt crashed at optimizer init with `EXIT_DISPATCH_ERROR` before any LLM cost. Fix: only pass `num_candidates` when `auto is None or "off"`. New `"off"` choice on `--mipro-auto` lets users opt out of auto for manual control.

2. **`requires_permission_to_run` deprecation.** Wrapped `compile()` in `try/except TypeError` so a future DSPy version that drops the kwarg keeps headless behavior. Defensive but cheap.

**Spike outcome separately reported:** macro regressed -9.1pp on val with EXIT_REGRESSION 14. That's a real datapoint for RFC #1800 §"Decision points to revisit" — the small-N caveat from PR #1812's CIs is going to be load-bearing in deciding whether the regression is meaningful or noise. Not a code issue with this fix.

**Test coverage:** 51 trainer tests pass. Live `--mipro` run on macro reaches bootstrap + trial loop (best score 93.75% on train minibatch), validate ran successfully against the produced artifact. The actual code path is exercised end-to-end against the real DSPy + LLM stack.

**CI green** including the Tier-3 dry-run check.

PR #1812, opened 2026-04-25
feat(evals): Wilson/Newcombe CIs + posterior P(lift>0) in --validate output

rozavala:
## Summary

Task #67. Surfaced by the small-N discussion after the macro spike (task #64): at N=10-15 per agent, single case flips dominate the "lift" point estimate. Reporting only point estimates hides that agronomist's "+10pp lift" and the macro spike's "−9.1pp regression" are both within the noise band of their sample.

## End-to-end demonstration on agronomist (currently shipped)

```
n_val_cases:  10
baseline_val: 0.700  CI=[0.397, 0.892]
trained_val:  0.800  CI=[0.490, 0.943]
lift_pp:      +10.0pp  CI=[-26.5pp, +43.5pp]
P(lift > 0):  0.68
power_warning: n=10 is below the statistical-power threshold (~30) for detecting ~10pp effects. Single case flips dominate the signal at this sample size...
```

The point estimate +10pp is consistent with anywhere from a 26pp regression to a 43pp improvement at 95% confidence. P(lift > 0) is 0.68 — barely better than coin flip. **This is the load-bearing motivation: existing point-estimate-only reporting overstates what the data can support.**

## What ships

Closed-form statistical helpers (no scipy dep):

| helper | purpose |
|---|---|
| `_wilson_ci(s, n)` | Wilson score 95% CI on a binomial proportion. Reliable at small n where Wald breaks. |
| `_newcombe_diff_ci(s1, n1, s2, n2)` | Newcombe Method 10 CI for the difference of two proportions. |
| `_lift_prob_positive(...)` | Bayesian posterior P(trained > baseline) with Uniform priors (MC, seed=42, n=20000). |
| `_power_warning(n)` | Warning string when n < 30. |

`validate_agent` output gains five new fields: `baseline_val_acc_95ci`, `trained_val_acc_95ci`, `lift_pp_95ci`, `lift_prob_positive`, `power_warning`.

## Ship gate behavior unchanged

This PR only **surfaces** uncertainty in the output — the gate still fires on point-estimate lift vs threshold. Changing the gate to require `P(lift > 0) > X` is a separate decision worth raising in the next monitoring-plan PR (task #59).

## Test plan
- [x] `pytest tests/test_eval_dspy_trainer.py -q` → 61 passed (10 new + 51 existing with updated stubs)
- [x] Newcombe CI on macro spike data MUST include zero (load-bearing assertion — confirms the "regression" is statistically meaningless at N=11)
- [x] Wilson CI on agronomist baseline (10/11) reflects real uncertainty (lower bound ≈ 0.59)
- [x] Power warning fires at n < 30, silent above
- [x] End-to-end live validate on agronomist confirms all 5 new fields populate correctly

## What's next

After this lands, two paths in flight (per the small-N discussion):
- task #68 — Author sample batch of new macro cases (then #69 to scale)
- task #59 — Update monitoring plan to reference these CIs and propose a P(lift > 0) ship gate


rozavala:
**Approved — high-value statistical honesty fix. The macro spike was the right teacher.**

**The motivation is load-bearing.** The agronomist case the README celebrates (+10pp lift, shipped) and the macro spike's "regression" (-9.1pp, blocked by gate) are **both within the noise band of N=10-15 samples**. Without CIs in the validate output, the operator has no way to see that. The PR's end-to-end demonstration on agronomist makes this concrete:

```
lift_pp:      +10.0pp  CI=[-26.5pp, +43.5pp]
P(lift > 0):  0.68
```

That's a coin-flip-plus-a-bit at 95% confidence, not the unambiguous "+10pp lift" the point estimate suggests. Surfacing this in the validate output is exactly the right primitive — even if the ship gate doesn't change today, the operator and future deploy-decision discussions now have honest uncertainty quantification.

**Statistical implementation is sound:**
- **Wilson score CI** (closed-form, valid at small n where Wald breaks). Clamps to [0,1]. Tests pin extreme cases (0 successes, all-successes, n=0).
- **Newcombe Method 10 CI** for the difference of two proportions. Standard textbook formula. Test pins that CI center is near observed difference.
- **Bayesian posterior P(lift > 0)** via MC with Beta(1,1) priors, n=20000, fixed `seed=SEED=42`. Reproducible. Comment notes SE ~0.003 — accurate enough for decision-support.
- **Power warning** at n < 30. Concrete operator guidance: "treat point estimate as exploratory; prefer the 95% CI bounds and lift_prob_positive for decision-making."
- **No scipy dependency** — closed-form math + stdlib `random.betavariate`. Keeps eval deps minimal.

**Critical anchor test** (`test_wilson_ci_macro_spike_baseline_is_wide_at_n_11`):
> "At N=11, p=10/11=0.91, Wilson lower bound should reflect the genuine uncertainty (we expect ~0.59). Got lo={lo:.3f}"

This test docstring is the load-bearing assertion: the observed 91% baseline is statistically consistent with anywhere from ~59% to ~99% true accuracy. Locks the behavior so a future Wilson refactor that silently narrows the CI fails loudly.

**Ship gate explicitly unchanged.** PR description: "This PR only **surfaces** uncertainty in the output — the gate still fires on point-estimate lift vs threshold." Right separation of concerns. The "should we make P(lift > 0) > X a ship gate?" decision belongs in the monitoring-plan PR (task #59), not bundled here.

**One observation worth noting (non-blocking):** Beta(1,1) is a uniform prior — appropriate for "no prior evidence" but at small n the posterior is heavily prior-influenced. If you wanted to be more conservative in flagging false-positive lifts, an informative prior centered on the per-agent historical baseline would be tighter. Not a v1 concern; flagging for if/when the ship gate pivots to use P(lift > 0).

**Test coverage:** 61 passed (10 new + 51 existing with stubs updated to provide the new `n` / `successes` keys the validate fields require). End-to-end live validate confirms all 5 new fields populate correctly.

**CI green** including the Tier-3 dry-run check.

**Stack alignment:** unblocks task #68 (sample batch of new macro cases — directly motivated by these CIs) and task #59 (monitoring plan).

PR #1814, opened 2026-04-25
feat(eval-cases): macro persona offset/composition tests (sample batch, 6 cases)

rozavala:
## Summary

Task #68. Sample batch for review before scaling to ~50 cases (#69). Targeted authoring for macro's persistent failure mode surfaced by both the demo-selector experiments (PR #1791) and the MIPROv2 spike (task #64): the persona over-anchors on a dominant signal and misses second-order offsets, OR misreads compounding factors as offsetting.

The case `macro.contrast.sb_ethanol_mix_003_a` — which flipped under both experiments — exemplifies the failure pattern. These 6 new cases teach the persona to *read the transmission channel*, not just the headline.

## What's in the batch

**2 adversarials:**
- `adversarial.007_kc_china_credit_demand_collapse` → **NEUTRAL** — PBoC RRR + LPR cuts (textbook bullish for EM commodities) offset by structural Asian arabica demand collapse. Different transmission channels.
- `adversarial.007_sb_energy_spike_substitution` → **BULLISH** (mirror) — Brent +18% (cost-push) AND ethanol mix shift (physical-supply) BOTH bullish. Counterweight so persona doesn't learn "always call NEUTRAL when complexity is mentioned."

**2 contrast pairs (axis = transmission channel, not headline):**
- `contrast.kc_real_rates_008_a/b` → **BEARISH / NEUTRAL** — 50bp Fed hike held constant; only 5y5y inflation breakeven response (falls 32bp = real rates UP / rises 30bp = real rates flat) flips the answer. Tests nominal-vs-real distinction.
- `contrast.zw_stagflation_009_a/b` → **BULLISH / BEARISH** — September CPI +5.6% YoY held constant; only GDP severity (mild slowdown +0.4% / recession -1.8%) flips the answer. Tests stagflation-vs-recession distinction.

## Validation

- [x] `python -m evals.runner --lint-only` → 254 cases clean (was 248)
- [x] `python -m evals.runner --mode mock --agent macro` → 38/38 pass
- [x] `pytest tests/test_eval_splits.py -q` → 9 split invariants pass
- [x] Adversarials correctly land val-only (per invariant #2)
- [x] Contrast pairs correctly land together by pair_id (per invariant #1)
- [x] Per-agent macro counts: train 21→23, val 11→15

## Statistical impact

[...]

PR #1816, opened 2026-04-25
feat(splits): per-cell RNG isolation — adding cases no longer reshuffles other cells

rozavala:
## Summary

Task #71 (the durable fix from PR #1814 review). Cross-cell stability: adding cases to one cell no longer reshuffles assignments in other cells.

## Root cause

Original `build_splits.py` used a single shared `random.Random(seed)` and advanced it as it processed cells in sorted order. Adding cases anywhere shifted the RNG state at later iterations, mutating unrelated cells' shuffles. Reviewer surfaced this on PR #1814: 6 macro cases shifted agronomist's val set even though agronomist counts didn't change.

## Fix

Per-cell RNG seeded from a stable hash of the cell key + global seed:

```python
contrast cells: random.Random(sha256("contrast:{agent}:42")[:16])
canonical cells: random.Random(sha256("canonical:{agent}:{category}:42")[:16])
```

Adding cases to cell X mutates only cell X's RNG. All other cells produce byte-identical shuffles.

## Pinned regression test

```python
def test_cross_cell_assignment_stable_under_case_addition(all_cases):
    """Synthesize an extra macro contrast pair, assert NO non-macro case
    changes assignment."""
```

## Cost paid

This PR is itself a case-addition event for the splitter algorithm, so existing splits change one-time: **41 cases shift between train/val on the regenerate.** After this lands, future case additions are stable.

## Re-validation against the new (stable) val set

| agent | n | baseline | trained | lift | 95% CI | P(lift > 0) |
|---|---|---|---|---|---|---|
| agronomist | 10 | 0.800 | 0.800 | **+0.0pp** | [-34.1, +34.1] | 0.51 |
| geopolitical | 11 | 0.909 | 0.909 | **+0.0pp** | [-29.6, +29.6] | 0.50 |

## Honest finding worth surfacing

Both agents now show 0pp lift on the stable val set, vs the original PR #1797 measurements of +10.0pp (agronomist) and +9.1pp (geopolitical). **This is the THIRD val sample measured** across three independent samples (with 41+ case shifts); the trained prompts' lift estimate has dropped to near zero each time. **The original lifts likely reflected sample-luck rather than persistent signal.**

**Decision per the pre-registered rule** (task #70: P(lift > 0) > 0.5 → flag, don't auto-rollback): **not shipping a rollback PR for either agent.** Both prompts have NOT regressed (trained accuracy = baseline accuracy on new val); the original "lift" signal was within noise. Actual prompt behavior in production is unchanged.

**What this raises:** production Brier and contribution-score monitoring (per `evals/_eval_optimized/MONITORING_METRICS.md`) is now the load-bearing signal for whether the shipped prompts deliver value. Eval-set evidence at this sample size is too noisy to be conclusive.

## Out of scope

Within-cell stability (adding cases to the SAME cell still reshuffles that cell's contents) is a known residual. If it becomes load-bearing, percentile-based hash assignment within cells is the next step. The user-reported failure (cross-agent reshuffling) is fully resolved.

## What this unblocks

- task #69 (50-case scale-up) was blocked on this. Now safe to add 50 more cases without compounding the reshuffling.
- task #59 (monitoring plan doc) becomes more important — production data is now the load-bearing signal.

## Test plan
- [x] `pytest tests/test_eval_splits.py -q` → 10 passed (9 existing + 1 new cross-cell stability)
- [x] `python evals/splits/build_splits.py` regenerates cleanly; same total counts (163 train, 91 val)
- [x] Re-validate live on agronomist + geopolitical against new val set with CIs (results in description above)
- [ ] After merge: task #69 case scale-up (50 more cases for macro + volatility)
- [ ] Production Brier/contribution monitoring becomes the deciding signal for the shipped agronomist + geopolitical toggles


rozavala:
**Approved — durable fix for the cross-cell split reshuffling I flagged on PR #1814, plus a critical empirical finding worth its own conversation.**

## Implementation is exactly right

**`_stable_hash_int` correctly uses SHA-256, NOT Python `hash()`.** The docstring explicitly calls out why: "PYTHONHASHSEED randomizes hash() per process — would silently break the determinism contract." This is a load-bearing detail; using `hash()` would have left the splits non-reproducible across runs.

**Per-cell RNG isolation:**
- Contrast cells: `_cell_rng(f"contrast:{agent}", seed)` per agent
- Canonical cells: `_cell_rng(f"canonical:{agent}:{category}", seed)` per (agent, category)

Adding cases to one cell mutates only that cell's RNG — other cells produce byte-identical shuffles. The pinned regression test `test_cross_cell_assignment_stable_under_case_addition` synthesizes an extra macro contrast pair and asserts NO non-macro case changes assignment. That's the right pin.

**Adversarials short-circuited (no RNG needed)** — they always go to val per invariant #2, so the per-cell RNG construction overhead is correctly avoided for that path.

**Within-cell residual disclosed honestly:** "adding new cases to the same cell still reshuffles assignments within that cell because the RNG produces different shuffle outputs for input lists of different lengths." That's the next layer of stability if it becomes load-bearing — percentile-based hash assignment within cells. Not pretending the fix is more complete than it is.

**One-time migration cost:** 41 cases shift on this regenerate. After this lands, future case additions are stable. Best time to pay this cost is now, before the corpus grows further.

## The empirical finding

This is the part that needs the most attention from anyone reading the PR — the technical fix is straightforward, but **what the new stable val measurements reveal** is significant.

| agent | n | baseline | trained | lift | 95% CI | P(lift > 0) |
|---|---|---|---|---|---|---|
| agronomist | 10 | 0.800 | 0.800 | **+0.0pp** | [-34.1, +34.1] | 0.51 |
| geopolitical | 11 | 0.909 | 0.909 | **+0.0pp** | [-29.6, +29.6] | 0.50 |

This is the **THIRD val sample** measured (PR #1797's original +10.0pp / +9.1pp, then PR #1814's reshuffled val, now this one). Across three independent samples with 41+ case shifts, the trained prompts' lift dropped to near zero each time. Sample-luck framing is increasingly the most parsimonious explanation for the original measurements.

Pre-registered decision rule applied correctly (task #70: `P(lift > 0) > 0.5` → flag, don't auto-rollback): **no rollback ships.** The trained prompts haven't *regressed* (trained accuracy == baseline accuracy on the new val); they just don't show *improvement* at this sample size. Production trading behavior is unchanged.

The right next move follows from the framing: **production Brier and contribution-score monitoring becomes the load-bearing signal** for whether the shipped agronomist + geopolitical prompts deliver value. Eval-set evidence at N=10-11 is too noisy to be conclusive in either direction. The CIs from #1812 are doing exactly the work they were designed to do — surfacing that the original ship decisions rested on noise-band measurements.

Worth tracking explicitly in `evals/_eval_optimized/MONITORING_METRICS.md` if not already documented: "the eval-trained prompts shipped with sample-luck-tier evidence; production Brier divergence between toggle-on / toggle-off is the deciding signal."

## What this unblocks

- **Task #69 (50-case scale-up)** was implicitly blocked on this — adding 50 cases without per-cell isolation would have rebuilt the entire split via cascading reshuffles, polluting every existing measurement. Now safe to scale up.
- **Task #59 (monitoring plan)** becomes more important — the eval-set's resolving power at N=10-15 isn't enough to distinguish sample-luck from real lift.

## Test plan verified

- `pytest tests/test_eval_splits.py -q` → 10 passed (9 existing + 1 new cross-cell stability)
- `python evals/splits/build_splits.py` regenerates cleanly; same total counts (163 train, 91 val)
- Re-validate live on agronomist + geopolitical against new val set with CIs (results in PR body)

CI green including the Tier-3 dry-run check from #1798.

PR #1831, opened 2026-04-26
feat(evals): Step 4 DA cases batch 1 + first DA baseline + RFC alignment finding

rozavala:
## Summary

First Step 4 Devil's Advocate canonical batch + first DA cached-mode baseline. Surfaces a real RFC-vs-prod-DA gap on `canonical_substantive_objection` and adjusts assertions accordingly.

### Changes

| File | Type | What |
|---|---|---|
| `canonical_weak_decision.001_kc_thin_thesis.yaml` | new | Fills the only canonical DA case_kind without prior coverage. Master at 0.42 confidence + vague reasoning + split specialists → DA blocks. Verified `proceed: false`. |
| `canonical_substantive_objection.001_zw_black_sea.yaml` | new | Multi-commodity coverage for substantive_objection (existing canary is KC). ZW Black Sea thesis at 0.78 confidence. |
| `canary.yaml` | strengthened | master_decision moved from single-factor Q3 thesis → multi-factor 5-day-explicit thesis; specialist_inputs expanded 1 → 2. Necessary regardless of the assertion change. |

### Finding — `canonical_substantive_objection` vs prod DA behavior

The first cached-mode DA snapshot exposed a gap. RFC v3.2 specified `expect_proceed: true` on substantive_objection cases (DA finds risks AND proceeds on a "reasonable" Master decision); prod DA consistently returns `proceed: false` on bullish directional trades, blocking on execution concerns (entry timing, sizing, crowded positioning, defined-risk preferred) even when it agrees the thesis is "directionally plausible".

Three case-content iterations all hit the same wall:
1. Single-factor Q3-horizon thesis (canary original) → blocked
2. Multi-factor 5-day-explicit language (first strengthening) → blocked
3. Fresh-catalyst pre-rally pricing (after-hours news, market not yet processed) → blocked

The model is correctly reading "bullish directional long on a rally" as catastrophe-prone in a 5-day pre-mortem framing. This is robust model behavior, not case-content noise.

**Resolution in this PR:** `expect_proceed` removed from both substantive_objection cases. The case_kind's load-bearing metric is `objection_substance_rate`, which gates only on `risks_min_count` + `recommendation_non_trivial` — both assertions remain. The substantive-risks property is what `objection_substance_rate` actually measures.

**Two follow-up paths possible** (both out of scope here):
- **DA prompt tuning** — adjust pre-mortem framing to distinguish "risks of this trade" from "veto this trade". User memory flags persona changes as fragile/expensive; high-risk path.
[...]

PR #1838, opened 2026-04-26
docs(rfc): tier3_evals v3.3 — DA expect_proceed amendments

rozavala:
## Summary

Formalises the finding from PR #1831 (Step 4 DA cases batch 1) and pre-emptively addresses the same pattern on adversarial DA case_kinds before Step 6 authoring — per the reviewer's recommendation in #1831.

**Pure RFC documentation update — no code changes.** Existing DA cases already conform (PR #1831 dropped `expect_proceed` from substantive_objection cases informally; this PR formalises the decision and extends it pre-emptively). Mock-mode DA run confirms all 8 existing cases still pass.

## Three case_kind amendments

| case_kind | Change | Rationale |
|---|---|---|
| `canonical_substantive_objection` | Drop `expect_proceed: true` | **Proven.** Three iterations in PR #1831 all hit the same wall — prod DA blocks bullish directional trades on execution concerns regardless of thesis quality. |
| `adversarial_strong_decision` | Drop `expect_proceed: true`; raise `risks_min_count: 2 → 3` | Pre-emptive. Same prod-DA framing applies; expecting `proceed=true` would falsely flag correct conservatism. Raised risk count so the case_kind tests *substance* on well-grounded decisions, not restraint. |
| `adversarial_consensus_alignment` | Drop `expect_proceed: true` | Pre-emptive. Same reasoning. Already used substance-axis assertions; only proceed-direction drop. |

## Metric updates

- **`adversarial_restraint_rate` obsoleted.** Under v3.3's understanding, the "restraint" framing tests nothing meaningful (any value consistent with prod behavior).
- **`objection_substance_rate` extended** to cover all 4 substance-bearing case_kinds: substantive_objection + weak_decision + adversarial_strong_decision + adversarial_consensus_alignment.
- **Per-PR red threshold updated:** `adversarial_restraint_rate drops ≥ 10pp` → `objection_substance_rate drops ≥ 10pp`.

## Approach to the prompt-vs-eval gap (new v3.3 §)

The eval matches production behavior, not the other way around (per user feedback memory: persona/prompt changes are fragile/expensive). If production DA's veto-prone behavior is a real problem (false blocks on sound trades), it surfaces in production trade-blocking metrics on the Scorecard, NOT in eval gate failures. The eval's job is regression detection on whatever DA does today; reshaping production DA is a separate concern.

## Test plan

- [x] `pytest tests/ -k 'tier3 or harness or canary'` — passes (no code changed; DA cases unchanged).
- [x] `python -m evals.runner --tier3-only --agent devils_advocate --mode mock` — 8 cases, 0 failed.
- [x] `grep adversarial_restraint_rate docs/rfcs/tier3_evals.md` — all references are now in the deliberate v3.3 changes-log + metrics-section deprecation note.

## What this unblocks

- **Step 6 adversarial DA case authoring** can begin without the `expect_proceed: true` mismatch derailing iteration.
- **Code follow-up** to formally remove `adversarial_restraint_rate` from `Tier3SubtypeMetrics` and the baseline machinery — minor, can land with PR-A.3.5.1 (the missing-metrics infra PR).


rozavala:
**Approved — formalizes the #1831 finding cleanly and pre-emptively addresses the same pattern on the adversarial DA case_kinds before Step 6 derails.**

**Coverage of every implication from the #1831 thread:**

| #1831 review point | v3.3 disposition |
|---|---|
| Drop `expect_proceed: true` from substantive_objection | ✓ done — documented as "proven" with the three #1831 iterations |
| Drop `expect_proceed: true` from adversarial_strong_decision | ✓ done — pre-emptive |
| Drop `expect_proceed: true` from adversarial_consensus_alignment | ✓ done — pre-emptive |
| Obsolete `adversarial_restraint_rate` | ✓ ~~strikethrough~~ + explanation that any value is consistent with prod behavior |
| Extend `objection_substance_rate` to substance-bearing case_kinds | ✓ now covers all 4: substantive_objection + weak_decision + adversarial_strong_decision + adversarial_consensus_alignment |
| Update per-PR red threshold | ✓ `adversarial_restraint_rate` ≥ 10pp → `objection_substance_rate` ≥ 10pp |

**Two thoughtful adjustments beyond the minimal fix:**

1. **`risks_min_count: 2 → 3` on `adversarial_strong_decision`.** v3.2's "nontrivial but minimal" rationale (2 risks) was about not penalizing DA for not-manufacturing-vetoes — under v3.3's understanding the test is substance not restraint, so matching substantive_objection's threshold of 3 is the right floor. Internally consistent.

2. **`proceed_accuracy` redefined to exclude cases without `expect_proceed`** ("Cases without an `expect_proceed` assertion ... are excluded from this metric — the case_kind doesn't gate on proceed direction"). Without this, the metric would misreport — substantive_objection cases without an `expect_proceed` assertion would either fail (if treated as `expect_proceed: true` default) or distort the rate. Cleaner to just exclude.

**The new "Why prod DA can't be assumed to proceed on sound theses" section is the right capstone.** Articulates the structural reason ("the prompt is a pre-mortem by construction") so a future engineer reading the RFC doesn't have to re-derive why the v3.2 framing was wrong from the iteration history. Cites the prompt's own framing — "assume this trade has FAILED CATASTROPHICALLY 5 days from now" — as the load-bearing evidence.

**The "Approach to prompt-vs-eval gap" section captures the user feedback memory explicitly.** "The eval matches production behavior, not the other way around (per user feedback memory: persona/prompt changes are fragile/expensive). If production DA's veto-prone behavior is a real problem (false blocks on sound trades), it surfaces in production trade-blocking metrics on the Scorecard, NOT in eval gate failures." This is the right operational framing — it tells future RFC editors that "the eval is failing because the prompt is wrong" is NOT a license to change the prompt; it's a signal to fix the eval (and check Scorecard separately).

**Status header updated correctly:** "Accepted (v3.3) — Step 0 approved 2026-04-24; Step 1 (PR-B) in flight; Steps 2, 3a, 4 merged" reflects current state. Useful for future readers landing on the RFC.

**Code follow-up correctly scoped as separate.** Removing `adversarial_restraint_rate` from `Tier3SubtypeMetrics` and the baseline machinery is identified as PR-A.3.5.1 work (the missing-metrics infra PR that's already on the follow-up list). Right placement — that PR already touches the metrics dataclass.

**CI doesn't fire on `docs/rfcs/**`** by design (correctly, RFC docs don't need eval-gate runs). PR body confirms mock-mode DA run shows all 8 existing cases still pass — consistent with the assertion change being already applied informally in PR #1831.

PR #1843, opened 2026-04-27
feat(monitoring): operational digest script for eval-trained prompts

rozavala:
## Summary

Operationalizes `evals/_eval_optimized/MONITORING_PLAN.md` as a programmatic digest script. Fills the doc's "future work" hook for an automated weekly check on the agronomist + geopolitical KC deploy from PR #1797.

`scripts/monitor_optimized_prompts.py` reads production `council_history.csv` + `enhanced_brier.json`, applies the decision matrix from `MONITORING_METRICS.md`, and emits a markdown digest to stdout. Exit code reflects worst-case signal.

## Why now

Per reviewer guidance after the macro persona-engineering session: monitoring work was sequenced second, useful but not urgent. The monitoring window is now active (deploy 2026-04-25, current 2026-04-27, market closed weekend). The script is ready before the first KC cycle since deploy lands.

## Tier coverage

| tier | what it checks | implemented |
|---|---|---|
| 1 | Toggle is on AND cycles since deploy actually loaded optimized prompts | ✅ |
| 2 | Per-agent direction-distribution shift, abstention drift | ❌ (dashboard-side per plan) |
| 3 (Brier) | Pre-deploy 4-week trailing Brier vs post-deploy mean per agent; >15% rel Δ → catastrophic-regression alarm | ✅ |
| 3 (Contribution) | Per-cohort contribution split | ❌ (current scores file has no cycle/timestamp metadata; reporting rolling-mean snapshot only) |

## Exit codes (cron integration point)

```
0  no action
1  investigate before next monitoring window
2  rollback recommended per decision matrix
```

The exit code is the wire-to-cron point — alerting (Pushover, GitHub issue) is out of scope for v1.

## Sample output (current production state)

```
# Eval-trained prompt monitoring digest

- Generated: 2026-04-27T02:28:13...+00:00
- Toggle status: enabled=['KC'], disabled=none

## KC

### Tier 1: KC
- Toggle: ✅ ON
- Last council cycle: 2026-04-24T15:04:03...
- Total cycles in history: 756
- ⏳ No cycles with optimized_prompts populated yet. Either no cycles
  since deploy, or loader is not finding the prompt files. Tier 3 skipped.

### Tier 3: KC Brier comparison
- ⏳ No deploy cutoff detected. Skipping until first cycle since deploy populates the column.

---
**Overall:** ✅ no action required.
```

[...]

PR #1845, opened 2026-04-27
fix(ci): unblock Step 7b — python-dotenv + --baselined-only (gate stays dry-run)

rozavala:
## Summary

Originally opened as Step 7b activation; **reverted to ship only the safe infra fixes** after the first live CI run caught two infrastructure bugs that would have failed every PR with the gate active. Per the reviewer's Option B recommendation.

## What stays (safe-to-ship infrastructure)

| Change | Why |
|---|---|
| `python-dotenv>=1.0` added to `requirements-eval.txt` | Real bug fix. `config_loader.load_config` and `calibrate._load_config_for_calibration` both `from dotenv import load_dotenv`. Cached-mode CI dies on `No module named 'dotenv'`. Benefits the existing `full` and `eval-drift-weekly` jobs too. |
| `--baselined-only` runner flag + 2 new tests | Useful machinery regardless of when Step 7b activates. Filters Tier-3 cached gate to subtypes with committed baselines so future Step 3b/c/d auto-join the gate without workflow changes. |
| `evals/baselines/cached/tier3/compliance.json` removed | Placeholder recording 0% pass rate from the canary. Real bug in `_extract_compliance_category` patterns — "no logic violation" matches "logic violation" substring. Step 3d preparation work. |

## What's reverted (Step 7b activation deferred)

- `tier3-adversarial` job back to mock mode + `continue-on-error: true` (Step 7a state from PR #1798).
- Header comment in `eval-gate.yml` updated to document why activation was deferred.

## Two CI infra bugs surfaced

From CI run [24975492052](https://github.com/rozavala/real_options/actions/runs/24975492052):

```
GEMINI_API_KEY: 
OPENAI_API_KEY: 
ANTHROPIC_API_KEY: 
XAI_API_KEY: 
[...]
WARNING __main__: Could not load config for cached mode; cached/live may fail: No module named 'dotenv'
INFO __main__: Run summary: 28 cases, 26 failed
```

1. `python-dotenv` missing — fixed in this PR.
2. `EVAL_*_API_KEY` secrets returning empty in workflow env. Either secrets aren't configured or named differently. The Tier-2 `full` job uses identical syntax but only runs on production-targeting PRs, so we have no prior validation that those secrets exist. **Needs investigation outside this PR.**

## What's next for Step 7b activation

1. Verify which `*_API_KEY` repo secrets exist. If `EVAL_*_API_KEY` aren't configured, either add them or update the workflow YAML to reference the actual names.
[...]

PR #1817, opened 2026-04-25
docs(monitoring): operational plan for shipped eval-trained prompts (task #59)

rozavala:
## Summary

Companion to `MONITORING_METRICS.md`. The metric doc decided **what** to measure and **at what threshold**; this doc decides **when** to look, **who** owns each check, and **how** to act on the data.

## Why now

PR #1816's load-bearing finding: across three independent val measurements (original deploy + two reshuffles), trained accuracy ≈ baseline accuracy on every sample. The eval-set "+10pp lift" that authorized shipping was likely sample-luck. **Production data is now the load-bearing signal for whether the shipped prompts deliver value.** Without an operational plan, the deploy decision is irreversible by inattention rather than by evidence.

## What ships

`evals/_eval_optimized/MONITORING_PLAN.md` (~180 lines, no code).

**Three-tier monitoring schedule:**

| Tier | When | Owner | Check |
|---|---|---|---|
| 1 | First 24h post-deploy | deployer / oncall | Toggle took effect (INFO log greps, v11 column populated) |
| 2 | First week | Weekly oncall | Distributional sanity (≤±10pp shifts, no compliance veto spike) |
| 3 | Week 2+, weekly | Maintainer review | Paired contribution + Brier vs baseline with CIs |

**Decision matrix for Tier 3** (six combinations of contribution × Brier movement → continue / investigate / rollback).

**Expansion gate:** staged rollout KC → NG → SB → ZW with 2-week hold per phase. Hold is default; expansion requires positive evidence.

**Rollback procedure:** single-ticker config edit + deploy. Instant revert to commodity-aware persona on next cycle. Rollback metadata recorded in this doc's changelog table.

**Risks acknowledged:** pre-deploy baseline staleness, tight sample size (window-extension rule from `MONITORING_METRICS.md` applies), three eval measurements pointing to near-zero true lift may resolve as "no harm, no help."

## What this unblocks

- **Task #60** (Brier dashboard `prompt_source` filter) — the doc identifies this as the next concrete implementation gap. Without it, Tier-3 checks require hand-querying the CSV.
- **Production decision-making** — there's now a documented procedure for when/how to rollback, when to expand. Without this, the deploy quietly accumulates without an evidence-based exit.

## Test plan
- [x] Cross-references to `MONITORING_METRICS.md`, `README.md`, PRs #1797 / #1799 / #1812 / #1816 are all to existing artifacts in main
[...]

PR #1803, opened 2026-04-25
docs(monitoring): Brier vs contribution score — pre-registered no-regression criterion

rozavala:
## Summary

Output of task #65 (you flagged this as worth investigating in the #62 PR thread). Decides which signal to use for the no-regression check on PR #1797's eval-trained-prompt deploy to KC, and pre-registers the criterion so task #59 can finalize the monitoring plan without re-litigating the metric question.

## Decision

**Brier as primary, contribution as secondary. Don't combine them.**

## Why both, not just one

| Brier | Contribution | Diagnosis |
|---|---|---|
| Improves | Improves | Working as intended |
| Improves | Flat | Better calibration but master overrides — agent value unclear |
| Flat | Improves | Lucky alignment at small N; revisit later |
| Degrades | Anything | **Rollback** |
| Anything | Degrades sharply | **Investigate** even if Brier looks fine |

Two signals side-by-side surface failure modes that a hybrid metric would hide.

## Pre-registered criterion (excerpted from doc)

**Primary — Brier:**
> Within 2 weeks of PR #1797 deploy, KC agronomist + geopolitical mean Brier scores on resolved decisions are within 10% relative of the pre-deploy 4-week trailing baseline. Window must contain ≥5 resolved decisions per agent (otherwise extend). If Brier degrades >10% relative: rollback that ticker; investigate before re-enabling.

**Secondary — contribution:**
> Same window, contribution within 15% relative (looser; sparser metric). If contribution regresses sharply (>15%) while Brier holds: investigate, do NOT auto-rollback.

## Why not a hybrid (e.g., contribution-weighted Brier)

`trading_bot/dspy_optimizer.py::_compute_metric_score` already does this for *training*. We deliberately don't use it for *monitoring*:

- Two separate signals are clearer than a combined one for go/no-go decisions
- A combined metric hides which axis moved when something regresses
- The training metric is calibrated for the optimizer; monitoring should be calibrated for human judgment under uncertainty

## What this unblocks

- **Task #59** (monitoring plan doc) — was blocked by this; now ground-clear to use Brier as primary
- **Task #60** (Brier dashboard prompt_source filter) — should split by `optimized_prompts` JSON column from PR #1799

## Risks acknowledged

- Pre-deploy baseline staleness (4-week trailing might span regime changes); secondary 2-week pre-deploy anchor proposed
- Lookahead window mismatch between Brier and contribution — confirmed both use the same window per `enhanced_brier.py`
[...]

PR #1847, opened 2026-04-27
feat(monitoring): per-cycle contribution log + cohort split in digest

rozavala:
## Summary

Closes the deferred MVP piece from PR #1843. The original digest could only report a rolling-mean snapshot for contribution because `contribution_scores.json` is a flat list of floats with no cycle/timestamp metadata. This PR adds per-cycle traceability so the digest can split contribution into the with-vs-without optimized-prompt cohorts that `MONITORING_METRICS.md` calls the **primary** check.

### `trading_bot/contribution_scorer.py`
New `contribution_history.csv` alongside the existing rolling-mean json. Each call to `record_contribution()` appends one row: `timestamp, cycle_id, contract, agent, regime, score`. Append-only — no trim, since the digest needs the long-window history. Defensively wrapped: any IO failure logs a warning but does NOT block scoring.

### `scripts/monitor_optimized_prompts.py`
New "Tier 3 contribution comparison" section, marked **PRIMARY** per `MONITORING_METRICS.md`. Joins `contribution_history.csv` against `council_history.csv` on `cycle_id` to split scores into with-vs-without optimized-prompt cohorts:

```
| agent | without-optim mean (n) | with-optim mean (n) | rel Δ | status |
```

Applies the **20% relative degradation threshold** (vs Brier's 15%, since contribution is sparse-but-primary). Excludes NEUTRAL-gated zero-contribution cycles per the metrics doc. Falls back to the rolling-mean snapshot when `contribution_history.csv` is absent (pre-PR-1845 deployments).

### Tests
10 new tests:
- `TestPerCycleHistoryLog` (4): path co-location, header on new file, append no-double-header, empty cycle_id graceful
- Cohort logic (6): JSON-parsed cycle filtering, score filtering (skip zeros), fallback path, sample-size guard, degradation alarm, within-tolerance

All 2030 existing tests still pass.

## Backwards compatibility

- Existing callers don't change. The new file is created lazily on first `record_contribution()` call.
- Existing `contribution_scores.json` format and rolling-mean computation are untouched.
- Pre-PR-1845 production data flows through the fallback path with no degradation in digest output quality (the digest just reports the rolling-mean snapshot instead of the cohort split, same as before).
- No migration script needed — new history file populates from this PR forward.

## Sample output (with synthetic cohort data, would look like this in production once cycles run)

```
### Tier 3: KC contribution comparison (PRIMARY)

- Threshold: ±20% relative Δ (per MONITORING_METRICS.md primary check)
- Min sample: ≥5 non-zero scores per agent per cohort

| agent       | without-optim mean (n) | with-optim mean (n) | rel Δ | status |
|---|---|---|---|---|
| agronomist  | +0.3000 (n=42)         | +0.2800 (n=8)       | -6.7% | ✅ within tol |
| geopolitical| +0.2400 (n=38)         | +0.1500 (n=7)       | -37.5%| 🛑 contribution degraded — investigate / consider rollback |
```

## Test plan

- [x] All 2030 tests pass locally (`pytest tests/ -x -q --timeout=120`)
- [x] Smoke-test against production data shows fallback path (rolling-mean snapshot) since `contribution_history.csv` doesn't exist yet — expected
- [ ] Once orchestrator restarts and runs first cycle on KC after merge, the new file populates and Tier 3 PRIMARY contribution check goes live
- [ ] Eval-gate CI re-measures on push (cached harness)

## What this does NOT do

- No migration script — pre-PR-1845 data has no cycle traceability so it can't be retroactively cohort-split. The history starts fresh from first record after merge.
- No --alert flag (still deferred to future work; exit code is the integration point per PR #1843)
- No changes to `contribution_bridge.py` — bridge already passes `cycle_id` and `contract` through to the scorer


rozavala:
## MECE review verdict — APPROVE

**Value.** Closes the deferred MVP from PR #1843: the digest could only report a rolling-mean snapshot for contribution because `contribution_scores.json` is a flat list of floats. This PR adds the `cycle_id`-keyed log that lets the monitor join contribution against `council_history.csv` `optimized_prompts` (PR #1799 v11) and produce the with-vs-without-optim cohort split that `MONITORING_METRICS.md` calls the **PRIMARY** check (line 4: "Contribution score is the primary no-regression metric"). 20% threshold and N≥5 sample-size guard match the pre-registered numbers in MONITORING_METRICS.md exactly. Proportionate: 4 files, ~400 LOC, 10 focused tests.

**Risk.** Touches `trading_bot/contribution_scorer.py` but only adds a side-effect log; the new `_append_history_row` call is wrapped in try/except (lines 169-178) so a logging IO failure cannot block scoring or the trade cycle — correct fail-soft pattern for logging (CLAUDE.md fail-closed applies to *trade* path, not observability). New CSV uses plain append (correct for an append-only log; analogous to `council_history.csv`); state-manager atomicity rule does not apply. New file under `data/{ticker}/` which is already in `.gitignore`. No regime-name, agent-name, or confidence-vocabulary changes. No PII/secrets in the schema (`timestamp, cycle_id, contract, agent, regime, score`). `contribution_bridge.py` already passes `cycle_id` and `contract` through (verified at lines 161-162, 180-181) — no caller changes needed. Backward-compat: existing `contribution_scores.json` rolling-mean json is untouched, and the digest falls back to the rolling-mean snapshot when `contribution_history.csv` is absent (verified in test `test_render_tier3_contribution_falls_back_when_history_absent`).

**Prior art.** This is the third in a four-PR sequence: #1799 (v11 schema with `optimized_prompts` column) → #1817 (MONITORING_PLAN.md) → #1843 (digest MVP, deferred contribution cohort) → **this PR** (closes the deferral). No duplicate prior attempt; #1803 was closed because it had Brier as primary, replaced by #1807 which made contribution primary — this PR finally implements the primary check.

**Trajectory fit.** Extends the existing `scripts/monitor_optimized_prompts.py` (PR #1843) rather than inventing a parallel pipeline. Joins existing `council_history.csv` `optimized_prompts` JSON (PR #1799). Reuses the existing `EXIT_INVESTIGATE` signal codes. Cohort logic is meaningful (per-agent optimized-prompt presence, not arbitrary slicing). Note: in-code comments reference "PR #1845" but #1845 was a CI fix; the relevant prior PR for the schema column is #1799. Cosmetic, not blocking.

**Tests.** 135/135 pass on the affected surface (`test_contribution_scorer`, `test_monitor_optimized_prompts`, `test_brier_*`, `test_dspy_contribution`, `test_enhanced_brier_math`). PR body asserts all 2030 existing tests pass locally. The trading_bot touch is observability-only and not in the safety-critical execution list (order_manager / ib_interface / compliance / weighted_voting / signal_generator / risk_management / orchestrator), so the focused-scope run is proportionate to the change.

Approving.

Issue #1621, opened 2026-04-16
refactor: Iron Condor wing selection should filter to liquid strikes at strategy.py, not patch at order creation


rozavala:
## Description

The April 15 incident (issue #1615) where Iron Condor order aborted because `OZWN6 C0615` had no market maker quote was a symptom of a deeper architectural issue: the strike selection in `strategy.py` (`find_strike_by_delta`, index-based fallback) picks strikes purely from the option chain definition without verifying market data availability.

PR #1612 initially added a runtime fallback in `create_combo_order_object` (ib_interface.py) that tried adjacent strikes inward when market data was missing. This was reverted because:
1. It silently changed the trade structure — risk validation ran against original strikes, execution used fallback strikes
2. `strategy_def['legs_def']` stayed at original strikes while `qualified_legs` used fallback → mismatch for overlap detection and logging
3. Moving the wing inward changes max loss, max profit, and breakeven without council re-evaluation

## Proposed fix

The right fix is earlier in the pipeline — at strategy selection time:

1. After `define_volatility_strategy` selects strikes (delta-based or index-based), do a pre-flight check: request snapshot market data for the 4 proposed strikes
2. If any wing (BUY leg) has no bid/ask, replace it with the nearest strike that does have a quote, OR skip this expiry entirely
3. Return the validated `strategy_def` with strikes that are confirmed tradeable

This way the risk validation, legs_def, and execution all see the same strikes from the start.

Note: This requires passing an `ib` connection into `define_volatility_strategy`, or doing the pre-flight check in `order_manager.py` between `define_func()` and `create_combo_order_object()`.

Issue #1615, opened 2026-04-16
bug: ZW Option orders aborted due to missing market data (OZWN6)


rozavala:
### Description
At 14:31 UTC, ZW option orders were aborted because market data could not be retrieved.

### Log Snippet
```
2026-04-15 14:31:31,349 - root - ERROR - Insufficient price data for OZWN6 C0615. Bid: None, Ask: None
2026-04-15 14:31:31,351 - root - ERROR - Failed to get market data for OZWN6 C0615. Aborting order.
```

rozavala:
## Diagnostic Results (April 16, 2026)

**Root cause: Dynamic CME strike grid + boundary liquidity — NOT a subscription problem.**

### What the diagnostic confirmed

A targeted diagnostic script () was run against the live OZWN6 chain during active CBOT overnight session (04:29 UTC).

**ZWN6 futures:** bid=602.25, ask=602.5 — connection and subscription healthy.

**Strike qualification results:**
- 18/30 probed strikes qualified (all 10-cent interval strikes)
- 5-cent interval strikes (575, 585, 595, 605, 615, 625, 635) returned **Error 200** (no security definition)
- All 18 qualified strikes had live bid/ask — **0 missing quotes today**

### Why C0615 failed on April 15

CME/CBOT dynamically lists option strikes near ATM. On April 15 with ATM ~595.75:
- The exchange temporarily listed **5-cent interval strikes** (575, 585, 595, 605, 615, 625) near ATM
- C0615 appeared in the chain and was qualifiable (conId returned)
- But it was at the **outer boundary of the temporarily-listed grid** with no market maker quoting it
- The Iron Condor selected C0615 as the outer call wing → market data request returned Bid=None, Ask=None → abort

Today the chain only shows 10-cent intervals (560, 570, 580, ...) — C0615 doesn't even qualify.

### Conclusion

This is **not a subscription issue** and **not a broad ZW data problem**. The abort was expected behavior when an Iron Condor wing lands on a strike that exists in the IB chain but has no active quotes.

### Fix

The Iron Condor wing selection should validate that proposed wing strikes have actual market data before proceeding. If the outer wing has no quote, the strategy should either:
1. Fall back to the nearest qualified adjacent strike that has a live quote
2. Skip Iron Condor entirely for that cycle if no valid outer wings can be found

This prevents selecting strikes at the boundary of the dynamic CME strike grid where liquidity is absent.

rozavala:
[...]

PR #1869, opened 2026-04-28
fix(strategy): pre-strategy affordability gate for LONG_STRADDLE

rozavala:
## Summary
- Closes #1394 (KC straddle compliance rejections — recurring)
- Adds a Black-Scholes pre-strategy affordability gate for LONG_STRADDLE in \`define_volatility_strategy\` so KC (and any future profile where 1-lot straddle premium exceeds the compliance threshold) skips early instead of running the full Tier 2 + Tier 3 + IV-fetch + VaR-refresh + order-build pipeline only to be rejected at compliance.

## The recurring problem

KC at ~\$270 underlying with \$375/pt multiplier produces a 1-lot ATM straddle around \$15-20k of premium. The compliance threshold for KC is \$10k (\`compliance.py:649\`, \`config/profiles/kc.json:411\`). Math:
- Notional per lot ≈ 270 × 375 = \$101k
- ATM straddle ~5-20% of notional depending on IV/DTE
- KC profile IV=0.35 + 8-month expiry → BS straddle ≈ \$19,463

The original issue was filed 2026-03-23, closed without a fix. Today's KCZ6 LONG_STRADDLE attempt at 15:04:01 UTC:

\`\`\`
2026-04-27 15:04:01 - signal_generator - WARNING - CONVICTION GATE: Suppressing BEARISH signal for KCZ6 (202612).
2026-04-27 15:04:01 - signal_generator - INFO - STRATEGY SELECTED: LONG_STRADDLE | catalyst=Drought conditions (urgent)
2026-04-27 15:04:01 - root - INFO - Creating Limit Order for BUY 1 @ 53.10 (Adaptive Cap: 56.05).
2026-04-27 15:04:13 - OrderManager - INFO - Pre-batch VaR refresh complete  ← whole pipeline ran
2026-04-27 15:04:16 - compliance - WARNING - Straddle risk \$19,912 exceeds threshold \$10,000 for Coffee (Arabica)
2026-04-27 15:04:16 - OrderManager - WARNING - Order blocked by Compliance: Straddle risk too high for account size
\`\`\`

\`compliance.py\` correctly catches it but only after ~30s of fetch + LLM + IB-API work that won't produce a trade.

## Fix

[...]

PR #1612, opened 2026-04-15
fix: Error 478, VaR resilience, ZW suffix, duplicate accumulation (#1613), FIFO leg-deduction (#1614)

rozavala:
## Summary

- **Error 478 / VaR SB=F (#1607, #1608)** — BAG symbol/exchange derived from \`combo_legs[0].contract\` instead of KC-default config. VaR continues per-symbol on yfinance failure instead of aborting all VaR. ZW suffix added to all 3 missing maps (\`dashboard_utils.py\`, \`6_Signal_Overlay.py\` ×2).
- **Duplicate spread accumulation (#1613)** — \`_check_contract_overlap\` Check 2 blocks same-direction re-entry of an identical spread.
- **FIFO leg-deduction fix (#1614)** — Root cause of false "reverse orphan" alert: the \`any_present\` fail-closed branch kept a partially-covered thesis alive but did not deduct its present legs from \`remaining_ib\`. Fix: deduct present legs, clamped to \`min(exp, avail)\` / \`max(exp, avail)\` to prevent phantom opposite-sign positions. D3 notification-only with enriched diagnostics.

## Issues fixed
- Closes #1607
- Closes #1608
- Closes #1613
- Closes #1614

## Root-cause issues opened (deferred)
- #1619 — \`position_monitor\` uses single KC config for all commodities in risk closure
- #1620 — consolidate 4 independent exchange→yfinance suffix maps
- #1621 — Iron Condor wing selection: filter to liquid strikes at \`strategy.py\` (replaces reverted wing fallback; #1615 reopened)
- #1622 — signal/strategy layer is stateless w.r.t. existing positions (cross-cycle dedup)

## Test plan
- [x] \`pytest tests/test_order_manager.py::TestCheckContractOverlapAccumulation\` — 6 tests for duplicate spread guard
- [x] \`pytest tests/test_exit_integration.py::TestReconcileOrphanedThesesAggregation\` — 9 tests including FIFO false-orphan regression and over-deduction clamp
- [x] \`pytest tests/test_ib_interface.py::TestCreateComboOrderMissingMarketData\` — tests abort-on-missing-market-data (extracted from removed dead class)
- [x] \`pytest tests/test_var_calculator.py::test_var_partial_symbol_failure_continues tests/test_var_calculator.py::test_var_all_symbols_fail_returns_none\` — VaR partial-data behavior
[...]

PR #1871, opened 2026-04-28
fix(strategy): require Greeks for Iron Condor — disable position-based fallback by default

### LINKED ISSUE #1855: bug: IRON_CONDOR strategy selection produces structurally unfilled spreads — 12+ selections, 0 fills
## Summary

Iron Condor has been selected by the strategy router **at least 12 times** in the available log history (10 on 2026-04-20, 2 on 2026-04-27) and has **never filled**. Two distinct failure modes have been observed; both are classes of \"strike selection produces theoretically valid but practically untradeable spread.\"

This is the operational consolidation issue for #1621 (closed 2026-04-17 with the architectural fix described but not implemented).

## Ground truth (across all log files)

\`\`\`
$ grep -h \"STRATEGY SELECTED: IRON_CONDOR\" logs/*.log | awk '{print substr(\$1,1,10)}' | sort | uniq -c
     10 2026-04-20
      2 2026-04-27

$ grep -hE \"IRON_CONDOR|Iron Condor\" data/*/archive_ledger/*.csv | wc -l
0
\`\`\`

## Failure mode A: Wing strike has no market data (#1615 / #1621)

The April 15 incident: \`OZWN6 C0615\` had no market maker quote, so the order aborted at runtime. PR #1612 added a runtime fallback (substitute adjacent strikes) but was reverted because risk validation ran against original strikes while execution used fallback strikes. #1621 proposed moving the fix earlier in the pipeline (\`strategy.py\`'s \`define_volatility_strategy\` should pre-flight market data on selected strikes) but was closed without implementation on 2026-04-17.

## Failure mode B: Negative-credit floor (NEW today)

\`\`\`
2026-04-27 10:09:24 - signal_generator - WARNING - CONVICTION GATE: Suppressing BEARISH signal for KCU6 (202609). |weighted_score|=0.0738 < threshold=0.2. Overriding to NEUTRAL (no trade).
2026-04-27 10:09:24 - signal_generator - INFO - STRATEGY SELECTED: IRON_CONDOR | regime=RANGE_BOUND, vol=BEARISH
2026-04-27 10:09:27 - root - INFO - Using index-based Iron Condor (no Greeks available or delta logic failed)
2026-04-27 10:09:40 - OrderManager - INFO - Iron Condor risk calc: width=5.0, dollar_mult=375.0, qty=1, max_loss=\$1,875.00
2026-04-27 10:09:44 - root - INFO - LIQUIDITY FILTER PASSED: Market spread (2.70) within allowed (4.00, binding=TICK_FLOOR).
2026-04-27 10:09:44 - root - WARNING - RISKLESS COMBO FILTER: SELL spread has zero/negative floor price (-3.90). IB would reject as riskless. Skipping.
\`\`\`

The IC builder fell through to the **index-based fallback** (\`no Greeks available or delta logic failed\`), then selected strikes whose net structure has a negative theoretical floor — meaning the market math says it's a riskless arbitrage. IB's pre-trade check correctly rejects this; the order never reaches the wire.

This is a sibling defect to A: same root cause (strike selection doesn't validate the resulting spread is actually tradeable), different surface symptom.

## Why both modes recur

\`define_volatility_strategy\` in \`strategy.py\` selects strikes by:
1. Delta-target (preferred) — requires Greeks from market data
2. **Index-based fallback** — purely chain-position math when Greeks unavailable

The fallback is what fires when Greeks are missing (timeouts, deep-OTM, illiquid). It doesn't validate:
- That the chosen strikes have live quotes (mode A)
- That the resulting net premium is positive after spread/slippage (mode B)
- That the four legs together produce a structure IB will accept

## Cost

- Every IC selection runs a full Tier 2 + Tier 3 pipeline (~\$0.40-\$0.60 of LLM cost) plus IB API calls, ending in a no-op
- Conviction-gate path (today's two cases): a bearish signal got suppressed to NEUTRAL → IC selected → IC rejected. Net effect: a real bearish signal was discarded *and* no vol-substitute trade replaced it. Brier reliability tracking gets a NEUTRAL / NO TRADE outcome instead of either the suppressed bearish signal or a credible IC trade.
- Council eats budget without producing any executable position whenever the IC path is the chosen vol strategy.

## Suggested fix (one of)

The #1621 architectural approach is still the right shape: validate strikes at strategy-selection time, before the order builder commits.

Concretely, add to \`define_volatility_strategy\` (or in \`order_manager.py\` between \`define_func()\` and \`create_combo_order_object()\`):

1. **Quote pre-flight**: snapshot bid/ask for the 4 selected strikes; if any wing has no quote, replace inward to the nearest quoted strike, or skip the expiry. (Mode A.)
2. **Net-premium pre-flight**: compute the theoretical net credit/debit using mid prices; if a credit IC has negative net (mode B), reject and either retry with wider wings or skip. (Mode B.)
3. **Index-fallback hardening**: when Greeks are unavailable, instead of falling through to position-based strike picks, refuse to construct an IC at this expiry. Force the strategy router to return NO TRADE rather than build a structurally invalid spread.

(3) is the most conservative — it accepts that without Greeks we can't pick safe strikes, and prefers no trade over a guaranteed-rejection trade.

## Cross-references

- #1394 — KC straddle compliance rejections (just reopened — same class of issue: vol-strategy execution layer is structurally blocked)
- #1621 — IC wing selection refactor (closed without fix; this issue subsumes it)
- #1615 — original IC market-data wing failure
- #1612 — reverted runtime-fallback PR
- #1268 — emergency-shadow routing mismatch (closed; surfaced the IC-doesn't-fill pattern earlier)

rozavala:
## Summary
- Closes #1855 (Iron Condor 12+ selections, 0 fills)
- Subsumes the architectural fix proposed in #1621 (closed without implementation)
- Mirrors the B-S delta enrichment already done by \`define_directional_strategy\` so Iron Condor can run the delta-based path on the same chain shape from \`build_option_chain\` instead of falling through to a position-based picker that's been producing structurally-invalid spreads.

## The recurring problem

12 IC selections across available log history, **0 fills**. Two distinct failure modes both rooted in the index-based fallback:

| Failure mode | Surface symptom | Filed under |
|---|---|---|
| Wing strike has no market data | Order aborts at runtime | #1615 / #1621 |
| Net credit is negative | \`RISKLESS COMBO FILTER: SELL spread has zero/negative floor price (-3.90)\` — IB rejects as riskless arb | observed live 2026-04-27 KCU6 |

Both come from the same root: when Greeks are missing, \`define_volatility_strategy\` falls through to an index-based picker (\`iron_condor_short_strikes_from_atm\` + \`iron_condor_wing_strikes_apart\`) that selects strikes purely by chain position without validating that the resulting spread is tradeable.

## Fix

Three changes in \`define_volatility_strategy\`:

### 1. B-S delta enrichment for IC

When the chain has no Greeks (the normal shape from \`build_option_chain\` in \`ib_interface.py\` — only \`strikes_by_expiration\`, no \`calls\`/\`puts\` lists), enrich it with B-S deltas using the profile's \`fallback_iv\` and \`risk_free_rate\`. Same call to \`_enrich_chain_with_bs_deltas\` that \`define_directional_strategy\` already uses for the same chain shape. The delta-based path (well-spaced 16/5-delta wings) now runs on the same chain that previously fell through to index picks.

### 2. Index-based fallback disabled by default

When the delta-based selection genuinely cannot produce valid wings (sparse chain, BS-enriched deltas don't bracket targets), return \`None\` rather than build a position-based spread. The strategy router falls through to NO TRADE — \`signal_generator\` records the cycle, no order build, no IB pre-trade rejection.

### 3. Legacy path preserved behind a toggle

\`iron_condor_allow_index_fallback\` (default \`False\`) re-enables the position-based picker per-profile. Profiles whose chains are dense enough that index picks reliably produce positive-credit spreads can opt in; the unguarded default protects everything else.

## Why this works (empirical check)

For today's KCU6 case (the negative-credit-floor failure):
- Pre-fix: index path picked strikes 2 + 2 from ATM regardless of whether the math produced a positive credit. Result: net credit -3.90, IB rejected.
- Post-fix: BS enrichment runs (KC has no live Greeks at strategy-define time), delta-based picker selects 16/5-delta wings whose spreads are structurally guaranteed positive credit by Black-Scholes pricing. Or, if the chain is too sparse for clean delta picks, returns None — preferable to building a guaranteed-rejection order.

## Risk assessment

- **No safety-critical impact**: only changes which strategies *get built*, not how orders are routed/filled/closed.
- **No new IB API calls**: enrichment is pure analytic, reuses functions already imported by this module.
- **Conservative profile-lookup fallback**: if \`get_active_profile\` raises, defaults to IV=0.30 / r=0.05 (matches \`CommodityProfile\` defaults).
- **Backwards compatibility for profiles that need the legacy path**: \`iron_condor_allow_index_fallback: true\` in \`config.json.strategy_tuning\` restores the old behavior verbatim.

## Test plan
- [x] \`test_chain_without_greeks_runs_through_bs_enrichment\` — load-bearing: delta path runs on a Greeks-less chain after enrichment; chain ends up with populated \`calls\`/\`puts\`; produces a 4-leg SELL spread.
- [x] \`test_index_fallback_disabled_by_default_returns_none\` — sparse chain where delta path can't produce ordered wings: returns \`None\` (no fallback) or a delta-path spread that satisfies \`lp<sp<sc<lc\`.
- [x] \`test_index_fallback_opt_in_still_works\` — \`iron_condor_allow_index_fallback: True\` restores the legacy index-based picker exactly as before.
- [x] Existing \`test_define_low_vol_strategy_iron_condor\` updated to pin the legacy path via the new toggle, preserving original index-picker coverage.
- [x] Regression scope: \`test_strategy.py + test_order_manager.py + test_signal*.py\` → 73/73 passing
- [ ] Live verification deferred to next NEUTRAL→IRON_CONDOR selection. Expect either (a) a clean delta-based IC build with positive net credit + actual fill, or (b) a clean NO TRADE log line — never another riskless-combo filter rejection.


rozavala:
## Verdict: APPROVE — Ready to be merged

Behavioral change is justified, fail-closed semantics are correct, scope is tight, and the empirical record (12+ selections, 0 fills) makes the prior fallback objectively worse than NO TRADE.

### MECE

**1. Value (high)**
- Issue #1855 documents the failure: index-based picker selected strikes purely by chain position, producing two structurally invalid spreads — wing-strike-no-quote (#1615) and **negative-credit floor** (KCU6 2026-04-27 — IB rejected as riskless arb at `-3.90`). 12 IC selections in available logs, 0 fills.
- The PR doesn't merely "disable" — it **upgrades**: it mirrors the B-S delta enrichment that `define_directional_strategy` already does for the same `build_option_chain` shape. So the common case (Greeks missing, KC chain at strategy-define time) now runs the *same* delta-based picker (16/5-delta wings) that all four other strategies already use. Index fallback is the failure path, not the primary path.
- Subsumes #1621 (closed without implementation). One coherent fix vs. an open architectural debt.

**2. Risk (correctly contained)**
- **Scope is IC-only.** Diff touches only the `signal['level'] == 'LOW'` branch of `define_volatility_strategy`. Long Straddle (`level=='HIGH'`) at line 377-378 is identical to pre-PR. Bull/Bear Call/Put live in `define_directional_strategy` (separate function), unchanged. Verified.
- **Fail-closed correct.** When delta path can't produce ordered wings AND no opt-in toggle, return `None`. Both callers (`order_manager.py:1677`, `orchestrator.py:5161`) already gate on `if strategy_def:` — None propagates as NO TRADE without errors.
- **Observability present.** WARNING log at suppression includes `localSymbol`, "(chain too sparse or BS-enriched deltas didn't bracket the targets)", and `#1855` reference. Specific enough to filter logs for "IC suppressed" rate.
- **Brier/risk continuity intact.** `risk_management.check_iron_condor_theses` and `brier_scoring`'s IC outcome scoring both operate on *executed* positions. Suppression at strategy-define means no thesis recorded — same as any other no-trade cycle. No Brier gap that didn't already exist for compliance-rejected ICs.
- **Reversibility:** `iron_condor_allow_index_fallback: True` per-profile restores legacy verbatim. Default `False` is the safe choice; flag preserves an escape hatch.
- **Profile-lookup safety:** `try/except` around `get_active_profile` falls back to IV=0.30 / r=0.05 (matches `CommodityProfile` defaults). Conservative.

**3. Prior art**
- #1621 closed 2026-04-17 architecturally proposing this exact fix; this PR implements it. #1612 reverted runtime fallback (different issue: risk validated original strikes while exec used fallback). This PR puts the validation upstream at strategy-define time, avoiding that desync.
- No prior toggle-flip cycle on this fallback that I found.

**4. Trajectory fit**
- Complements #1869 (LONG_STRADDLE affordability gate) — both are pre-strategy validation guards that prefer NO TRADE over guaranteed-rejection orders. A "strategy pre-validation" pattern is emerging across the four strategies; this is the third one (after directional already had this via delta path, and #1869 for straddles). Iron Condor catches up here.

### Verification

- **Diff scope:** IC-only (LOW vol branch). +186/-3, 2 files. Confirmed.
- **Default value:** `iron_condor_allow_index_fallback: False` by default at line 453. `iron_condor_use_delta: True` by default at line 382.
- **Failure logging:** WARNING with full context at line 454-461. Specific.
- **Tests cover both branches:**
  - `test_chain_without_greeks_runs_through_bs_enrichment` (delta path runs after BS enrichment, 4-leg SELL produced)
  - `test_index_fallback_disabled_by_default_returns_none` (sparse chain, no opt-in → None)
  - `test_index_fallback_opt_in_still_works` (toggle restores legacy)
  - Existing `test_define_low_vol_strategy_iron_condor` updated to pin legacy via toggle.
- **No regression on other strategies:** Long Straddle path at line 377-378 unchanged. Directional unchanged.

### Tests

Full suite: `pytest tests/ -x -q --timeout=120 --ignore=tests/test_dashboard_pages.py` → **2049 passed, 0 failed** in 6m 39s. No IC-related test was passing pre-PR and is now failing — the only change is `test_define_low_vol_strategy_iron_condor` was *intentionally* pinned to the legacy path via the new toggle to preserve coverage of the index-picker contract for opt-in profiles.

### Action

Adding `Ready to be merged`. Recommended monitoring after merge: grep `"Iron Condor: delta-based selection did not produce valid wings"` in production logs to track suppression rate. If it spikes, that's the signal to investigate KC chain sparsity or tune `iron_condor_sell_delta` / `iron_condor_buy_delta` defaults.

PR #1895, opened 2026-04-29
docs(inventory): v2 engineering attempt — no ship, structural finding holds

rozavala:
## Summary

Pre-registered v2 engineering protocol after PR #1889 model swap resolved dispatch fragility. Three iterations; **result: NO SHIP**. Best iter result: 0.0pp (iter 2). All other iters negative.

## Per-iteration results

| iter | edit | val acc | lift vs OLD |
|---|---|---|---|
| OLD (post-swap baseline) | bare INVENTORY_CONTEXT | 15/17 (0.882) | — |
| 1 | broad pace-vs-seasonal + data-source-divergence | 13/17 (0.765) | **−11.76pp** |
| 2 | narrow pace + strengthened data-source | 15/17 (0.882) | **0.0pp** (best) |
| 3 | data-source-divergence only | 14/17 (0.824) | **−5.88pp** |

## What we learned

- **The data-source-divergence rule fixed its target** (`sb_implied_vs_reported_stocks` BULLISH → NEUTRAL) cleanly each iteration — the rule design is sound
- **But each iter ALSO broke 1-3 directional cases** — different cases each time, suggesting non-deterministic model interpretation of MANDATORY-rule additions
- **Inventory persona is rule-fragile** even with the dispatch fragility resolved. The same engineering protocol that shipped 4 other agents (macro/volatility/technical/supply_chain) at +10-13pp produces 0pp or negative lifts on inventory across multiple distinct rule designs.

## The cumulative inventory engineering arc

| PR | What happened |
|---|---|
| #1865 (val coverage) | added 5 inventory val cases; surfaced inventory not actually 1.000 |
| #1874 (engineering v1) | hit dispatch errors; "structural ceiling" framing |
| #1879 (engineering retry) | confirmed dispatch fragility |
| #1887 (root cause finding) | identified Haiku as the cause |
| #1889 (model swap) | swapped Haiku → gpt-4o-mini; resolved dispatch fragility |
| this PR (engineering v2) | engineering still doesn't clear bar |

**Cumulative val direction-accuracy: 15/17 = 88.2% (unchanged from start to finish, but with much cleaner dispatch reliability via PR #1889).**

## Why this validates the discipline

5 attempts, honest measurements, only 1 substantive ship (the model swap), no false-positive ships. The +9pp bar correctly filtered noise from real lift on a hard-to-engineer agent.

## What ships in this PR

- **No persona changes** — INVENTORY_CONTEXT reverted to OLD state.
[...]

PR #1877, opened 2026-04-28
feat(supply_chain): manual persona engineering — ship at +12.5pp val lift

rozavala:
## Summary

Same protocol as macro/volatility/technical (PRs #1840/#1844/#1870) applied to the fourth ship-blocked agent surfaced by the "1.000-baseline-agent val coverage" series. Two iterations; **iter 2 ships at +12.5pp val direction-accuracy lift (13/16 → 15/16)**, above the +9pp threshold.

Pre-registered protocol committed before reading any --validate output: see commit 0fbbb6acf and `evals/_eval_optimized/SUPPLY_CHAIN_PERSONA_ENGINEERING_LOG.md`.

## Changes

`trading_bot/prompts/base_prompts.py` (LOGISTICS_CONTEXT) — adds two MANDATORY rules covering basis-delta failure modes the existing persona didn't handle:

1. **Basis compression delta**: when origin/destination basis MOVES FROM elevated back toward the 5y seasonal average, the COMPRESSION ITSELF is the directional signal (BEARISH on the delta). Includes an explicit anti-trigger: "Do NOT default to NEUTRAL on phrases like 'basis at seasonal norm' or 'metrics restored to normal' when the evidence describes a PATH from elevated → normal. The destination state is not the signal — the PATH is."

2. **Operational normalcy with flat basis**: when activity metrics (rail loadings, port throughput) are elevated above seasonal average BUT basis is flat / unchanged, the elevated activity reflects demand strength operating WITHIN system capacity. Tag NEUTRAL.

## Per-iteration measurements

| iter | edit | val acc | lift vs OLD | decision |
|---|---|---|---|---|
| 1 | basis-compression-delta + operational-normalcy rules | 14/16 (0.875) | +6.25pp | iterate (operational-normalcy fired; basis-compression didn't despite trigger language in case) |
| 2 | strengthen basis-compression with anti-trigger + different example | **15/16 (0.938)** | **+12.5pp** | **SHIP** |

## Per-case status

**2 of 3 pre-registered failures fixed:**
- `contrast.zw_basis_polarity_007_b`: BEARISH → NEUTRAL ✓ (operational-normalcy rule, iter 1)
- `contrast.rail_basis_004_b`: NEUTRAL → BEARISH ✓ (basis-compression rule, after iter 2 anti-trigger strengthening)

**1 remains wrong:**
[...]

PR #1870, opened 2026-04-28
feat(technical): manual persona engineering — ship at +12.5pp val lift

rozavala:
## Summary

Same protocol as macro PR #1840 / volatility PR #1844 applied to the third ship-blocked agent surfaced by the "1.000-baseline-agent val coverage" series (PRs #1852/#1857/#1864/#1865). Two iterations; **iter 2 ships at +12.5pp val direction-accuracy lift (13/16 → 15/16)**, above the +9pp threshold.

Pre-registered protocol committed before reading any --validate output: see commit 5ad882ba7 and `evals/_eval_optimized/TECHNICAL_PERSONA_ENGINEERING_LOG.md`.

## Context

The technical persona was **severely under-specified** vs the other personas — it had no "Sentiment Rule (MANDATORY)" section like macro/volatility/agronomist do. Just listed 4 things to analyze (trend, S/R, momentum, volume) with no directional rules. That's why it defaulted to NEUTRAL on anything ambiguous, including obvious range-bound mean-reversion setups and multi-timeframe conflicts.

## Changes

`trading_bot/prompts/base_prompts.py` (TECHNICAL_CONTEXT) — adds the first MANDATORY sentiment-rule section the technical persona has ever had:

- **Range mean-reversion at SUPPORT (BULLISH)**: price within ~1-2% of established range support that has held ≥2x in 4-8 weeks AND rejection candle (long lower wick + close in upper half) → BULLISH
- **Range mean-reversion at RESISTANCE (BEARISH, mirror)**: same condition at resistance with long upper wick + close in lower half → BEARISH. **Parallel paragraph structure** (not nested bullet) so both directions are equally salient to the LLM parser — iter 1 lesson
- **Multi-timeframe conflict**: when evidence explicitly describes higher-timeframe (weekly) trend conflicting with lower-timeframe (daily) signal → tag the higher-timeframe direction

## Per-iteration measurements

| iter | edit | val acc | lift vs OLD | decision |
|---|---|---|---|---|
| 1 | range rule (bullets) + trend continuation | 13/16 (0.813) | 0.0pp | iterate (1 fix + 1 break from over-broad trend rule) |
| 2 | drop trend rule, parallel range rules, multi-TF | **15/16 (0.938)** | **+12.5pp** | **SHIP** |

## Per-case status

**3 pre-registered failures fixed:**
- `contrast.range_001_a`: NEUTRAL → BULLISH ✓ (range support + rejection rule, iter 1)
[...]

PR #1898, opened 2026-04-29
fix(prompts): loader accepts demos-only optimized prompts

rozavala:
## Summary

**Production bug fix.** Discovered while inspecting production monitoring state: agronomist + geopolitical shipped optimized prompts have been **silently inactive for 2+ days** because the loader was rejecting them.

## The bug in production logs

```
WARNING - Optimized prompt /.../agronomist/prompt.json missing 'instruction' key, skipping
WARNING - Optimized prompt /.../geopolitical/prompt.json missing 'instruction' key, skipping
```

These warnings appeared on every KC cycle since the last orchestrator restart. As a result:

- Both shipped prompts have been INACTIVE in production
- `council_history.csv` shows `optimized_prompts: {}` for every recent cycle
- The monitoring digest (PR #1843/#1847/#1849) sees no cohort data because there's no with-optimized cohort
- **The ship intent of PR #1797 (agronomist +10pp / geopolitical +9.1pp lift on val) has not actually materialized in production**

## Root cause

The shipped `prompt.json` files have:
```json
{
  "instruction": "",         <-- empty (demos-only optimization)
  "demos": [...4 entries...] <-- the actual payload
}
```

The loader's validation:
```python
if not isinstance(data, dict) or not data.get("instruction"):
    logger.warning(...missing 'instruction' key...)
    return None
```

`data.get("instruction")` returns `""` (empty string), which is falsy in Python → the prompt is rejected. This silently disabled the **demos-only optimization path** that the trainer explicitly supports and that `build_specialist_prompt` correctly handles downstream as `prompt_source = "dspy_optimized_demos_only"`.

## Fix

`trading_bot/agents.py:_read_optimized_prompt_file` now accepts any prompt with at least ONE of:
- non-empty `instruction` (REPLACE-mode)
- non-empty `instruction_supplement` (APPEND-mode, MIPROv2 RFC)
- non-empty `demos` array (demos-only — what agronomist/geopolitical actually ship)

Empty/missing across all three still rejects with a clearer warning.

## Tests

13 new tests in `tests/test_optimized_prompt_loader.py`:

- ✅ demos-only payload accepted (the regression case)
- ✅ instruction-only accepted
- ✅ supplement-only accepted
- ✅ combined supplement + demos accepted
- ✅ empty optimization rejected
- ✅ non-dict JSON rejected
- ✅ missing file → None
- ✅ malformed JSON → None
- ✅ defensive: demos as non-list types
- ✅ smoke tests: actual shipped agronomist + geopolitical `prompt.json` files load successfully (these would have failed pre-fix)

All 2090 tests pass.

## Production impact after merge

Once deployed:
1. Next KC cycle will load the optimized prompts (INFO log: `Optimized prompt loaded for agronomist from eval-case fallback...`)
2. `council_history.csv optimized_prompts` column will populate with `{"agronomist": {"source": "dspy_optimized_demos_only", "demo_count": 4}, "geopolitical": {...}}`
3. The monitoring digest will start getting real cohort data
4. The PR #1797 deploy intent (+10pp / +9.1pp val lift) will finally be active

This is also why the inventory engineering arc + the cumulative work on macro/volatility/technical/supply_chain has only delivered eval-set lift so far — the demos-only loader path that ships those personas has been broken end-to-end.

## Why we caught it now

The monitoring digest from PR #1843 was the right tool for this — it surfaces "toggle ON but no cycles loaded the prompt" as an INVESTIGATE signal. I noticed it during a routine production check and traced through the code to find the root cause.


rozavala:
## Verdict: Approve — production bug fix unblocking shipped optimization

**1. Value (high — silent prod regression):** The bug is a real, two-day-old silent failure: shipped agronomist + geopolitical optimized prompts (PR #1797's +10pp / +9.1pp val lift) have been INACTIVE in production because `data.get("instruction")` returns `""` (falsy) and the loader rejected the file. `council_history.csv optimized_prompts={}` and the monitoring digest's "toggle ON but no cycles loaded" signal (PR #1843) caught this. The fix unblocks the demos-only path that the trainer always supported and that `build_specialist_prompt` already handles as `prompt_source = "dspy_optimized_demos_only"` (specialist_prompt.py:118). Not new workflow scaffolding — restoring an end-to-end-broken existing one.

**2. Risk (low):** Three orthogonal acceptance branches (instruction OR supplement OR non-empty demos list); empty-across-all still rejects with a clearer warning; non-dict / malformed JSON / missing file all return None (fail-closed — caller falls through to legacy persona). No regime/confidence vocab touched, no schema change, persona output unchanged so Brier continuity preserved. Per `feedback_eval_persona_model_swap_fragile.md`, this is in the safer category — loader plumbing, not persona-text engineering.

**3. Prior art:** `_read_optimized_prompt_file` was added by the eval-trainer ship arc (PR #1797 et al.); the supplement branch by PR #1798 (instruction_supplement append semantics). The demos-only path was supported downstream from day one — only the validator was over-restrictive. No prior PR has tried to relax this; this is the first.

**4. Trajectory fit:** Strong. Once merged, the cumulative engineering arc (macro #1840 / volatility #1844 / technical #1870 / supply_chain #1877 / inventory v2 #1895) finally gets cohort data in production, enabling the optimization-floor finding's empirical follow-up.

**Specific verification:**
- Layer: `trading_bot/agents.py:_read_optimized_prompt_file` only — correct (loader, not `base_prompts.py` / persona text).
- Backward compat: `bool(data.get("instruction"))` branch preserves all REPLACE-mode files; `instruction_supplement` branch preserves MIPROv2 outputs.
- Demos-only test: `test_loader_accepts_demos_only_config` + smoke tests against actual shipped `agronomist/prompt.json` + `geopolitical/prompt.json` (would have failed pre-fix).
- Malformed input: tests cover non-dict JSON, missing file, malformed JSON, demos-as-non-list, empty-across-all-three (all return None or rejected with warning).
- `base_prompts.py`: UNCHANGED (diff stat: 2 files, both expected).

**Test result:** `pytest tests/ -x -q --timeout=120` → **2090 passed** in 419.53s (full suite, production-touching code per `feedback_full_test_suite_for_execution_prs.md`).

**Action:** Labeled `Ready to be merged`.

Issue #1749, opened 2026-04-23
refactor: #1670 Phase 2 remaining steps — slot-based IB pool allocator + legacy cleanup


rozavala:
Tracking issue for the remaining work on #1670's Phase 2 refactor after Step 1 (#1745) and Step 2 (#1748).

## What's done

- **Step 1** (#1745, merged) — every order gets a ticker-prefixed `orderRef` at origination via `_make_order_ref(ticker)`.
- **Step 2** (#1748, in review) — `DISCONNECT_GUARD` can filter by `orderRef` prefix behind `connection_pool.use_orderref_filtering` (default `false`); both paths are evaluated on every non-terminal trade and disagreements are logged so we can validate before flipping.

## Step 3 — Slot-based allocator + caller migration

Per the architecture in #1670's issue body: replace the keyed pool (`KC_reader`, `KC_orders`, `master_orders`, ...) with a fixed-size slot pool so socket count stops scaling with engine count.

```
Read pool:   2 shared sockets (round-robin across all engines)
Order pool:  2 shared sockets (orderRef-tagged, engine-agnostic)
Master:      1 socket (equity_service + monitor)
```

At 4 engines: 5 sockets. At 10 engines: 5 sockets. At 30 engines: 5 sockets.

### Work items

- [ ] Add `acquire_read()`, `acquire_order(ticker)`, `acquire_master()` to `IBConnectionPool`.
- [ ] Internal state: `_read_slots: List[Optional[IB]]`, `_order_slots: List[Optional[IB]]`, `_master_slot`, `_socket_ownership: Dict[int, Set[str]]` (id(ib) → tickers bound to that slot).
- [ ] Liveness failover: round-robin read pool skips unhealthy slots; order pool prefers already-bound slots then least-loaded.
- [ ] Keep `get_connection(purpose, config)` as a deprecated thin wrapper for one release — routes purposes to `acquire_*` via purpose lookup.
- [ ] Migrate the 14 call sites enumerated in #1670's issue body (orchestrator.py × 8, commodity_engine.py, order_manager.py × 5, master_orchestrator.py, position_monitor.py).
- [ ] Expose pool sizes in `config.json`: `connection_pool.read_slots`, `connection_pool.order_slots` (default 2/2).
- [ ] Integration test: socket count stays at exactly 5 with 4, 6, 9 engines running.
- [ ] Integration test: adding a 5th commodity adds **0** new sockets.
[...]

PR #1900, opened 2026-04-29
chore(orderref): flip use_orderref_filtering to true in DEV (#1670 Step 2.5)

rozavala:
## Summary

- Flip `connection_pool.use_orderref_filtering` from `false` to `true` in DEV `config.json`. This starts the 3-trading-day validation clock for the orderRef-based DISCONNECT_GUARD filter shipped in #1748 (Step 2 of #1670 Phase 2).
- Refresh CLAUDE.md's stale "currently Coffee Arabica / KC on NYBOT" line to reflect the actual 4 active commodities (KC, NG, SB, ZW; CC deactivated).

## What this enables

Per the sequencing in #1749:

> Step 2.5: DEV `use_orderref_filtering: true` flipped → 3 trading days of DEV logs with zero `filter divergence` warnings.

After this PR, the orderRef path becomes the primary `DISCONNECT_GUARD` filter. The clientId path is still computed in parallel and any disagreements are logged as `DISCONNECT GUARD [<purpose>]: filter divergence` (`connection_pool.py:~1052`). Three trading days with no divergence hits → flip the same flag in PROD config and run another week → Step 3 (the slot allocator) unblocks.

## What to watch for

```bash
grep "filter divergence" /home/rodrigo/real_options/logs/*.log
```

Any non-empty output = the orderRef filter would have drained a different set of orders than the clientId filter. That's the kind of disagreement we need to root-cause before letting orderRef be the only source of truth for the slot allocator.

## Risk

Low. The flag only changes which prefix the `DISCONNECT_GUARD` filter compares against — both paths are evaluated regardless. Worst-case orderRef-vs-clientId disagreement is logged, not acted on differently than today (the orderRef path's drain decision becomes authoritative when the flag is on, but the clientId-path drain set is still cross-checked).

## Why update CLAUDE.md

The "currently KC only" line caused me to underestimate the value of #1670's socket refactor by ~4x in conversation. The system is already at 4 commodities, where Phase 1's per-engine sockets count starts to matter — anyone reasoning about whether multi-commodity scaling work is worth doing should not be misled by the stale claim.

## Test plan

- [x] DEV pulls new config on next service reload.
- [ ] Watch DEV logs for 3 trading days: zero `filter divergence` warnings.
- [ ] If clean → open follow-up PR flipping the same flag in PROD; if not clean → root-cause before flipping.


rozavala:
## Review: APPROVE — DEV-only flag flip, well-tested toggle

**Verdict:** Approved. Tiny, reversible config flip; the flag's code path is fully tested in both states; Phase 1 + Step 2 (orderRef tagging + behind-flag filter) already landed.

### MECE

**Value (high):** Flips `connection_pool.use_orderref_filtering` from `false` → `true`, switching DISCONNECT_GUARD's authoritative filter from `clientId` to `orderRef`-prefix-by-ticker (`connection_pool.py:1011,1039`). Starts the 3-day DEV validation clock per #1749 sequencing — required to unblock the slot allocator (Step 3). The CLAUDE.md drift fix (KC-only → 4 active commodities) is a free correction worth shipping.

**Risk (low) — what could break in DEV with the flag flipped:**
- (a) DEV-only via branch deploy: `deploy.yml` deploys `main` → DEV droplet, `production` branch → PROD. PR targets `main`. PROD config.json on the `production` branch remains `false`. Verified.
- (b) Both filters still execute regardless of flag (`connection_pool.py:1021-1037`); divergence is logged, not silently swallowed (`filter divergence` warning, line 1051). Worst case if orderRef misses an order: drain skips it → cancel-on-disconnect deferred — but the clientId-only diagnostic surfaces it immediately.
- (c) Backward-compat: orders without `orderRef` (`""`) → `_order_ref_matches_ticker` returns False → safely skipped under flag-on; under flag-off they still match by clientId. CATASTROPHE_{ticker}- and CATASTROPHE_CLOSE_{localSym} prefixes both handled (lines 962-972).
- (d) Master pool keys (no ticker) deterministically fall back to clientId regardless of flag (line 1011, `ticker is not None` guard).
- (e) Rollback: revert the one-line config change. No schema, no state, no migrations.

**Prior art:** #1745 (Step 1 — ticker-prefixed orderRef tagging at all placeOrder sites, merged), #1748 (Step 2 — orderRef filter behind this flag, merged 2026-04-29), this PR (Step 2.5 — flip in DEV). Sequence is documented in #1670 and #1749. No prior failed flip.

**Trajectory fit:** Standard DEV → observe → PROD rollout. Observability is built in (the divergence-logging diagnostic is the gate criterion). After 3 clean DEV days, follow-up PR flips the same line on the `production` branch.

### Specific verification
- **DEV-only:** confirmed via `deploy.yml` branch routing (main→DEV, production→PROD). PR targets `main`.
- **Flag → code path:** `_use_orderref_filtering()` (line 976) reads `connection_pool.use_orderref_filtering`; consumed at line 1011 to gate `use_orderref` selection between `owned_by_orderref` and `owned_by_clientid` at line 1039.
- **Both-state coverage:** `tests/test_connection_pool.py:1347-1470` — 8 tests covering flag-off (clientId behavior preserved), flag-on (orderRef prefix), CATASTROPHE prefixes, master-key fallback, terminal-status skip, divergence logging, no-divergence-when-agree, master-key-no-divergence, and config-read.
- **Reconciliation backward-compat:** `order_manager.py` orderRef references (lines 160-196, 241, 276, 2754, 3693-3845) are independent of the disconnect-guard filter — orderRef is already used for position attribution and catastrophe-stop cleanup regardless of this flag.

### Test result
Focused suite: `pytest tests/ -k 'orderref or order_manager or reconciliation or connection_pool or ib_interface' -x -q --timeout=120` → **193 passed, 0 failed, 1897 deselected, 88s**. Scope justified: this is a one-line config flip whose only behavioral surface is the disconnect-guard drain logic — the focused selection covers every relevant module (`connection_pool`, `order_manager`, `reconciliation`, `ib_interface`) plus the dedicated orderref tests. Full suite would add ~10 min of tangential agent/eval tests with no incremental signal on this change.

### Non-routine question — does orderRef filtering actually work when enabled?
Yes. Phase 1 shipped ticker-prefixed orderRefs at every `placeOrder` site (#1745). Step 2 shipped the filter logic behind this flag (#1748) with the divergence-log safety net. The flag is the last lever; the underlying code is implemented and tested. The 3-day divergence watch is the empirical confirmation step.

PR #1901, opened 2026-04-29
feat(prompts): startup self-check for optimized-prompt loading

rozavala:
## Summary

**Defensive add on top of PR #1898.** That PR fixed the loader bug; this PR adds a startup-time self-check that would have surfaced the bug 2 days earlier and will catch the next "silent failure" of this class immediately at orchestrator startup.

## The problem this prevents

PR #1898's bug: shipped agronomist + geopolitical optimized prompts were silently inactive in production for ~2 days because the loader rejected them with a single low-visibility WARNING buried among thousands of other log lines:

```
WARNING - Optimized prompt /.../agronomist/prompt.json missing 'instruction' key, skipping
```

There was NO positive INFO log when prompts succeeded either, so the silent-failure state looked identical to "no toggle enabled" or "no prompt files present." The bug was only caught when I noticed the monitoring digest showing zero cohort data.

## What ships

`TradingCouncil.__init__` now calls `verify_optimized_prompts_at_startup()`, which probes every enabled-toggle ticker × specialist persona and emits one INFO summary line per ticker:

```
[startup-verify] KC: loaded=2 not_found=6 failed=0
  (loaded: ['agronomist', 'geopolitical'];
   not_found: ['inventory', 'macro', 'sentiment', ...];
   failed: none)
```

Plus a **loud WARNING when toggle is on but ZERO prompts load** — the exact production state PR #1898 fixed:

```
[startup-verify] KC: toggle ON but ZERO optimized prompts loaded —
  production will run on legacy personas. Check
  evals/_eval_optimized/{agent}/prompt.json files exist and pass validation.
```

## Per-(ticker, agent) outcomes recorded

- `LOADED ({mode})` — file present and accepted; mode is "instruction" / "supplement" / "{N} demos" / combination
- `SKIPPED` — toggle off for this ticker (no probe attempted)
- `NOT_FOUND` — toggle on but no prompt.json at either source path
- `FAILED: {exception}` — exception during load (rare; details logged)

Returns a summary dict for callers / future health-check exposure.

## Safety

The self-check is wrapped in try/except in `__init__` — failures NEVER block engine startup. "Observability must never kill trading" per the codebase principle. Wraps with `logger.warning` on any unexpected raise.

## Tests

`tests/test_optimized_prompt_startup_check.py` — 7 tests:

- ✅ no toggles enabled → empty result, single info line
- ✅ toggle off → SKIPPED for every persona, no warning
- ✅ demos-only prompts loaded successfully (regression for PR #1898)
- ✅ toggle on but zero loaded → loud WARNING fires
- ✅ partial load → no zero-loaded warning
- ✅ per-ticker INFO summary lines emitted
- ✅ loader exception → recorded as FAILED, never propagated

All 2097 tests pass.

## Why this is the right defense

PR #1898 added unit tests for the loader, which catches the specific bug class in CI. This PR adds a startup-time check that catches the BROADER class of "silent failure" bugs — including future ones that wouldn't be covered by the loader tests alone (e.g., file permissions changing, eval_optimized_dir misconfiguration, JSON file corruption between deploys, etc.).

After deploy, every orchestrator restart will emit the per-ticker summary INFO line, providing a positive signal that optimized prompts are actually loading — not just an absence of warnings.


rozavala:
**Verdict: Approve.** End-to-end self-check that catches the exact #1898 regression class; full suite green (2097 passed in 7m13s).

**Value (does it catch #1898?):** Yes — calls the same `_load_optimized_prompt` (agents.py:331) that production `research_topic` invokes (agents.py:1192), so the failure mode is identical. `test_verify_loaded_when_demos_only_prompts_present` is the explicit #1898 regression test (demos-only payload with empty `instruction`). Plus the toggle-on-zero-loaded WARNING line (`[startup-verify] KC: toggle ON but ZERO optimized prompts loaded`) would have screamed in deploy logs immediately, vs. the ~2-day silent gap. Coverage matches `_SPECIALIST_PERSONA_KEYS = (agronomist, geopolitical, macro, sentiment, technical, volatility, inventory, supply_chain)` — identical to the 8-agent `reflexion_agents` roster in config.json:91-98.

**Risk:** Failure semantics deliberately fail-soft (try/except in `__init__`, logger.warning on raise) — this is the right call here. It's a probe, not a precondition; observability must never kill trading (CLAUDE.md "fail closed" applies to trade execution, not telemetry hooks). The loud WARNING + per-ticker INFO summary make any silent regression visible at every restart. No state writes, no regime/vocabulary changes, no startup latency concern (8 personas × 1 ticker × small file reads = sub-second). False-positive risk is low — outcomes are filesystem + JSON validation; the LOADED-mode reporter (`instruction` / `supplement` / `N demos`) won't reject legitimate variants.

**Prior art:** Direct response to #1898 and the cohort-split-monitoring (#1847) gap — monitoring caught it after-the-fact, this catches it at deploy time. One small note: `verify_system_readiness.py` is the natural standalone home per CLAUDE.md, but `TradingCouncil.__init__` is also defensible because it's exercised on every orchestrator restart (every CI deploy). Acceptable as-is; could be referenced from `verify_system_readiness.py` later as a one-liner extension.

**Trajectory fit:** Generalizable pattern (returns a summary dict for "future health-check exposure" per the docstring). Reusable for DSPy demos / model registry / sentinel configs.

**Specific verification:**
- End-to-end vs file-exists: end-to-end (calls real `_load_optimized_prompt` → real `_read_optimized_prompt_file` validation, same path as production line 1192).
- All-agents coverage: 8/8 specialists; all `use_optimized_prompts.{ticker}` entries iterated (currently `KC: true` only).
- Failure semantics: fail-soft self-check + loud WARNING on zero-loaded; appropriate for an observability probe.
- Test coverage: 7 tests cover empty-toggles / toggle-off / demos-only-loaded (#1898 regression) / toggle-on-zero-loaded WARNING / partial-load-no-warning / per-ticker INFO summary / loader-exception-graceful.

**Tests:** `pytest tests/ -x -q --timeout=120` → 2097 passed in 433s (full suite, per `feedback_full_test_suite_for_execution_prs.md` since this touches `TradingCouncil.__init__`).

**Action:** Adding `Ready to be merged`.

PR #1907, opened 2026-04-29
fix(prompts): signal_generator writes optimized_prompts column + shared helper

rozavala:
## Summary

**Second production bug** discovered after PR #1898 + #1901 confirmed the loader fix was working. The `optimized_prompts` column was populated for emergency-path cycles (`orchestrator.py:4981`) but **empty string for scheduled-path cycles** (`signal_generator.py:825` `council_log_entry` was missing the field entirely).

Production data Apr 29 (after PR #1898 merge):
- 4 populated cycles (emergency path) ✓
- 10+ empty-string cycles (scheduled path) ✗

Scheduled cycles are the bulk of trading activity. So even though the loader fix made optimized prompts ACTUALLY load, the data wasn't being persisted on the most common code path — the monitoring digest's cohort split was still blind to scheduled cycles.

### 1. New shared helper `trading_bot.utils.build_optimized_prompts_meta(reports)`

Extracts per-agent optimized-prompt provenance into a dict suitable for the `optimized_prompts` column. Centralized to prevent the two writers from drifting again — PR #1898 surfaced one drift, this PR closes the second drift, the helper prevents the third.

### 2. `signal_generator.py` — adds the missing field

Calls the helper to build `optimized_prompts_meta`, adds it to `council_log_entry`. Mirrors the orchestrator emergency path exactly.

### 3. `orchestrator.py` — refactored to use the helper

Same behavior, no functional change. Eliminates the "two copies of the same logic" risk.

## Tests

`tests/test_build_optimized_prompts_meta.py` — 9 unit tests covering:
- empty/None reports → empty meta
- skips legacy/commodity_profile (default sources)
- extracts `dspy_optimized_demos_only` (the shipped pattern, regression test)
- extracts all `dspy_*` sources
- demo_count defaults to 0 when missing/None
- skips non-dict reports defensively
- skips empty/missing `prompt_source`
- coerces string `demo_count` to int

All 2106 tests pass.

## Production impact after merge

Scheduled KC cycles will populate the `optimized_prompts` column the same way emergency-path cycles do. The monitoring digest's cohort split (PRs #1843/#1847) will start getting scheduled-cycle data — the bulk of production trading activity.

## Validation methodology

- PR #1898 fixed the LOADER (made optimized prompts actually load)
- PR #1901 added STARTUP check (positive signal that prompts loaded)
- This PR fixes the LOG PERSISTENCE (scheduled cycles now record provenance)

The three together close the loop on "shipped prompts → loading → recorded → measurable in monitoring." Pre-PR-1898, the entire chain was broken silently.


rozavala:
## Approve — closes the cohort-split monitoring gap that hid PR #1898

**Verdict:** Ship. Helper output is byte-for-byte equivalent to the pre-PR emergency-path inline logic, the `signal_generator.py` change is a strict additive write to a dict that was missing the field, and full test suite passes (2106 passed).

**MECE review:**

1. **Value (high).** This is the second half of the regression that made #1898 silent for two days. PR #1898 fixed the loader (optimized prompts now actually load); this fixes log persistence on the scheduled path (`signal_generator.py:825` `council_log_entry` was missing `optimized_prompts` entirely → schema backfill defaulted to `"{}"` → cohort split saw all scheduled cycles as "no optim", inverting the comparison). After merge, scheduled-path KC cycles — the bulk of trading — populate the column the same way emergency-path cycles do, so `scripts/monitor_optimized_prompts.py` finally sees the cohort it's supposed to be measuring.

2. **Risk (low — though touches signal_generator).** **Helper format matches existing rows exactly.** Verified by reading the pre-PR emergency-path inline block (`orchestrator.py` main @4906–4926) and diffing against `build_optimized_prompts_meta`: same skip rule (`ps and ps not in ('legacy', 'commodity_profile')`), same output shape (`{key: {'source': ps, 'demo_count': int(...)}}`), same `int(... or 0)` coercion. Both writers emit `json.dumps(meta)` into the column — schema continuity preserved (`schema.py:103` default `"{}"` already matches helper's empty-dict output for cycles where no agent used an optimized prompt). The `signal_generator.py` addition is purely additive — adds a key that was missing — and the value is `json.dumps({})` in the no-optim case, so old downstream readers (digest, dashboard, backfill) already handle it. Fail-closed semantically correct: if no agent loaded optim prompts, the helper returns `{}`, which the digest correctly classifies as "without-optim cohort." No regime/confidence/persona naming touched.

3. **Prior art.** PR #1799 (v11 schema), #1843/#1847/#1849 (digest + alert), #1898 (loader fix), #1901 (startup self-check). This is the missing rung between #1898 (load) and #1847 (measure).

4. **Trajectory fit.** Layered monitoring: #1901 = deploy-time loading check, this PR = steady-state persistence into cohort split. Both layers needed; this completes the loop.

**Verification:**
- **Helper format:** byte-equivalent to pre-PR emergency-path inline logic (verified line-by-line).
- **Call site coverage:** orchestrator emergency path AND signal_generator scheduled path are the only two writers (`grep` across repo). Both now route through the helper. No third writer exists.
- **Fail-closed:** missing/non-dict reports → skipped (defensive); no `prompt_source` → skipped; default sources (`legacy`, `commodity_profile`) → skipped. Result is always a serializable dict, always JSON-dumped before write. Exception path in signal_generator is wrapped in the existing `try/except` that logs failures.
- **Cohort-split unblock:** consumer (`monitor_optimized_prompts.py:137,251,308`) reads `(row.get("optimized_prompts") or "").strip()` — handles both empty-string legacy rows and the new `"{}"` populated rows correctly.

**Tests:** 2106/2106 passed in 440s (full `pytest tests/ -x -q --timeout=120`). 9 new unit tests in `tests/test_build_optimized_prompts_meta.py` cover empty/None/non-dict reports, default-source skipping, all `dspy_*` source variants, missing/None `demo_count`, and string-int coercion.

**The non-routine question** — does the helper produce output in the EXACT format that existing rows have? Yes, verified line-by-line against the pre-PR emergency-path block. Same skip predicate, same dict shape, same int coercion, same `json.dumps` at the call site. No schema split.

rozavala:
**Claude Code review — LGTM, ready to merge.**

Clean fix for a real production blind spot. Code review summary:

- **`trading_bot/utils.py`**: `build_optimized_prompts_meta` helper is well-designed — filters defaults, handles `None`/non-dict/missing fields defensively, coerces `demo_count` to int. Centralized to prevent future drift.
- **`trading_bot/signal_generator.py`**: Scheduled cycles now write `"optimized_prompts"` to `council_log_entry`, exactly mirroring `orchestrator.py:4981`. This closes the monitoring blind spot where scheduled cycles (the majority of production activity) were writing empty string for the column.
- **`orchestrator.py`**: Refactored to use the helper — no functional change to the emergency path.
- **Tests**: 9 cases covering the helper's edge cases. Test run: 28/30 pass; 2 failures are `ModuleNotFoundError: No module named 'requests'` in `notifications.py` — pre-existing sandbox environment issue, unrelated to this branch.

No trading logic modified. Safe to merge.

---

PR #1910, opened 2026-04-29
fix(ledger): defensive read in load_and_prepare_ledger

rozavala:
## Summary

Wraps `pd.read_csv` in `load_and_prepare_ledger` (`trading_bot/order_manager.py`) with try/except so a single malformed CSV cannot take down `get_trade_ledger_df()` and cascade across every consumer (`close_stale_positions`, `emergency_hard_close`, `TradeReconciler`, `PerformanceAnalyzer`).

## Why

Today (2026-04-29 ~17:15-17:38 UTC) a single broken file did exactly that. KC's `data/KC/trade_ledger.csv` had a stale 11-col header (carried over from leaked test fixtures) plus 12-col rows from today's real fill at 14:09. pandas raised `ParserError`. With no exception handling around the read, the consolidator function blew up and four critical paths failed in lockstep:

- `close_stale_fallback` at 17:15 — couldn't close KC position
- `emergency_hard_close` at 17:25 — same error → "Emergency close FAILED entirely"
- `TradeReconciler` at 17:38 — same error → reconciliation completed but missed today's fill
- `PerformanceAnalyzer` at 17:41 — same error → analysis blocked

KC position was stuck open until the file was manually repaired.

The proximate cause (test pollution writing 11-col fixtures into a per-ticker production data dir) is being addressed separately. This PR is the **defensive layer** so any future schema drift, encoding error, truncation, or other CSV corruption is contained to a single file rather than cascading.

## Change

```python
try:
    df = pd.read_csv(file_path)
except (pd.errors.ParserError, pd.errors.EmptyDataError, UnicodeDecodeError) as e:
    logger.error(f"Skipping unreadable ledger file '{file_path}': {e}")
    return pd.DataFrame(columns=['timestamp', 'position_id'])
except Exception as e:
    logger.error(f"Unexpected error reading ledger file '{file_path}': {e}")
    return pd.DataFrame(columns=['timestamp', 'position_id'])
```

Bad file logs and returns empty; remaining files in `archive_ledger/` still consolidate.

## Tests

`tests/test_ledger_defensive_read.py` — 3 cases reproducing today's incident:

1. **Mixed 11-col / 12-col file** (today's exact failure mode — uniform mismatched files silently shift columns, only the mix triggers `ParserError`, so the test pattern needs to match production).
2. **Garbage archive file** (binary noise inside `archive_ledger/`) doesn't block other archives.
3. **Empty live ledger** doesn't raise `EmptyDataError`.

Plus 23 existing related tests still pass:

```
[...]

PR #1841, opened 2026-04-27
fix(connection-pool): cooldown-aware retry for recovery-cascade callers

rozavala:
## Summary
- Closes the cluster of CRITICAL post-disconnect-hold-off errors fired by the EOD RECOVERY queue: #1837 (umbrella + dups #1836/#1835/#1811/#1810/#1806), #1755 (`close_stale_positions FAILED`), #1756 (`emergency_hard_close FAILED`).
- Introduces a typed `ConnectionInCooldown(ConnectionError)` exception carrying `wait_remaining` + `cooldown_type` so recovery-cascade callers can sleep through the cooldown instead of `_force_reset_connection`-ing into a non-recoverable backoff cascade.
- Touches three callers that currently escalate cooldowns to CRITICAL where the existing VaR/sentinel paths quietly defer.

## Root cause (DEV log, 2026-04-26 19:23 UTC)

After an orchestrator restart, every engine logs `LATE START DETECTED: 19 scheduled tasks already passed` and the recovery queue fires the missed schedule back-to-back. On `{ticker}_orders` (or `{ticker}_reader`), the first task's `release_connection(...)` stamps `_last_clean_disconnect`, opening a 60s hold-off window (`RECONNECT_HOLDOFF_AFTER_CLEAN_DISCONNECT = 60`). The next task lands inside the window:

```
16:55:46 RECOVERY: close_stale_primary    → SB_orders connects, succeeds, releases (hold-off stamped)
16:55:56 RECOVERY: close_stale_fallback   → SB_orders in post-disconnect hold-off, 54s remaining
16:55:56 attempt 1/3: hold-off → _force_reset_connection (stamps 60s BACKOFF floor)
16:56:01 attempt 2/3: backoff period → _force_reset_connection again
16:56:11 CRITICAL: close_stale_positions FAILED after 3 attempts
16:56:17 RECOVERY: emergency_hard_close → CRITICAL: backoff period (no retry)
```

The retry budget is `RETRY_DELAYS = [5, 10, 20]` = 35s — never enough to outlast a fresh 60s backoff floor.

## Fix

1. **`ConnectionInCooldown(ConnectionError)`** in `trading_bot/connection_pool.py`. Both the hold-off and backoff branches of `get_connection` now raise this typed subclass with the exact `wait_remaining` and a `cooldown_type` of `'hold-off'` or `'backoff'`. Existing `except ConnectionError:` callers still catch it; existing tests that match on the message string still match.

[...]

Issue #1837, opened 2026-04-26
[CRITICAL] uncategorized: An error occurred during trade reconciliation: Connection ZW_reader in post-d...


rozavala:
## Error Report

| Field | Value |
|-------|-------|
| **Category** | `uncategorized` |
| **Severity** | CRITICAL |
| **Environment** | DEV |
| **First seen** | 2026-04-26 19:23:29,829 |
| **Last seen** | 2026-04-26 19:23:29,829 |
| **Occurrences** | 1 |

### Sample Log Entry

```
An error occurred during trade reconciliation: Connection ZW_reader in post-disconnect hold-off
Traceback (most recent call last):
  File "/home/<USER>/real_options/orchestrator.py", line 3663, in reconcile_and_notify
    recon_ib = await IBConnectionPool.get_connection("reconciliation", config)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/<USER>/real_options/trading_bot/connection_pool.py", line 341, in get_connection
    raise ConnectionError(
ConnectionError: Connection ZW_reader in post-disconnect hold-off
```

### Impact

Investigate to determine impact on trading operations.

> The `claude-fix` label will be added after triage — Claude will automatically attempt a fix.

---
*Automated report by `scripts/error_reporter.py` — fingerprint: `b1250fce440d`*

rozavala:
## Umbrella tracker — recovery-queue + post-disconnect hold-off cascade

This is the keep-open tracker for the cluster: also #1836 `NG_reader`, #1835 `KC_reader` (2026-04-26 19:23, all closed as dups), plus the older 2026-04-25 reports #1811 / #1810 / #1806.

### Root cause (DEV log, 2026-04-26 19:22–19:23 UTC)

Orchestrator restarted at ~19:22:26, triggering `LATE START DETECTED: 19 scheduled tasks already passed` on each engine. The RECOVERY queue runs missed tasks back-to-back. On each `{ticker}_reader` pool key, an early recovery task (e.g. sentinel/audit) acquires + releases with the default `holdoff=True`, stamping `_last_clean_disconnect`. The window is `RECONNECT_HOLDOFF_AFTER_CLEAN_DISCONNECT = 60s` (`connection_pool.py:160`).

[...]

PR #1912, opened 2026-04-30
fix(reconciler): clamp cutoff when partial Flex Query failure

rozavala:
## Summary

When one of multiple `FLEX_QUERY_ID`s configured for trade reconciliation returns no data (today: query 1328326 returned IBKR Code 1001 "Statement could not be generated at this time" all day, while 1331422 kept succeeding), the reconciler used to concatenate whatever returned and run the comparison anyway. The surviving narrow query covered a smaller time window than the local ledger, so legitimate prior-day local trades looked "superfluous" — false-positive Pushover alerts for 12 (then 122) trades on 2026-04-29.

This PR: track which queries returned no data; if any failed AND we still have IB data, advance the comparison cutoff to the oldest IB trade timestamp in the surviving data. Restricts the comparison to the window where IB coverage is complete and suppresses the false positives.

Mirrors the existing `ACCOUNT_RESET_DATE` clamp pattern, but driven by runtime query failures rather than a configured floor.

## What changed

`reconcile_trades.py:main()`:

1. Track `failed_query_ids` during the fetch loop.
2. After timestamp normalization, if any queries failed AND `ib_trades_df` is non-empty, advance `cutoff_date` to `max(cutoff_date, min(ib_trades_df['timestamp']))`.
3. Log the clamp at WARNING with the failed query IDs and the new cutoff.

When ALL queries fail, existing behaviour holds — no comparison runs (the early-return at *"No trade data was fetched from any IB Flex Query. Exiting."* is unchanged).

## Test plan

`tests/test_reconcile_partial_flex_failure.py` — 3 cases:

| Test | Scenario | Assertion |
|---|---|---|
| `test_partial_flex_failure_clamps_cutoff` | 1 of 2 queries fails; surviving query covers only today; local has prior-day + today's trades | 0 superfluous (prior-day clamped out) |
| `test_all_queries_succeed_no_clamp` | Both queries return identical today's data | 0 missing, 0 superfluous (existing behaviour preserved) |
| `test_all_queries_fail_returns_empty` | Both queries return None | empty dfs (existing early-return preserved) |

Plus 32 existing related tests still pass:

```
tests/test_reconcile_partial_flex_failure.py    (new, 3 tests)
tests/test_reconciliation.py
tests/test_resolve_original_position_ids.py
tests/test_phantom_ledger_reconciliation.py
tests/test_emergency_close_ledger_continuity.py
35 passed
```

## Related

- Issue #1911 — tracks the upstream IBKR-side cause of query 1328326's continuing 1001 failures (recreate query in IBKR Account Management if it persists tomorrow).
- PR #1910 — defensive read in `load_and_prepare_ledger` (different defense, same incident).


rozavala:
## Verdict: Approve — Ready to be merged

Production-incident fix for the 2026-04-29 false-positive cascade (12 → 122 superfluous-trade Pushovers when Flex Query 1328326 returned IBKR Code 1001 all day while 1331422 kept succeeding). Clamp lives at the right layer (after timestamp normalization, before the `ib_trades_df`/`local_trades_df` cutoff filter), mirrors the existing `ACCOUNT_RESET_DATE` v3.2 clamp pattern, and is gated on `failed_query_ids and not ib_trades_df.empty` so it cannot fire on the all-success path.

### MECE

**Value.** Real incident with Pushover alert receipts and a tracker (#1911) for the upstream IBKR-side cause. Without the clamp, surviving narrow queries inflate "superfluous" counts whenever any configured query returns no data — the false-positive flood blocks operator attention to genuine reconciliation drift.

**Risk.** All four risk surfaces clear:
- *Backward-compat:* clamp is a no-op when `failed_query_ids` is empty (success path) and a no-op when `oldest_ib_ts <= cutoff_date` (clamp would not advance). `test_all_queries_succeed_no_clamp` proves this.
- *Brier continuity:* fix is forward-only — narrows comparison window, does not modify `trade_ledger.csv` or backfill anything. No upgrade-time discontinuity. Schema unchanged.
- *State atomicity:* cutoff is computed in-memory per run; not persisted. State writes via `state_manager.py` not touched.
- *Idempotency:* re-running on the same partial-failure window yields the same clamped cutoff (deterministic from `ib_trades_df.min()`), so no duplicate trades. *Failure mode the operator should know:* during the unrecovered window, prior-day local trades are excluded from comparison — so a *genuine* superfluous trade older than the surviving window will be silently invisible until the failing query recovers. Acceptable trade-off vs. the alert flood, and the WARNING log at line 996 documents it.

**Prior art.** Cutoff has been touched before (#1525 `ACCOUNT_RESET_DATE` floor — same pattern, same site, configured-floor sibling) and reconciliation has churned (#1684 OPEN-pid resolver, #1334 cross-thesis netting). Distinct from #1837 / #1841 (hold-off retry, already closed). Not duplicating prior work.

**Trajectory fit.** Complements PR #1910 (defensive ledger read — orthogonal defense, same incident) and is referenced in #1911 (IBKR-side tracker). Doesn't interact with orderRef filtering work.

### Specific verification

- *Diff scope:* only `reconcile_trades.py` (+26) and a new test file (+204). Confirmed no execution-layer modules (`order_manager.py`, `ib_interface.py`, `compliance.py`) touched.
- *Cutoff clamp logic:* `oldest_ib_ts = ib_trades_df['timestamp'].min()` — pinned to the **earliest timestamp in the surviving IB data**, which is exactly the right floor. Not "before now," not arbitrary. The `if oldest_ib_ts > cutoff_date` guard prevents the clamp from ever moving the cutoff backward.
- *Test coverage:* three branches — (1) `test_partial_flex_failure_clamps_cutoff` reproduces the exact 2026-04-29 incident shape (1328326 returns None, 1331422 returns today's KC, local has Apr 21 + Apr 29 trades, expect 0 superfluous); (2) `test_all_queries_succeed_no_clamp` proves the success path is preserved; (3) `test_all_queries_fail_returns_empty` proves the all-fail early-return is preserved. The prior-bug scenario IS covered.

### Tests

Worktree: `/home/rodrigo/real_options/.claude/worktrees/agent-ad296a10bbb2798a3` (inside `real_options/`, so I used `--basetemp=/tmp/wt-1912-tmp` to keep side effects out of the live DEV `data/`/`logs/`/`state.json`). Verified no writes hit those paths.

- Full suite: `pytest tests/ -x -q --timeout=120 --basetemp=/tmp/wt-1912-tmp` → **2100 passed in 6m45s**, 9 pre-existing warnings (unrelated `coroutine never awaited` in `order_manager` mocks).
- Filtered first pass: `-k 'reconcil or flex or ledger or trade or brier'` → 132 passed in 80s including the 3 new partial-failure tests.

Per `feedback_full_test_suite_for_execution_prs.md`: full suite run, this is reconciliation/Brier-adjacent.

Action: `Ready to be merged`.

PR #1913, opened 2026-04-30
fix(catastrophe-writer): include trend_alignment in row dict

rozavala:
## Summary

`_log_catastrophe_fill_to_ledger`'s row dict had 11 keys (no `trend_alignment`). If a catastrophe stop fill is the first writer to a missing/empty `trade_ledger.csv`, the resulting file gets an 11-col header. A subsequent regular fill writes 12-col rows (schema v10) on top — `pd.read_csv` then raises `ParserError` and every ledger consumer fails in cascade.

That's the same failure shape as the 2026-04-29 KC incident, just sourced from production code rather than the test pollution that actually triggered it on Apr 29. Latent (no production occurrences in logs to date) but real.

PR #1910 added the consumer-side defense (try/except around `pd.read_csv`); this PR adds the producer-side schema match. Both layers complement each other.

## Change

Two-line addition to the catastrophe row dict:

```python
row = {
    ...
    'reason': 'CATASTROPHE_FILL (auto-detected)',
    'trend_alignment': '',
}
```

The catastrophe writer's existing logic already does the right thing when the ledger *exists* (reads the existing header and matches the row to it). The bug is only when the catastrophe writer is the *first* writer — the `else` branch builds `fieldnames` from the row dict's keys, so missing keys → missing columns.

## Test plan

`tests/test_catastrophe_writer_schema.py` — 2 cases:

| Test | Asserts |
|---|---|
| `test_catastrophe_fill_creates_12col_header_when_no_ledger_exists` | First catastrophe write produces a 12-col header containing `trend_alignment`. |
| `test_subsequent_regular_write_does_not_break_pandas_read` | Catastrophe write followed by a regular schema-v10 row appends cleanly; `pd.read_csv` parses both rows without raising. |

Plus 20 existing `test_catastrophe_stop_remediation.py` tests still pass:

```
tests/test_catastrophe_writer_schema.py        (new, 2 tests)
tests/test_catastrophe_stop_remediation.py
22 passed
```

## Related

- PR #1910 — defensive read (consumer-side)
- PR #1912 — reconciler partial-query clamp (different defense, same incident)
- Issue #1911 — IBKR-side Flex Query 1328326 monitoring


rozavala:
**Verdict: Approve.** Surgical 2-line fix to producer-side schema; complements PR #1910's consumer-side defense.

**Value:** Catastrophe writer's row dict had 11 keys, missing `trend_alignment`. When it was the FIRST writer to a missing/empty `trade_ledger.csv`, the `else` branch at line 650-652 built `fieldnames` from `row.keys()` → 11-col header. Subsequent regular writes (12-col schema v10) would append on top → `pd.read_csv` ParserError → cascading consumer failures (close_stale_positions, TradeReconciler, PerformanceAnalyzer). Fix adds `'trend_alignment': ''` so first-writer header matches schema v10.

**Risk (load-bearing schema column — verified):**
- **Source-of-trend_alignment:** empty string `''`, NOT a recomputation. This is semantically correct: catastrophe fills are terminal events written from `on_exec_details` where only fill-level data (`detected_fill`) is in scope, not the lot's stored alignment. Mirrors how `log_trade_to_ledger` defaults `trend_alignment or ''` at utils.py:809/824 when the caller passes None.
- **Backward-compat:** consumer at `_classify_lot_for_close` line 4023 — `str(trade_row.get('trend_alignment', '') or '').upper()` — handles empty string by falling into the UNKNOWN branch (`max_holding_days` default). Legacy 11-col rows handled identically. Catastrophe lots are already closed, so holding-period classification is moot.
- **Brier orphan window (336h):** unaffected — `enhanced_brier.py` and `brier_scoring.py` don't reference `trend_alignment` directly.
- **CLAUDE.md vocabulary:** empty string is the existing UNKNOWN-fallback convention used by `log_trade_to_ledger`. No new sentinel value introduced.
- **Schema column order:** row dict ordering (timestamp → trend_alignment) exactly matches `utils.log_trade_to_ledger`'s fieldnames at utils.py:647-650.
- **Atomic writes:** writer holds `TRADE_LEDGER_LOCK` (line 637) — preserved.

**Prior art:** Same incident-class as PR #1907 (signal_generator missing column) and #1901 (startup self-check). Schema-divergence between writers is a recurring failure mode; a producer-side helper that derives schema from a single fieldnames source could prevent the next instance, but that's scope creep here.

**Trajectory fit:** Closes the 2026-04-29 KC ParserError incident at the producer side; defense-in-depth with PR #1910 (consumer) and PR #1912 (reconciler clamp). Same incident, three angles — all justified.

**Tests:** 235 passed, 0 failed, scope `catastrophe or order_manager or trend_alignment or trade_ledger or brier or schema or close_stale or reconcil`. New file `tests/test_catastrophe_writer_schema.py` directly verifies (a) 12-col header on first write and (b) pandas can parse a catastrophe-then-regular ledger. Worktree: `/home/rodrigo/real_options-claude` (review repo, not under `/home/rodrigo/real_options/`); `--basetemp=/tmp/wt-1913-tmp`. Test plan would be stronger with WITH_TREND/COUNTER_TREND/UNKNOWN coverage of the downstream consumer over a catastrophe-written row, but that's `_classify_lot_for_close`'s domain (already covered elsewhere) and a catastrophe row's alignment is always empty by design — so the missing test wouldn't add signal.

**Action:** Labeled `Ready to be merged`.

PR #1977, opened 2026-05-05
fix(price-sentinel): inner 12s timeout on reqHistoricalDataAsync (#1975)

rozavala:
## Summary

Fixes \`PriceSentinel TIMED OUT after 30s (backing off 120s)\` on ZW (and occasionally NG) at session-boundary hours (~04:21–04:27 UTC ≈ 00:21–00:27 ET). At that time the contract is in the overnight Globex session — RTH ended hours ago — and IB's historical-data subsystem can hang past 30s on a \`useRTH=True\` 1-hour bar request.

Wrapping each \`reqHistoricalDataAsync\` call in \`asyncio.wait_for(timeout=12)\` lets the sentinel fail fast and try again next minute, instead of cascading into the orchestrator's 30s outer timeout + 120s backoff (which effectively takes the detector offline for an hour).

## Why this size

The fix is minimal on purpose. Other paths considered and not taken:

- **Switching \`useRTH=True\` → \`useRTH=False\`**: would change data semantics and affect KC/SB. The hang is the actionable problem; semantic changes can ride a separate PR.
- **Skip PriceSentinel entirely during PASSIVE state**: loses the deferred-trigger pipeline (PASSIVE EMERGENCY in \`orchestrator.py:5817\`), which is exactly when these sentinels matter for ZW.
- **Reduce orchestrator outer timeout**: doesn't fix the call hanging; just trims the wait.

## Files

- \`trading_bot/sentinels.py\` — wraps both calls (1-hour intraday + cumulative daily) in \`asyncio.wait_for(timeout=12)\`. Existing \`except Exception\` handlers absorb the resulting TimeoutError → sentinel returns None.
- \`tests/test_sentinels.py\` — two new tests:
  - \`test_price_sentinel_inner_timeout_returns_none\`: TimeoutError from inner wait_for must not propagate.
  - \`test_price_sentinel_inner_timeout_uses_real_wait_for_budget\`: patches \`wait_for\` to a 0.2s budget and confirms a hung mock IB call short-circuits via the production code path.

## Test plan

- [x] Targeted: 2 new tests pass; 3 existing PriceSentinel tests still pass (no behavioral change for non-hung path).
- [x] Full suite: \`pytest tests/\` — 2253 passed, 1 skipped, 0 failures.
- [ ] Operational: monitor \`PriceSentinel TIMED OUT after 30s\` log line over the next 3-5 days. With the fix, that line should disappear; instead the existing \`Price Sentinel check failed: TimeoutError\` ERROR-line will appear (logged at line 418 / 469 of sentinels.py) but without triggering the 120s backoff. ZW \`sentinel_stats.json\` errors:alerts ratio should improve as the next-minute retry has a chance.

Closes #1975


rozavala:
Approved.

**Fail-closed posture preserved**: `asyncio.wait_for(..., timeout=12)` raises `TimeoutError`, the existing `except Exception` handlers in PriceSentinel.check absorb it and return `None`. Orchestrator interprets `None` as "no trigger this cycle, retry next minute" — exactly the right behavior. Faster recovery than the prior 30s outer timeout + 120s backoff, no missed alerts beyond what was already silently dropped at midnight ET.

**Test coverage is solid**:
- `test_price_sentinel_inner_timeout_returns_none` mocks `reqHistoricalDataAsync` to raise `TimeoutError` directly, asserts `trigger is None` and `elapsed < 1s` (no propagation, no zombie wait).
- `test_price_sentinel_inner_timeout_uses_real_wait_for_budget` patches `asyncio.wait_for` to a 0.2s budget and exercises the actual production wrapper with a hung mock — the better belt-and-suspenders test, confirms the wrapper itself fires.
- Existing happy-path tests still pass (no behavioral change for non-hung path).

**Timeout choice**: 12s leaves ~18s of the orchestrator's 30s outer budget for the rest of the sentinel loop. Typical historical-data RTT during market hours is <2s, so 12s won't cause spurious failures during normal slow responses but will short-circuit the multi-minute hangs observed at session boundaries on ZW/NG.

**Resource cleanup**: ib_insync's `reqHistoricalDataAsync` returns a coroutine whose cancellation propagates to the library's internal request bookkeeping — explicit `cancelHistoricalData` is not required at this fail rate (4 per 4 days for ZW well under the 50-concurrent-request IB cap). If timeout rates ever climb above ~10/day, worth revisiting.

**Scope discipline**: PR description correctly considers and rejects three larger alternatives (useRTH semantic change, skip during PASSIVE, reduce outer timeout) for sound reasons. The 12-LOC change targets the actionable problem (the hang) without changing data semantics or affecting other commodities. Closes #1975 directly; should also chip into #1920 (ZW cycle volume gap).

PR #1971, opened 2026-05-05
chore(evals): refresh Tier-3 cached baselines after #1963 + #1967

rozavala:
## Summary

Captures the **7-failure → 0** win from the case-authoring triage in #1963 and #1967, and picks up `llm_mode: \"cached\"` provenance (added by #1960) on every baseline.

| Subtype | Before | After | Δ | Failures cleared |
|---|---|---|---|---|
| compliance | 24/25 (96.0%) | **25/25 (100%)** | +1 | `categorize_as` |
| devils_advocate | 20/20 (100%) | 20/20 (100%) | — | (already clean) |
| master | 27/30 (90.0%) | **30/30 (100%)** | +3 | `must_not_cite`, `direction_in`, `override_must_reference_tokens` |
| permabear | 30/30 (100%) | 30/30 (100%) | — | (already clean) |
| permabull | 28/31 (90.3%) | **31/31 (100%)** | +3 | `must_acknowledge_opposing`, `forbid_fabricated_quotes`×2 |
| **total** | **129/136 (94.85%)** | **136/136 (100%)** | **+7** | |

Direction distribution on master shifted BULLISH 16→15 / BEARISH 8→9 — that's the `override_consensus.001_sb` case-authoring correction in #1967, where the strengthened Indian-quota policy-shock anchor now genuinely warrants a BEARISH override.

## How it was generated

```
python -m evals.runner --mode cached --tier3-only --baselined-only \
    --baselines-dir evals/baselines/cached --snapshot
```

All 136 cases hit cache (no live LLM calls, zero API spend). The case-authoring fixes operate on assertion-config and metadata that don't participate in the LLM cache key, so the existing cached real-LLM responses are reused; only the assertion logic now correctly evaluates them.

## Test plan

- [x] `python -m evals.runner --mode cached --tier3-only --baselined-only --baselines-dir evals/baselines/cached` (no `--snapshot`) — confirmed 136/136 pass before snapshotting
- [x] Reviewed JSON diffs on all 5 baseline files — only `snapshot_at`, `passed`, `pass_rate`, `failures_by_check`, `direction_distribution` (master only), and the new `llm_mode` field changed
- [ ] CI eval gate green on this PR — the cached-mode gate now compares 100% against 100% on every subtype

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1979, opened 2026-05-05
chore(evals): refresh Tier-2 cached baselines with llm_mode provenance

rozavala:
## Summary

Refreshes the 8 Tier-2 cached baselines (agronomist, geopolitical, inventory, macro, sentiment, supply_chain, technical, volatility) from clean cron run [25353970540](https://github.com/rozavala/real_options/actions/runs/25353970540) — the first working cached-mode cron after the #1970/#1974/#1976 infra fixes.

## Why now

Every Tier-2 agent diff in #1973 surfaced:

> [WARN] Baseline pre-dates llm_mode tracking; can't verify it was snapshotted under the same mode as this run (cached). Refresh baseline with --snapshot to record provenance.

Existing baselines are from **2026-04-26 / 2026-04-28**, *before* the provenance schema landed (PR #1958 fix). Without `llm_mode` recorded, a `--snapshot --mode mock` could silently land synthesizer 100%-pass-rate metrics into `evals/baselines/cached/` — exactly the inflated-baseline failure mode #1958 documented. Refreshing now under `cached` mode kills the WARN noise across all 8 agents and re-anchors the drift signal so next week's cron produces genuine model-side drift instead of mock-vs-cached comparison artifacts.

## Source of truth

Built directly from `evals/reports/drift-weekly/results.json` of run 25353970540 (50-min cold-cache run on `bb837b4ae`, after the timeout bump). Every baseline file carries `"llm_mode": "cached"` and the run's `generated_at` as `snapshot_at`.

Format-identical to the existing schema (`schema_version: "1.0"`, same `metrics` keys); just refreshed values + provenance.

## What's NOT in this PR

- **Tier-3 baselines** — already refreshed yesterday in 61c08e200 under the llm_mode-aware schema. Run 25353970540 confirms them clean (green/yellow only, no WARN noise).
- **Investigation of geopolitical hallucination spike (0% → 15.6%)** and **volatility distribution shift** (BEARISH +14.6pp, NEUTRAL -19.5pp). These persist post-refresh and *become* the new normal in these baselines. Whether they reflect real model regressions worth swapping override models is a **separate investigation issue** (coming next).
- **Triage of the 62 case-level failures** (mostly canonical_neutral `confidence_max` over-strict, contrast direction flips, hallucination/`must_not_cite` false positives) — deferred per #1973 comment to a separate triage pass.

## Test plan

- [ ] After merge, dispatch `Eval Drift Weekly` once. Expected: cache hits everywhere (~5 min runtime), no `Baseline pre-dates llm_mode tracking` WARNs, Tier-2 worst severity drops from red to green/yellow on actual distribution movement only.


rozavala:
Approved.

**Provenance verified end-to-end:**
- Source run 25353970540: confirmed `workflow_dispatch` of Eval Drift Weekly, ran 02:04:17Z–02:54:21Z (50 min — matches "cold-cache" claim), headSha `bb837b4ae` = #1976. The run's `conclusion: failure` is the drift gate firing on real distribution movement vs 9-day-old pre-llm_mode-schema baselines — not an infra failure. Cache saved + `results.json` produced, which is the source-of-truth for this PR.
- Commit 61c08e200 = #1971 Tier-3 baseline refresh — Tier-3 correctly excluded from this PR (already llm_mode-aware).
- The infra-unblock chain (#1970 → #1974 → #1976) culminates here: this is the first cron-driven baseline refresh after the chain.

**Schema integrity**: all 8 files keep `schema_version: "1.0"` and the same metrics keys; only `snapshot_at` updates and `llm_mode: "cached"` is added. Trailing-newline cleanup is a free bonus.

**Numeric movements internally consistent**: geopolitical total 40 unchanged with passed 34→28 (6 more failures) aligns with hallucination_rate jumping 0.0→0.156. Sentiment total 34→39 and supply_chain 32→37 reflect newly-added cases from prior case-authoring PRs.

**One follow-up to make sure happens** (not blocking — explicitly deferred in PR description, but worth pinning):

The geopolitical `hallucination_rate: 0.0 → 0.15625` is a meaningful signal for that agent specifically, and baking it into the new baseline means next week's cron won't flag it as drift. Same for volatility's distribution shift (BEARISH +6, NEUTRAL -8). The sequencing choice (unblock drift detection first, investigate movements separately) is correct, but please open the investigation issue **before** the next cron runs — otherwise, if the spike is a real model-side regression (e.g. provider quietly retuned Flash, the exact failure mode this cron exists to catch), the alarm is now silenced.

Suggested issue scope:
1. **geopolitical**: 6/40 cases newly hallucinating — which 6? (results.json has per-case detail). Are they the same case_kind, same commodity, same prompt template?
2. **volatility**: BEARISH/NEUTRAL ratio swung 0.34/0.37 → 0.44/0.17. Is this case-set composition (new cases added) or agent behavior change at constant case set? (compare per-case directions across the two runs).

CI green (Eval Gate mock SUCCESS, Tier-3 active SUCCESS, cached SKIPPED — expected since no prompt changes). Closes the WARN noise across all 8 Tier-2 agents and re-anchors next week's drift signal.

PR #2028, opened 2026-05-06
fix(reconcile): use per-commodity profile for IBKR symbol prefixes (#2022)

rozavala:
## Summary

Closes #2022. ZW reconciliation was treating every ZW trade as superfluous and every ZW position as orphan because three hardcoded copies of `_IBKR_SYMBOL_PREFIXES` drifted from the source of truth in `config/profiles/`. Today's logs:

```
Filtered IB trades by commodity ZW (prefixes ('ZW',)): 16 -> 0
Successfully wrote 2 superfluous local trade(s) to 'data/ZW/archive_ledger/superfluous_local_trades.csv'
```

The 2 trades were the legs of a real bull call spread on `OZWZ6` (IBKR's options-symbol form for wheat options). The prefix filter looked for `"ZW"`, didn't match `"OZW"`, dumped everything to the superfluous file. **The forensic record was wrong; IB had the trades correctly.**

## Root cause

`config/commodity_profiles.get_ibkr_symbols_for_commodity('ZW')` already returns `('ZW', 'OZW')` correctly — that helper sources from `config/profiles/zw.json`. But three call sites kept their own hardcoded mapping dict instead of using the helper:

- `reconcile_trades.py:160` (active position filter)
- `reconcile_trades.py:937` (trade reconciliation filter)
- `dashboard_utils.py:451` (untracked-position filter)

All three had `{KC, CC, SB, NG}` and silently omitted ZW. The fallback `.get(ticker, (ticker,))` returned `("ZW",)` — a "valid" tuple that just doesn't match anything. Silent failure mode.

This matches the recurring transcription-drift anti-pattern (`feedback_review_transcription_drift.md`) — five prior PRs traced to parallel literal blocks silently diverging.

## Fix

Replace each of the 3 hardcoded dicts with `get_ibkr_symbols_for_commodity(ticker)`. The helper reads from the profile JSON, so adding a 6th commodity now auto-propagates to all 3 use sites without manual dict edits.

| Site | Lines removed | Lines added |
|---|---|---|
| `reconcile_trades.py:160` | 5 | 1 import + 1 call |
| `reconcile_trades.py:937` | 5 | 1 import + 1 call |
| `dashboard_utils.py:451` | 6 | 1 import + 1 call |

Net: -23 lines, +17 lines. Functional behavior preserved (same `(ticker,)` fallback for unknown tickers via the helper's `try/except`).

## Regression tests (4 new)

`tests/test_ibkr_symbol_prefixes.py`:

1. **`test_zw_resolves_to_zw_and_ozw_prefixes`** — pins the exact bug (ZW must include 'OZW'). Direct regression guard.
2. **`test_every_active_commodity_has_futures_and_options_prefixes`** — each option-trading commodity must have ≥2 prefixes. Catches the general shape: someone adds a 7th commodity and forgets the options-symbol mapping → test fails immediately.
3. **`test_all_listed_profiles_resolve_via_helper`** — every `list_available_profiles()` entry must successfully resolve through the helper. Catches malformed profiles.
4. **`test_no_hardcoded_ibkr_prefix_dict_outside_helper`** — fails if anyone re-introduces a hardcoded `_IBKR_SYMBOL_PREFIXES` literal in `reconcile_trades.py` or `dashboard_utils.py`. **Closes the drift surface entirely.**

## Tests

```
$ pytest tests/test_ibkr_symbol_prefixes.py tests/test_reconciliation.py
  tests/test_reconcile_holdoff_retry.py tests/test_reconcile_partial_flex_failure.py
20 passed in 7.89s

$ pytest tests/ --ignore=tests/test_eval_*.py
2342 passed, 1 skipped, 11 warnings
in 464.68s (7m44s)
```

Per memory `feedback_full_test_suite_for_execution_prs.md` — `reconcile_trades.py` is production code, full suite required.

## Production impact

After merge, the next ZW reconciliation cycle will correctly identify all ZW positions and trades. No data loss expected — the trades currently in `superfluous_local_trades.csv` are also in the live `trade_ledger_*.csv`, so the system has the truth; the "superfluous" file is a misleading historical artifact that can be deleted post-merge as a one-shot cleanup.

## MECE

- **Value**: closes the SBL forensic gap that mislabels every ZW trade as superfluous, eliminates the recurring transcription-drift surface in this code path, and adds a regression guard that prevents the next variant.
- **Risk**: production-touching change to reconciliation code. Mitigated by the helper's identical fallback semantics, the 4 regression tests, and the full pytest pass.
- **Prior art**: same shape of fix as the centralization pattern recommended in `feedback_review_transcription_drift.md`; same shape of bug as #1984/1987/1996/2010 row-builder drift.

Closes #2022.


rozavala:
Re-review: fix verified. Removing `claude-fix` label.

**Both flagged items now closed:**

1. ✅ **`pages/1_Cockpit.py:170` (the missing 4th site)**: dict literal replaced with the helper call:
   ```python
   from config.commodity_profiles import get_ibkr_symbols_for_commodity
   prefixes = get_ibkr_symbols_for_commodity(ticker) if ticker else None
   ```
   Identical pattern to the 3 sites in the original PR. ✓

2. ✅ **Regression test scope widened**: `sources_to_check` now globs `pages/*.py` recursively:
   ```python
   sources_to_check = [
       repo_root / "reconcile_trades.py",
       repo_root / "dashboard_utils.py",
       *sorted((repo_root / "pages").glob("*.py")),
   ]
   ```
   Any future hardcoded `_IBKR_SYMBOL_PREFIXES` literal anywhere in dashboard pages now fails the test, regardless of which page it lands in. **The drift surface is actually closed now**, not just "3 of 4 known copies replaced".

**Commit**: `6392c1ff1` — non-trivial fix (real diff to `pages/1_Cockpit.py` + test scope expansion), not a no-op rebuild. Title "fix(cockpit + test): close 4th hardcoded prefix copy + widen test scope" matches the diff.

**PR description updated**: now correctly says "4 copies total — the 4th in the Cockpit page was missed by the original PR and only caught on review". Honest framing of the iteration.

This is now the load-bearing version of itself — the regression test would have caught the original ZW miss, AND the wider scope means a 5th hardcoded copy anywhere in `pages/` (Scorecard, Council, Financials, Utilities, etc.) would also fail the test on introduction. Closes #2022 cleanly. Good to merge.

PR #2067, opened 2026-05-09
refactor: extract lenient LLM-JSON parser to shared module (#2058)

rozavala:
## Summary

Closes #2058. Extracts the generic markdown-fence + bracket-search lenient JSON parser shared between \`compliance.py:from_llm_response\` (Layers 1+1.5) and \`var_calculator.py:_robust_parse_llm_json\` into a new \`trading_bot/llm_json.py\` module.

## Why

PR #2055 review flagged the duplication as a transcription-drift trap (#2058 explicitly): future edits to one parser would silently diverge from the other. This PR makes the recovery primitive a single source of truth.

## What stays where

- **Shared (\`trading_bot/llm_json.py\`):** fence stripping + strict \`json.loads\` → bracket-search \`raw_decode\` recovery.
- **Compliance-only:** dataclass shape validation, regex Layers 2/3 matching \`approved\`/\`reason\` fields, forensic fail-closed Layer 4.
- **VaR-only:** raise \`JSONDecodeError\` on hard-fail (so the callsite falls through to default narrative dict), recovery WARNING that only fires when the strict path missed.

## Audit-trail invariant

The new \`extract_json_value\` returns \`(value, parse_method)\` where \`parse_method\` is one of three module-level constants — \`PARSE_DIRECT\`, \`PARSE_LENIENT\`, \`PARSE_PASSTHROUGH\`. **Their string values (\`direct_json\`, \`json_first_object\`, \`passthrough\`) are preserved exactly** so \`ComplianceDecision.parse_method\` keeps the same labels in council_history / decision-signal CSVs. \`test_llm_json.py::test_parse_method_constants_are_stable_strings\` pins this.

## Test plan

- [x] \`pytest tests/test_llm_json.py tests/test_compliance.py tests/test_var_risk_agent_robust_json.py\` — 56 pass
- [x] \`pytest tests/test_compliance_*.py tests/test_var_calculator.py\` — 123 pass focused on the refactored modules
- [x] Wider pass excluding eval/harness — 1987 pass, 1 skipped, 425 deselected
- [x] 2 pre-existing weekend-only failures in \`test_connection_pool.py::TestRecoveryCascadeNoForceReset\` confirmed unrelated (\`Today is a weekend. Skipping position closing check\`) — also fail on \`origin/main\`

## Subtle behavioral detail

[...]

PR #2076, opened 2026-05-09
test: pin weekday in TestRecoveryCascadeNoForceReset (#2066)

### LINKED ISSUE #2066: test: TestRecoveryCascadeNoForceReset tests fail on weekends (missing weekday mock)
## Symptom

Two tests in `tests/test_connection_pool.py::TestRecoveryCascadeNoForceReset` fail when the test runs on a weekend:

- `test_close_stale_positions_does_not_force_reset_on_holdoff`
- `test_close_stale_positions_still_force_resets_on_real_failure`

Captured log:

```
WARNING  OrderManager:order_manager.py:4178 Today is a weekend. Skipping position closing check.
```

`close_stale_positions` early-returns on weekends, so the assertion `force_reset.assert_called_once_with(...)` finds 0 calls instead of 1.

## Reproduction

Run the suite on Saturday or Sunday — both fail. On weekdays, both pass.

## Fix

Mock the weekday check (or freeze the date with `freezegun` / `monkeypatch`) so the tests are deterministic regardless of which day they run.

## Why filing now

Surfaced today (2026-05-09, a Saturday) when the full suite was run for an unrelated PR (`fix/2063-close-verifier-poll`). Not blocking that PR — the 2 failures are completely independent of #2063's change. But CI weekend-runs would have caught this earlier; worth fixing so weekend test runs don't false-alarm.

## Severity

Low — tests are deterministic in CI which usually runs weekdays. Annoyance for local dev on weekends.

rozavala:
## Symptom

Two tests in \`tests/test_connection_pool.py::TestRecoveryCascadeNoForceReset\` fail when the suite runs on a weekend:

\`\`\`
WARNING  OrderManager:order_manager.py:4178 Today is a weekend. Skipping position closing check.
AssertionError: Expected 'mock' to be called once. Called 0 times.
\`\`\`

\`close_stale_positions\` early-returns on weekends (production-time guard, correct), so the retry-behavior assertions never fire. Surfaced today (Saturday) during the full-suite run for an unrelated PR (#2068).

## Root cause framing

- **Surface**: tests fail on weekends.
- **Proximate**: tests don't mock the weekday check.
- **Deeper**: \`close_stale_positions\` mixes three concerns — (a) market-day guard, (b) connection-pool retry logic, (c) actual close. The 2 failing tests target only (b), but (a) blocks them on Sat/Sun.

For #2066's scope, the mock is the right intervention because the weekday gate is genuinely orthogonal to the retry behavior under test. A clean decomposition (separate the three concerns into individually-testable units) is a bigger refactor — out of scope here, mentioned in the commit body for tracking.

## Fix

Add a \`_FakeWeekdayDatetime\` subclass that pins \`.now()\` to a Wednesday and apply it via \`patch('trading_bot.order_manager.datetime', _FakeWeekdayDatetime)\` in both tests' patch blocks. Subclassing \`datetime\` (rather than MagicMock) preserves all other datetime behavior — constructors, comparisons, arithmetic — so the rest of \`close_stale_positions\` (which does its own \`today_date + timedelta(days=1)\` etc.) works unchanged.

## Files

- \`tests/test_connection_pool.py\` — adds the subclass + patch lines, plus a \`from datetime import datetime, timezone\` import.

## Test plan

- [x] Targeted: \`TestRecoveryCascadeNoForceReset\` 2 tests pass (previously failed on Sat/Sun, passed Mon-Fri).
- [x] Full: \`pytest tests/\` — **2408 passed, 1 skipped, 0 failures**. Pre-PR baseline had the same 2 failures from this class.
- [ ] Operational: weekend test runs no longer false-alarm on these two assertions. Filed because the failures showed up during PR #2068's CI; this clears the path for legitimate weekend test runs.

## MECE

- **Value**: removes a weekend-only flake that masks real test signal. Tests are deterministic regardless of which day they run.
- **Risk**: zero — test-only change, production code untouched. Subclassing \`datetime\` for \`.now()\` override is a stable pattern that doesn't bleed into other test behavior.
- **Prior art**: similar weekday-fixed patterns elsewhere in \`tests/test_holding_time_gate.py\` for related order-manager weekday logic.

Closes #2066


rozavala:
Approved. Tiny test-only fix that unblocks weekend testing.

**Verified clean**:
- 22 added / 2 removed, all in `tests/test_connection_pool.py`
- Production code untouched
- Subclassing `datetime` (rather than mocking) preserves all other datetime behavior — constructors, comparisons, arithmetic — so the rest of `close_stale_positions` (which does `today_date + timedelta(days=1)` for tomorrow holiday checks) works unchanged
- Wednesday 2026-05-13 is a sensible pin (mid-week, mid-session)

**Root-cause framing is right** (per memory `feedback_review_root_cause_check.md`):
- Surface: tests fail on weekends
- Proximate: tests don't mock the weekday check
- Deeper: `close_stale_positions` mixes three concerns (market-day guard, connection-pool retry logic, actual close); the 2 affected tests target only retry logic but the market-day guard early-returns before they can run
- The PR consciously defers the deeper decomposition ("separate the three concerns into individually-testable units") as out of scope

This unblocks the path for several other PRs in the open queue (#2067, #2068, #2077, #2078, #2080) that all reported the same pre-existing weekend failures. Should ship promptly.

**MECE**:
- Value: removes weekend-only flake masking real signal. Tests deterministic regardless of which day they run.
- Risk: zero — test-only.
- Prior art: similar weekday-fixed patterns in `tests/test_holding_time_gate.py` for related order-manager weekday logic.

**Filed follow-up** (per the new `feedback_review_file_followups_as_issues.md` discipline): the deeper decomposition of `close_stale_positions` into separate testable units will be tracked as a separate issue.

Closes #2066.

Issue #1755, opened 2026-04-24
[CRITICAL] uncategorized: close_stale_positions FAILED after 3 attempts. Positions may remain open!


rozavala:
## Error Report

| Field | Value |
|-------|-------|
| **Category** | `uncategorized` |
| **Severity** | CRITICAL |
| **Environment** | DEV |
| **First seen** | 2026-04-24 16:56:11,378 |
| **Last seen** | 2026-04-24 16:56:11,378 |
| **Occurrences** | 1 |

### Sample Log Entry

```
close_stale_positions FAILED after 3 attempts. Positions may remain open!
```

### Impact

Investigate to determine impact on trading operations.

> The `claude-fix` label will be added after triage — Claude will automatically attempt a fix.

---
*Automated report by `scripts/error_reporter.py` — fingerprint: `60065a8e74e2`*

rozavala:
## Root cause (from DEV logs, 2026-04-24 16:55–16:56 UTC)

This is **not** the same incident as #1751 (Gateway wedge, ~12:16–14:03 UTC) — by 16:55 the Gateway was already recovered. This is a distinct bug in the recovery-queue + connection-pool interaction.

**Timeline:**
```
16:55:46  RECOVERY: Running missed close_stale_primary  → SB_orders connects (ID 837), succeeds, releases with hold-off
16:55:56  RECOVERY: Running missed close_stale_fallback → SB_orders in post-disconnect hold-off, 54s remaining
16:55:56  attempt 1/3 failed: …hold-off. Retrying in 5.0s   → _force_reset_connection (puts pool in 60s BACKOFF)
16:56:01  attempt 2/3 failed: …backoff period. Retrying in 10.0s → _force_reset_connection again
16:56:11  CRITICAL: close_stale_positions FAILED after 3 attempts  ← THIS ISSUE
```

**The bug** (`order_manager.py:4117–4126`): the retry loop catches `ConnectionError("…hold-off")` and calls `IBConnectionPool._force_reset_connection`, which floors backoff at `FORCE_RESET_BACKOFF_FLOOR = 60s` and stamps a fresh `_last_reconnect_attempt`. So a 60s hold-off (an *intentional* cooldown) is converted into a 60s backoff (a *failure-driven* cooldown) that the bounded retry budget (5 + 10 + 20 = 35s) cannot outlast.

[...]

Issue #2069, opened 2026-05-09
investigate: Gemini timeout silently drops 2/7 council voices — agronomist + geopolitical fall back to error struct, not OpenAI


rozavala:
## Symptom

When Gemini times out at 60s during initial research, the affected agents (agronomist, geopolitical) skip reflexion to avoid poisoned-state loops, return an error struct, and \`weighted_voting\` excludes them from the council vote. **No automatic per-call fallback to OpenAI** — the heterogeneous router only falls back when the *circuit breaker* trips on quota exhaustion, not on a single timeout.

Net effect: every emergency cycle that hits a Gemini timeout silently loses 2 of 7 specialist voices, even though OpenAI (the configured fallback) is healthy and the router has a fallback path that's clearly working in other contexts.

## Today's (2026-05-08) trace

\`\`\`
09:25:27 - Fallback event: geopolitical gemini -> openai. Reason: gemini quota exhausted (circuit breaker)
... circuit breaker is now OPEN for geopolitical ...

09:26:39 - [agronomist] Skipping reflexion — initial research returned error
09:26:46 - Attempt 1 failed for generate, retrying in 1.0s: Gemini timed out after 60.0s
09:26:49 - [agronomist] Skipping reflexion — initial research returned error
09:27:16 - Attempt 1 failed for generate, retrying in 1.0s: Gemini timed out after 60.0s
09:27:24 - [geopolitical] Skipping reflexion — initial research returned error
09:27:24 - Agent agronomist: Research failed (dict-wrapped) - Error: Gemini timed out after 60.0s
09:27:24 - Agent geopolitical: Research failed (dict-wrapped) - Error: Gemini timed out after 60.0s
\`\`\`

The circuit breaker for geopolitical was already OPEN at 09:25:27 (so it should fall back to OpenAI), but the agent still attempted Gemini at 09:27:16 and timed out again. Two failure modes overlap:

1. **Per-call fallback missing.** A single timeout doesn't trigger fallback — only the circuit breaker quota threshold does. So agents waste 60s+ each call on already-known-slow Gemini.
2. **Circuit breaker not consulted in reflexion path.** Even with the circuit breaker open for geopolitical at 09:25:27, the 09:27:16-09:27:24 reflexion attempt still tried Gemini. Suggests the reflexion path doesn't use the same router as initial research, or doesn't check breaker state.

## Frequency

[...]

Issue #2058, opened 2026-05-08
refactor: extract lenient LLM-JSON parser to shared module


rozavala:
## Background

PR #2055 added \`_robust_parse_llm_json\` to \`trading_bot/var_calculator.py\` to absorb markdown-fenced and prose-wrapped LLM responses. The helper is **a near-mirror** of \`compliance.py:from_llm_response\` (PR #1972 / #1929):

- Both strip markdown fences (\`\`\`json ... \`\`\`)
- Both try strict \`json.loads\` first
- Both fall back to \`JSONDecoder.raw_decode\` from the first \`{\` bracket
- Recovery semantic is intentionally identical across the two modules

PR #2055's review surfaced this as a transcription-drift trap: if someone adds support for a new fence syntax, a new edge case, or a different recovery rule to one helper, they need to update both. There's no enforcement that they will.

## Scope

Extract the shared recovery logic to a new module — proposed location: \`trading_bot/llm_json.py\` — with a single function the two modules import. The wrapper structures (\`ComplianceDecision\` dataclass for compliance, plain dict/list for var_calculator) stay in their respective modules; only the parsing/recovery primitive is shared.

## What this PR would touch

- New: \`trading_bot/llm_json.py\` with \`robust_parse_llm_response\` (or similar)
- Modify: \`trading_bot/compliance.py\` — replace inline parser with import
- Modify: \`trading_bot/var_calculator.py\` — replace \`_robust_parse_llm_json\` with import
- Tests: existing test files already cover the recovery cases; add a single new test file for the shared module if the helper has its own contract worth pinning.

## Why not now

PR #2055's review explicitly approved the conscious-mirror decision for shipping the fix today. This refactor is improvement work, not a bug fix — it stops a future drift, doesn't fix a current one. Worth tracking so the divergence trap doesn't materialize later.

## Refs

- PR #2055 — review comment that flagged the duplication
- PR #1972 — the original compliance.py lenient parser pattern (issue #1929)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2092, opened 2026-05-09
fix(connection-pool): wait_for_holdoff kwarg consolidates 4× retry sites (#2054)

rozavala:
## Summary

Closes #2054. Adds an opt-in \`wait_for_holdoff: bool = False\` kwarg to \`IBConnectionPool.get_connection\` that absorbs the post-clean-disconnect hold-off cooldown internally. Migrates the 4 manual \`except ConnectionInCooldown → sleep → retry\` sites to use it.

## What was happening

The same retry pattern existed at 4 caller sites with subtly different shapes:

| File | Line | Retry count | Log level | Trigger |
|---|---|---|---|---|
| orchestrator.py | 3970 | 1 | INFO | reconciliation |
| orchestrator.py | 6589 | 2 | WARNING | emergency_hard_close |
| order_manager.py | 1330 | 2 | WARNING | order generation |
| order_manager.py | 4217 | 3 | WARNING (with CRITICAL Pushover on exhaustion) | close_stale_positions |

Each one caught \`ConnectionInCooldown\`, slept \`e.wait_remaining + 1.0\`, retried. The differences in retry counts and log levels weren't principled — accumulated through different PRs that didn't cross-reference each other (transcription drift).

## What changed

Single-source-of-truth at the pool layer. Caller sites collapse from ~12-15 lines of try/except retry to a single \`get_connection(..., wait_for_holdoff=True)\` call. The semantics are explicit:

- **Hold-off cooldown** (post-clean-disconnect, max 60s) → opt-in waits in-pool
- **Backoff cooldown** (failure-driven, up to 600s) → still raises even with the kwarg

The backoff distinction is intentional: hold-off is benign hygiene ("we just disconnected, give Gateway a moment"), but backoff means a previous connection genuinely failed — sleeping through it would mask the problem. Migrated sites that still see \`ConnectionInCooldown\` after passing \`wait_for_holdoff=True\` know it's a backoff and escalate appropriately (defer / Pushover).

## Why this matters for #2054

1. **Deploy-time recovery race** (today's audit_pre_close failure at 18:27:37): recovery scheduler hits 60s hold-off between back-to-back same-pool tasks. Centralizing the wait makes it a single line to opt-in at the recovery loop's call site rather than editing every per-task path.
2. **NG_orders steady-state polling** (#2054 observation): polling cycle racing its own hold-off, burning ~50% of every cycle on fail-fast. With the kwarg, the polling caller waits one hold-off and resumes.

This PR adds the kwarg and migrates the 4 known sites. Opting-in *new* callers (recovery scheduler, polling loops) is a follow-up since each has its own retry/timeout context to consider.

## Test plan

- [x] 3 new tests in \`TestCleanDisconnectHoldoff\`:
  - \`test_wait_for_holdoff_true_sleeps_through_holdoff_then_proceeds\` — pins the core invariant (asyncio.sleep called with wait_remaining+slack, then connection acquired)
  - \`test_wait_for_holdoff_false_still_raises\` — default-behavior regression guard (both omitted-kwarg and explicit False)
  - \`test_wait_for_holdoff_does_not_swallow_backoff_cooldown\` — pins that backoff cooldowns bubble up even with the kwarg
- [x] 9 hold-off tests pass
- [x] 179 tests across connection_pool, order_manager, close_attribution, partial_close_orphan_guard, shared_contract_aggregation pass
- [x] 2 pre-existing weekend-only failures (\`TestRecoveryCascadeNoForceReset\`) confirmed unrelated — same \`Today is a weekend. Skipping position closing check\` short-circuit that's been failing on \`origin/main\` all weekend; test design issue independent of this PR

## Refs

- #2054 — issue this PR closes (with full trace + 4 fix candidates I considered)
- #1711 Q8 — original PR that introduced the post-disconnect hold-off; rationale ("avoid open/close churn during shutdown / cycle transitions") still holds, just better encapsulated now
- #1755 / #1837 — earlier work that added the manual retry sites this PR consolidates
- #2025 — recently consolidated retry policy at the LLM-router layer; this is the equivalent consolidation at the connection-pool layer


rozavala:
## Code Review + Test Results: ❌ 1 test fix required before merging

The `wait_for_holdoff` kwarg design is sound and the production code changes are correct. However the test run found one genuine failure that needs fixing.

---

### Test results

**6 failed, 2434 passed, 1 skipped:**

| Failure | Cause | Status |
|---|---|---|
| `test_mipro_*` (3 tests in `test_eval_dspy_trainer.py`) | Missing `dspy` dependency in test env | Pre-existing on `main`, unrelated |
| `TestRecoveryCascadeNoForceReset::test_close_stale_*` (2 tests) | Weekend-only: branch doesn't include #2076's `_FakeWeekdayDatetime` fix | Will resolve after #2076 merges + rebase |
| `test_reconcile_holdoff_retry.py::test_holdoff_then_success_does_not_log_critical` | **Test written for old behavior** | **Blocking — needs fix** |

---

### Issue 1 (blocking): `test_holdoff_then_success_does_not_log_critical`

The test simulates the old `reconcile_and_notify` pattern: mock `get_connection` to raise `ConnectionInCooldown(type="hold-off")` and expect the caller to sleep through it and retry. But the new code calls `get_connection(..., wait_for_holdoff=True)`, which absorbs the hold-off *inside* `get_connection` before raising — so a mock that forces the exception at the `get_connection` call boundary bypasses the internal sleep entirely.

The test needs to reflect the new behavior:
- **Old model**: caller catches hold-off, sleeps `wait_remaining + 1`, retries
- **New model**: `get_connection` internally sleeps and succeeds; caller only sees `ConnectionInCooldown` if a *backoff* (failure-driven) cooldown is also active

The fix is to update the mock setup: either (a) mock `get_connection` to succeed on first call (simulating that `wait_for_holdoff` absorbed the hold-off transparently), or (b) mock at the lower-level hold-off check so the `wait_for_holdoff` logic inside `get_connection` can actually execute.

---

### Issue 2 (non-blocking): Weekend test failures

`test_close_stale_positions_does_not_force_reset_on_holdoff` and `test_close_stale_positions_still_force_resets_on_real_failure` fail because today is a Sunday and `close_stale_positions` returns early before connection logic runs. These are the same failures that PR #2076 fixes — rebase on `main` after #2076 merges and they'll pass.

---

### Code review (production logic is correct)

- **Kwarg design:** Clean opt-in with fail-fast preserved as default. The hold-off vs. backoff distinction is correctly maintained: `wait_for_holdoff=True` absorbs hold-off cooldowns internally; backoff cooldowns still raise.
- **Sleep inside the lock:** Acknowledged in comments. Serializes hold-off waits per pool key — the intended behavior. Other callers on the same key queue behind, which is correct for the pool's serialization invariant.
- **`close_stale_positions` behavior change:** On backoff cooldown, the new code immediately sends CRITICAL + returns instead of retrying 3× first. This is correct — a backoff cooldown signals a real connection failure, not just timing. Faster escalation to the operator is the right tradeoff.
- **`emergency_hard_close`:** Simpler than before. Hold-off is absorbed; if backoff escapes, one explicit manual sleep + second attempt is tried. Behavior equivalent to the old 2-loop pattern.
- **`reconcile_and_notify`:** On backoff cooldown, defers immediately (non-fatal). Correct.

---

### Merge order (once fixes are applied)

Rebase in this order after prior PRs merge to `main`:
1. #2067 (llm_json refactor — stack base)
2. #2075 (router fix)
3. #2076 (weekday mock fix — resolves Issue 2 above)
4. #2078 (close verifier)
5. #2080 (recovery sleep)
6. **#2092** — fix Issue 1, then rebase on updated `main`

---

rozavala:
## Re-review (2026-05-10): same blocker, now also stale-vs-main

No new commits since yesterday's review. The blocking `test_holdoff_then_success_does_not_log_critical` failure is unchanged. Re-running `pytest tests/` confirms 3 failures, all already characterized in the previous review:

| Failure | Status |
|---|---|
| `test_holdoff_then_success_does_not_log_critical` | **Blocking — test still written for old behavior, needs to be rewritten for the `wait_for_holdoff=True` design** |
| `TestRecoveryCascadeNoForceReset::test_close_stale_*` (2 tests) | Resolves on rebase — `_FakeWeekdayDatetime` is now on `main` via #2076 |

(The 3 `test_mipro_*` failures from the prior review are gone in this run — `dspy` is presumably installed in the env now.)

---

### Branch is now stale: rebase needed before fix

Since this PR's review yesterday, three of its four stacked commits have been independently merged via separate PRs:

| Commit on this branch | Equivalent on `main` |
|---|---|
| `a107c5422` — refactor llm_json (#2058) | #2067 (`51194ff15`) |
| `c552bfa4a` — router fix (#2069) | #2075 (`c64e390da`) |
| `939854d3e` — close verifier (#2063) | #2078 (`a9460eea5`) |
| `e4195d0f2` — **wait_for_holdoff (#2054)** | **NOT on main — this is the PR's unique content** |

`main` also got #2076 (`194eab44f`, weekday pin) and #2080 (`658a796c0`, recovery inter-task sleep). #2080 is sibling work for the same #2054 root cause but uses a different mechanism (inter-task sleep in the recovery scheduler) — it does not conflict with `wait_for_holdoff`, the two are complementary.

**A clean `git rebase origin/main` should drop the three already-merged commits as no-ops** (their content is now identical via the merge-commit equivalents) and leave only `e4195d0f2` (wait_for_holdoff) on top of current `main`. The diff against `main` should then shrink to the actual delta this PR adds:

- `wait_for_holdoff` kwarg in `connection_pool.IBConnectionPool.get_connection`
- Migration of 4 caller sites in `orchestrator.py` + `order_manager.py`
- 3 new tests in `TestCleanDisconnectHoldoff`
- The fix to `test_holdoff_then_success_does_not_log_critical`

---

### Fix for `test_holdoff_then_success_does_not_log_critical`

The test mocks `IBConnectionPool.get_connection` with `side_effect=[cooldown, recon_ib]` and expects the orchestrator to sleep `wait_remaining + 1.0` between the two calls. Under the new design, `get_connection(wait_for_holdoff=True)` absorbs the hold-off **inside** the pool — the orchestrator calls it exactly once and does not sleep itself.

Two equally valid rewrites:

**(a) Reflect the new contract — no caller-side retry, no caller-side sleep:**
```python
async def test_holdoff_absorbed_by_pool_does_not_log_critical(self, base_config, caplog):
    """With wait_for_holdoff=True, the pool absorbs the hold-off internally
    and returns a usable IB. The orchestrator sees a single successful
    get_connection call — no caller-side sleep/retry."""
    recon_ib = MagicMock()
    get_conn = AsyncMock(return_value=recon_ib)  # absorbed; success on first call

    with patch.object(IBConnectionPool, 'get_connection', get_conn), \
         patch.object(IBConnectionPool, 'release_connection', new_callable=AsyncMock), \
         patch('orchestrator.run_reconciliation', new_callable=AsyncMock,
               return_value=(_empty_recon_df(), _empty_recon_df())), \
         patch('orchestrator.reconcile_active_positions', new_callable=AsyncMock), \
         patch('orchestrator.reconcile_council_history', new_callable=AsyncMock), \
         caplog.at_level(logging.WARNING, logger='Orchestrator'):
        await orchestrator.reconcile_and_notify(base_config)

    # Single call, with wait_for_holdoff=True
    get_conn.assert_called_once()
    assert get_conn.call_args.kwargs.get('wait_for_holdoff') is True

    critical_msgs = [r for r in caplog.records if r.levelno >= logging.CRITICAL]
    assert not critical_msgs, \
        f"Pool-absorbed hold-off must not log CRITICAL: {[r.message for r in critical_msgs]}"
```

**(b) Test backoff-bubbles-up path** (closer to what the original test was named for):
Replace the hold-off cooldown with a backoff cooldown — `wait_for_holdoff` does NOT absorb backoff cooldowns, so they still bubble up as `ConnectionInCooldown` and the orchestrator should log a WARNING (not CRITICAL) and defer. The existing `test_two_cooldowns_demotes_to_warning_not_critical` already tests something close to this, so consider whether (a) is sufficient on its own.

I recommend (a) — the existing two-cooldown test already covers the bubble-up-defer path, so test (a) only needs to assert the new contract (single call, kwarg present, no caller-side retry).

---

### Suggested merge order (revised)

1. Rebase `fix/connection-pool-wait-for-holdoff-2054` on `origin/main` (the 3 already-merged commits should drop as no-ops; `e4195d0f2` lands on top of #2076/#2077/#2080).
2. Apply test fix above.
3. Re-run `pytest tests/test_reconcile_holdoff_retry.py tests/test_connection_pool.py -v` — should be clean.
4. Push, re-review.

Code review verdict from yesterday stands: **production logic is correct.** Only the test mismatch and the rebase remain.

Re-applying `claude-fix` label.

---

Issue #1628, opened 2026-04-16
bug: VaR computation fails silently — var_state.json stale 20+ hours


rozavala:
## Summary

VaR computation is failing with "Failed to fetch aligned returns — data quality guard triggered" (2 occurrences Apr 15). The \`var_state.json\` file is now **20+ hours stale** (last computed: 2026-04-15T18:32:11) yet reports \`is_stale: false\`, making the staleness flag unreliable.

## Evidence

From \`error_reporter_state.json\` (fingerprint \`8fc0d33239da\`):
\`\`\`
VaR computation failed: Failed to fetch aligned returns — data quality guard triggered
\`\`\`
- Count: 2 occurrences on 2026-04-15 14:10:45

From \`data/var_state.json\`:
\`\`\`json
{
  "timestamp": "2026-04-15T18:32:11.887321+00:00",
  "is_stale": false,
  "var_99": 749.54,
  "position_count": 10,
  "commodities": ["NG", "SB"],
  "last_attempt_status": "OK",
  "last_attempt_error": ""
}
\`\`\`
The state was last computed on Apr 15 at 18:32 UTC — over 20 hours ago. Despite this, \`is_stale\` is \`false\` and \`last_attempt_status\` is \`"OK"\`, giving a false sense of freshness.

## Root Cause Hypothesis

1. **Data quality guard** triggers when aligned return series cannot be constructed (likely due to the ZWN26.NYB 404 error — see related #1620). The VaR attempt fails, but the state file is left with the previous successful values and status \`OK\`.

2. **is_stale not updated after failure**: The staleness flag is only set when the computation runs but produces suspicious results, not when the computation itself fails to run. A failed attempt should set \`last_attempt_status = "ERROR"\` and trigger staleness.

3. **No alerting on VaR staleness**: After 24 hours without a successful VaR update, no Pushover notification is sent.

## Impact

- Portfolio-level risk estimates are 20+ hours out of date
- \`is_stale: false\` falsely signals valid data to consumers
- Portfolio dashboard shows stale VaR without any visual indicator
- Risk decisions may be based on yesterday's positions and prices

## What to Fix

1. When VaR computation fails, set \`last_attempt_status = "ERROR"\` and \`last_attempt_error\` to the failure reason in the state file
2. Define an absolute staleness threshold (e.g., 2 hours) — if \`computed_epoch\` is older than that, force \`is_stale = true\` regardless of the last attempt status
3. Send a Pushover alert when VaR is stale beyond the threshold

PR #2118, opened 2026-05-10
fix(var): emit NO_DATA / DEGRADED status when positions can't be priced (#2100)

rozavala:
## Summary

Closes #2100. The VaR cycle was silently saving an empty result (`position_count=0, equity=0.0, var_95=0.0`) with `last_attempt_status="OK"` whenever the underlying / IV fetch failed for all positions. Downstream consumers — Article V compliance gate, dashboards — trusted the empty result as a fresh authoritative reading and went blind to the degraded state.

This PR makes the failure visible by adding two new status values, both setting `is_stale=True`:

| Status | Meaning |
|---|---|
| `OK` | Normal computation |
| `FAILED` | Exception, prior numbers preserved (existing behavior) |
| **`NO_DATA`** | Positions exist in portfolio but ALL excluded due to missing underlying / IV data (the main #2100 bug) |
| **`DEGRADED`** | Positions priced but every option position used fallback IV (the "N/N positions fall back" case from the issue) |

## Mechanism

Detection lives on `PortfolioVaRCalculator._last_snapshot_meta`, populated by `_snapshot_portfolio` with `total_input_positions`, `option_input_count`, and `fallback_iv_count`. Using an instance attribute rather than changing the return type avoids churn in the 3 other callers and 5 test sites that use `_snapshot_portfolio` / patch it.

`compute_portfolio_var` reads the meta after the snapshot call:
- empty snapshots **AND** `total_input > 0` → `NO_DATA` + `is_stale=True` + descriptive error
- positions priced **AND** `fallback_iv == option_input` → `DEGRADED` + `is_stale=True`
- otherwise → `OK` as before

Partial fallback (< 100%) keeps `OK` — the existing WARNING at the 50% threshold already covers that case; only N/N escalates to status.

## Consumer updates

- **`pages/1_Cockpit.py`** and **`pages/9_Portfolio.py`** — dashboards now surface any non-`OK` status (was: `FAILED`-only), so the new statuses are visible to operators.
- **`trading_bot/compliance.py`** Article V — explicit `NO_DATA` fail-closed check at the gate, with startup grace. The defense-in-depth is intentional: `get_cached_var`'s current OK-only filter would already hand `None` to the gate (which the existing `cached is None` branch fail-closes), but documenting the policy at the call site makes it discoverable for future maintainers and testable directly.

## What this does NOT fix

The underlying connection root cause flagged in #2100 — \`VaR: Failed to fetch underlying for KC: Not connected\` — is a separate IB connection / pool-lifecycle question (why does the calculator's IB connection drop out during a fetch?). That investigation needs its own PR.

Per `feedback_review_root_cause_check.md`, this PR is fixing a **distinct** root cause (status reporting that was hiding degraded output) from the **proximate symptom** (connection drops causing the fetch failures). Both are real; this PR closes the visibility gap so the symptom can no longer silently bypass compliance gates.

## Test plan

- [x] `test_var_calculator.py`:
  - `test_empty_portfolio_var_is_zero` — extended to assert `is_stale=False`
  - `test_no_data_status_when_positions_excluded` — new: NO_DATA + stale
  - `test_degraded_status_when_all_options_on_fallback_iv` — new: DEGRADED + stale, position_count > 0
  - `test_ok_status_when_some_options_on_fallback_iv` — new: partial fallback stays OK
- [x] `test_compliance_var.py`:
  - `test_enforce_no_data_status_rejects` — new: NO_DATA fail-closes Article V in enforce mode
- [x] Full \`pytest tests/\` — **2534 passed, 1 skipped, 0 failed** (7m 47s)

## Refs

- #2100 — the originating issue with full reproduction evidence
- #2009 — original PR that wired live IV fetch into VaR path (likely surface for the underlying connection issue)
- #2091, #2040 — adjacent VaR follow-ups in the same code area


rozavala:
## Approved — MECE Review

**Value: HIGH.** Fixes a real fail-open masquerading as fail-closed. The pre-fix shape (`position_count=0, equity=0.0, var_95=0.0` with `last_attempt_status=\"OK\"`) made Article V's `> var_limit` check silently pass on all-zero output. Exactly the kind of \"Safety First\" / \"fail closed\" gap CLAUDE.md mandates closing. Echoes the same anti-pattern as #1628 (silent VaR failure) so the producer-side hardening is well-precedented.

**Risk: LOW.**

- **Article V fail-closed wiring verified.** `get_cached_var()` already filters non-OK statuses to `None` (line 403/407 of `var_calculator.py`), and compliance's existing `cached is None` branch (line 679) fail-closes with startup grace. The new explicit `NO_DATA` branch (line 692) is defense-in-depth — technically unreachable today, but documents policy at the call site for future maintainers, and the test directly proves the gate fails closed without relying on the implicit filter. Both layers tested.
- **Idle-portfolio false-positive: SAFE.** `_snapshot_portfolio` resets `_last_snapshot_meta` to zeros every call. Empty `ib.portfolio()` → `total_input_positions=0` → producer takes the `OK` branch (line 208). Confirmed by `test_empty_portfolio_var_is_zero` — explicitly asserts `is_stale=False` and `last_attempt_status=\"OK\"`. No idle-system regression.
- **Backward compat.** `VaRResult.from_dict` uses `.get(\"last_attempt_status\", \"OK\")` (line 146) — pre-#2118 state files parse cleanly. Defaults preserve OK-shape.
- **Other VaR consumers** (`weighted_voting`, `signal_generator`, `orchestrator` emergency briefing, `order_manager`) all gate on `if cached_var:` truthy checks. Since `get_cached_var` returns `None` for NO_DATA/DEGRADED, each silently skips its VaR branch — non-fatal and consistent with current degraded-data handling.

**Prior art: SOUND.** Pattern aligns with #1628 (silent VaR failure, closed) and adjacent VaR work #2009/#2040/#2091. The PR correctly scopes out the connection root cause (the `Not connected` IB-pool issue) per `feedback_review_root_cause_check.md`: this fixes the *distinct* root cause of status reporting that was hiding degraded output, leaving the proximate-symptom (connection drops) for a separate PR.

**Tests.** 2534 passed, 1 skipped, 0 failed (548s locally — full suite). New test coverage:
- `test_no_data_status_when_positions_excluded` (producer)
- `test_degraded_status_when_all_options_on_fallback_iv` (producer)
- `test_ok_status_when_some_options_on_fallback_iv` (partial-fallback boundary)
- `test_enforce_no_data_status_rejects` (Article V consumer)
- `test_empty_portfolio_var_is_zero` extended to assert idle is NOT stale

**Follow-up filed: #2120** — `_snapshot_portfolio` swallows `ib.portfolio()` exceptions and returns `[]` with zeroed meta, which the new producer logic treats as \"genuinely empty\" → emits `OK`. Same fail-open shape, different upstream failure mode. Non-blocking; scoped as residual case for the connection-root-cause investigation track.

LGTM — clean fail-open → fail-closed flip with verified Article V wiring and good regression coverage.

PR #2123, opened 2026-05-11
fix(var): emit UNAVAILABLE when ib.portfolio() raises — close residual fail-open (#2120)

### LINKED ISSUE #2120: investigate: VaR _snapshot_portfolio swallows ib.portfolio() exception → emits status=OK (residual fail-open after #2118)
## Summary

PR #2118 fixes the main VaR fail-open by emitting `NO_DATA` when positions exist but all fail to price. However, one residual silent-fail path remains in `trading_bot/var_calculator.py:_snapshot_portfolio`:

\`\`\`python
try:
    portfolio_items = ib.portfolio()
except Exception as e:
    logger.error(f\"Failed to read ib.portfolio(): {e}\")
    return []
\`\`\`

When this exception fires (e.g. the \`Not connected\` IB-pool issue called out as the separate root cause in #2100), `_snapshot_portfolio` returns `[]` with `_last_snapshot_meta[\"total_input_positions\"]=0`. The producer in \`compute_portfolio_var\` then takes the \"genuinely no positions\" branch and emits `last_attempt_status=\"OK\"` — the same fail-open shape #2118 was designed to eliminate, just from a different upstream failure mode.

## Reproduction

Trigger any condition that makes `ib.portfolio()` raise (broken connection, RPC timeout, IB Gateway wedged mid-call). The VaR cycle will silently report `OK` with zero positions despite an underlying connection error.

## Why this matters

- Article V's \`NO_DATA\` defense-in-depth (added in #2118) does NOT catch this path — the result is genuinely OK-shaped.
- Dashboards display green status icon.
- Downstream consumers (signal_generator, weighted_voting, orchestrator emergency briefing) all use truthy checks that silently skip the section, hiding the issue from operators.

## Proposed fix

Distinguish \`ib.portfolio()\` exception from a genuinely-empty portfolio. Options:

1. Emit a new \`UNAVAILABLE\` status when \`portfolio_items\` call raises, separate from \`NO_DATA\` (positions exist but unpriced) and \`OK\` (truly empty).
2. Or extend the meta dict with an \`ib_call_failed: bool\` flag that \`compute_portfolio_var\` checks before deciding OK vs the new status.

## Scope

Non-blocking follow-up to #2118. The PR explicitly scopes out the connection-root-cause investigation per the issue body; this complements that work by closing the residual visibility gap on the producer side.

## Refs

- #2118 — main fail-open fix (this issue is the residual case it doesn't cover)
- #2100 — original fail-open report; explicitly flagged the \`Not connected\` IB-pool issue as a separate investigation

🤖 Generated during /review-prs cycle on 2026-05-10.

rozavala:
## Summary

Closes the residual VaR fail-open path that survived PR #2118.

PR #2118 fixed the main case: positions exist in `ib.portfolio()` but all fail to price → emit `NO_DATA`. This PR closes the complementary case: `ib.portfolio()` itself raises (e.g. "Not connected" from a wedged IB pool) → `_snapshot_portfolio` swallowed the exception, returned `[]`, and `compute_portfolio_var` fell through to `last_attempt_status="OK"` with zero positions — the same fail-open shape #2118 was designed to eliminate, just from a different upstream failure mode.

## Changes

- **`_snapshot_portfolio`**: records `ib_call_failed` / `ib_call_error` in `_last_snapshot_meta` when `ib.portfolio()` raises (in addition to the existing `logger.error`).
- **`compute_portfolio_var`**: checks the flag *before* the existing `total_input_positions > 0` branch and emits `last_attempt_status="UNAVAILABLE"` + `is_stale=True` with the captured error message. The genuinely-empty-portfolio path keeps `OK`.
- **`compliance.py` Article V**: treats `UNAVAILABLE` the same as `NO_DATA` — fail-closed in enforce mode, with startup grace. Reason string distinguishes the two for forensic clarity.
- **`VaRResult` docstring**: enumerates all five status values (`OK`, `FAILED`, `NO_DATA`, `DEGRADED`, `UNAVAILABLE`).

Dashboards (`pages/1_Cockpit.py`, `pages/9_Portfolio.py`) need no change — they already render any non-`OK` status with a red icon and surface `last_attempt_error`.

## Test plan

- [x] `tests/test_var_calculator.py::test_unavailable_status_when_ib_portfolio_raises` — mirrors the NO_DATA test, asserts status=UNAVAILABLE, is_stale=True, error includes `ib.portfolio()` + captured underlying error.
- [x] `tests/test_compliance_var.py::test_enforce_unavailable_status_rejects` — mirrors the NO_DATA enforce-mode test, asserts approved=False and reason includes both "Article V" and "UNAVAILABLE".
- [ ] Verify existing NO_DATA tests still pass (gate now uses `in ("NO_DATA", "UNAVAILABLE")` tuple instead of `==`).
- [ ] Confirm `get_cached_var` continues filtering non-OK results (returns `None` for `UNAVAILABLE` just like `NO_DATA`), so the primary fail-closed path is the "VaR not computed" branch above the explicit status check.

Local tests not run — sandbox lacks `numpy`. Both touched source files and both test files pass `python -m py_compile`.

Closes #2120.

https://claude.ai/code/session_01EmBXjcsngyG2oL99oPLhjC

---
_Generated by [Claude Code](https://claude.ai/code/session_01EmBXjcsngyG2oL99oPLhjC)_


rozavala:
## MECE Review — APPROVED

**Value: HIGH.** Closes the residual fail-open class identified in #2118's review. `ib.portfolio()` raising (e.g. "Not connected" from wedged IB pool) → was silently returning `[]` → `compute_portfolio_var` emitted `status=OK` with zero positions. After fix: `_snapshot_portfolio` stamps `_last_snapshot_meta["ib_call_failed"]=True` + captured error; producer checks the flag **before** the existing `total_input_positions > 0` branch and emits `status=UNAVAILABLE` + `is_stale=True`. Same fail-closed shape as the NO_DATA fix.

**Risk: LOW — verified the wiring.**

1. **Article V wiring (`compliance.py:692`):** gate clause expanded from `== "NO_DATA"` to `in ("NO_DATA", "UNAVAILABLE")`. Both paths fail-closed in enforce mode with shared startup-grace branch; distinct rejection reasons for forensic clarity. `get_cached_var` (line 424/428) only returns `OK` results, so UNAVAILABLE also flows through the `cached is None` → reject branch — defense in depth.

2. **Idle-portfolio distinction (the #2118 concern):** producer correctly orders the three cases:
   - `meta["ib_call_failed"]` → UNAVAILABLE
   - `meta["total_input_positions"] > 0` (excluded) → NO_DATA
   - else (genuinely empty) → OK
   The flag is reset on every `_snapshot_portfolio` call, so no stale-flag risk. Idle-portfolio false-positive **not present**.

3. **Other direct `ib.portfolio()` callers:** two found — both safe.
   - `compliance.py:321` `_check_concentration_risk` — already wraps whole block in try/except returning `False, "fail-closed"` (line 350-352).
   - `risk_management.py:230` `get_unrealized_pnl` — no try/except; exception propagates to `_calculate_combo_risk_metrics` which would crash, not silently fail-open. Not the same class.

**Prior art:** #2118 (NO_DATA producer fix), #2100 (umbrella issue). This is the residual case those explicitly scoped out. Pattern is consistent across all three.

**Tests: 2537 passed, 1 skipped (full suite, 7m35s).** Regression tests present:
- `test_unavailable_status_when_ib_portfolio_raises` — asserts status=UNAVAILABLE, is_stale=True, error contains `ib.portfolio()` + underlying cause.
- `test_enforce_unavailable_status_rejects` — Article V mirror of NO_DATA enforce test.
Existing NO_DATA enforce tests still pass with new tuple membership check.

**Article V wiring verified: yes.** **Idle false-positive risk: no.** **Other unfixed callers: none in same class.**

Ship it.

PR #2173, opened 2026-05-11
perf(ib): cache 1Y historical IV fetch with 1h TTL (#2040)

rozavala:
## Summary

Closes #2040. PR #2009 wired \`reqHistoricalDataAsync\` for trailing-1Y daily IV into \`get_underlying_iv_metrics\`, which is called per cycle by the VaR / regime / volatility-specialist paths. One IB call per ticker per engine per cycle compounds during Gateway-wedge windows where the fetch path is already saturated.

Caches the fetched bars per-contract for 1h. Strict latency win — the 1Y rolling distribution doesn't drift meaningfully within an hour.

## What this PR adds

| Symbol | Purpose |
|---|---|
| \`_HISTORICAL_IV_CACHE\` | \`dict[conId, (timestamp, closes)]\` — module-level cache |
| \`_HISTORICAL_IV_TTL_SECONDS = 3600\` | 1h staleness boundary |
| \`_fetch_historical_iv_closes(ib, contract)\` | Helper: cache lookup → fetch-on-miss → return closes |
| \`_reset_historical_iv_cache()\` | Test-only reset (per-test \`setUp\`) |

\`get_underlying_iv_metrics\` now delegates to the helper instead of calling \`reqHistoricalDataAsync\` directly.

## Design choices

- **Empty fetches are NOT cached.** A transient Gateway hiccup returning 0 bars must not poison the cache for an hour.
- **Per-process cache.** Each engine is its own process; no cross-process sharing needed.
- **conId-only key.** IBKR re-assigns conIds on contract roll, so the roll-over case is implicit. (Issue body suggested \`(conId, lastTradeDateOrContractMonth)\` but the date is redundant given that property.)

## Test plan

5 new tests in \`tests/test_ib_interface.py::TestHistoricalIVCache\`:

- \`test_cache_miss_calls_ib\` — first call hits IB
- \`test_cache_hit_avoids_ib_call\` — second call within TTL serves cached value (verified IB call count stays at 1)
- \`test_different_conid_doesnt_share_cache\` — per-contract isolation
- \`test_empty_result_not_cached\` — Gateway-hiccup recovery (cache not poisoned with empty)
- \`test_ttl_expiry_forces_refetch\` — staleness boundary

- [x] \`pytest tests/test_ib_interface.py::TestHistoricalIVCache\` — 5/5 pass
- [x] \`pytest tests/\` — full suite **2613 passed, 1 skipped, 0 failed** (7m 41s)

## Operational impact (estimate per issue body)

- ~500ms-2s saved per cycle when warm
- ~10-20 cycles per ticker per day → 20-40s of latency saved per ticker per day during normal Gateway response times
- ~95% reduction in IB historical-data API calls (cache hit ratio)
- More-noticeable savings during Gateway congestion — the path's main pain point

## Refs

- PR #2009 — wired the per-cycle fetch this caches
- #2039 — sibling PR using the same IV-rank metric (per-commodity \`condor_iv_rank_breach\`)
- \`feedback_review_root_cause_check.md\` — the per-cycle fetch wasn't broken, just expensive; this is an optimization, not a correctness fix


rozavala:
## MECE approval — clean perf optimization

- **Value:** ~95% reduction in IB historical-data API calls per cycle (per PR body); ~500ms-2s saved when warm; especially valuable during Gateway-congestion windows where the per-cycle fetch was already compounding the wedge problem. The 252-day rolling distribution doesn't drift meaningfully in 1h, so the staleness tradeoff is sound.

- **Risk:** **LOW** — verified independently at `trading_bot/ib_interface.py`:
  - Cache affects ONLY the historical 1Y denominator at line 414 (`_fetch_historical_iv_closes`). **The fresh `current_iv` numerator at line 390 (`ticker.modelGreeks.impliedVol` via `reqMktData`) is uncached** — so iv_rank/iv_percentile are computed against (fresh anchor) ÷ (cached distribution), which is exactly the right shape.
  - **Empty fetches NOT cached** (line 361: `if closes and key:`) — Gateway-hiccup recovery preserved, no cache poisoning.
  - **Per-conId key** — IBKR re-assigns conId on contract roll, so the old roll's cached entry simply becomes unreachable (small memory leak bounded by the number of roll events × engine lifetime; not a practical concern).
  - **`if key:` guard** at line 340 handles the edge case of an unqualified contract gracefully (falls through to a real fetch, which is the safe behavior).
  - **Off-RTH fallback path** at line 432 also benefits (uses the same cached closes for both the anchor and the comparison set — internally consistent within the same cycle).
  - **Test reset helper** (`_reset_historical_iv_cache`) exposed at module level for test isolation — required for global-state, correctly addressed.

- **Prior art:** #2009 (the per-cycle fetch this caches), #2039 (sibling IV-rank consumer), #2040 (this PR's authorizing issue). `feedback_review_root_cause_check.md` correctly cited — this is an optimization, not a correctness fix.

- **Tests:** 5/5 new `TestHistoricalIVCache` tests pass (2.27s); author's full suite reports 2613 passed.

---

PR #2180, opened 2026-05-11
feat(observability): warn when primary-engine dedup misses a day (#2116)

rozavala:
## Summary

Closes #2116. PR #2112 added \`commodity_filter='primary'\` so observability monitors fire only on KC — correct dedup, but #2087 documented the tradeoff: **if KC's engine is down through its post-close window, the monitor doesn't fire that day, silently**. This PR makes that brittleness observable.

## Implementation

\`_check_yesterday_abstention_monitor_output\` runs from \`generate_system_digest_task\` (18:45 ET, every engine) and:

1. Computes yesterday's date.
2. Skips weekends and exchange holidays (expected absence).
3. Checks \`logs/abstention_monitor/{YYYY-MM-DD}.json\` exists.
4. If missing → WARNING log + Pushover.
5. \`pushover_throttle.should_send_pushover\` keyed by yesterday's date — 4 engines running system_digest fire only **one** Pushover per missed day.

Failures NEVER propagate (observability cannot kill the digest task).

## Why Option 2 (system_digest) not Option 1 (KC startup)

#2116's body offered two options. Option 1 (check at next-day's KC startup) has the same failure mode it's trying to detect: if KC stays down, the check itself doesn't fire. Option 2 (system_digest, runs on every engine) is more robust — any healthy engine catches the miss.

## Test plan

\`tests/test_abstention_monitor_observability.py\` — 5 tests:

| Test | Scenario |
|---|---|
| missing output on business day | warning + Pushover fires |
| output present | no warning, no Pushover |
| weekend yesterday | no alert (expected absence) |
| throttle dedupes across 4 engines | one alert per missed day |
| warning logs to 'Orchestrator' logger | operator grep target |

- [x] \`pytest tests/test_abstention_monitor_observability.py\` — 5/5 pass
- [x] \`pytest tests/\` — full suite **2632 passed, 1 skipped, 0 failed** (8m 19s)

## Why this is not an escape hatch

Per \`feedback_review_root_cause_check.md\`: the dedup (#2112) is intentionally Option A. The brittleness is a known design tradeoff. This PR doesn't paper over it — it surfaces it. Per #2116's escalation criterion: if the alert fires \"more than ~2×/month, that's the signal to do Option B (the cleaner global hook).\"

## Refs

- #2087 — original Option-A vs Option-B decision
- #2112 — the dedup PR that introduced the brittleness
- #2137 / \`trading_bot/pushover_throttle.py\` — the cross-engine cooldown helper used for dedup


rozavala:
## MECE Approval

**Value: HIGH** — closes #2087's documented blind spot from #2112. Without this, a multi-day primary-engine (KC) outage silently leaves observability dark; with this, missed days become a Pushover. Sensible Option 2 (digest hook, every engine) > Option 1 (KC startup, same failure mode).

**Risk: LOW** — pure-additive (+252/-0). Failures wrapped (`observability must never kill the digest task`). Weekend ✓ and exchange-holiday ✓ short-circuits. Throttle keyed on yesterday's date prevents 4× page storm across engines (engines are asyncio Tasks sharing one process under MasterOrchestrator, so process-local `_pushover_cooldowns` works as cross-engine dedup).

**Prior art: confirmed novel.** #2087 documented Option-A brittleness explicitly. #2116 authorized this fix. Choice of digest-hook over startup-hook is justified (Option 1 has same failure mode it's detecting).

## Test verification
- New `tests/test_abstention_monitor_observability.py`: **5/5 pass**
- `tests/test_abstention_monitor.py + test_optimized_prompt_*.py + test_schedule_*.py`: **63/63 pass**
- `tests/test_system_digest.py`: **44/44 pass** (no regression in the hosting task)
- PR body: `pytest tests/` full suite **2632 passed, 1 skipped, 0 failed**

## Non-blocking follow-ups filed
- **#2182** — bootstrap suppression. First business day after deploy, `logs/abstention_monitor/` legitimately doesn't have yesterday's file → one-time false-positive page. Throttle bounds the noise; worth a guard later.
- **#2183** — extend the check to `opt_prompt_monitor` (Fridays-only sibling, also primary-engine-deduped, also brittle). The script doesn't persist a daily JSON, so needs a small persistence change before the same pattern can apply.

Approved for merge.

PR #2181, opened 2026-05-11
feat(config): per-commodity condor_iv_rank_breach with global fallback (#2039)

rozavala:
Closes part of #2039 (the infrastructure half).

## What this is

#2039 surfaced as a deferred concern in #2009 (which computed real IV-rank from 1Y historical IV). With per-commodity IV distributions now measurable, the single global `exit_logic.condor_iv_rank_breach` threshold (70) is probably wrong for at least NG — IV-rank 70 fires often during weather/EIA spikes that mean-revert.

This PR adds the **infrastructure** for per-commodity tuning. It does **not** change any actual threshold values — every commodity still uses the global 70 until data-driven tuning happens (the analysis half, deferred because it needs 3-month observation windows of live IV-rank data).

## What changed

* `config/commodity_profiles.py` — new optional field `condor_iv_rank_breach: Optional[float] = None` on `CommodityProfile`. `None` means "use global"; the current posture for all 4 active profiles.

* `trading_bot/thesis_validation.py` — new helper `_resolve_condor_iv_rank_breach(config, exit_cfg)` with resolution order:
  1. `CommodityProfile.condor_iv_rank_breach` (None → skip)
  2. `config['exit_logic']['condor_iv_rank_breach']`
  3. Hardcoded 70 (pre-#2039 default)

  Profile lookup failure falls through to the global — the IC early-exit check is **fail-open by design**; a config-load issue must not accidentally force-close every IC.

* `_validate_iron_condor` reads from the helper instead of the inline `exit_cfg.get(...)`.

## What does NOT change

* No commodity profile JSON file is edited. All four profiles still have `condor_iv_rank_breach` absent → resolver picks the global `70` from `config.json`. **Zero behavior change** in production for this PR alone.

* The data-driven tuning step (per-commodity threshold values from 6–12 month IV-rank distribution analysis) is a separate task. The PR description of #2039 explicitly notes \"unknown without data whether the global 70 is in fact the right answer.\" This PR enables the override; it doesn't presume a direction.

## Tests

`tests/test_condor_iv_rank_threshold_2039.py` — 5 tests:

* Global wins when profile field is None (current production)
* Profile override wins when set
* Hardcoded 70 fallback when neither set
* Profile-lookup failure falls through to global (fail-open)
* Both `config['symbol']` and `config['commodity']['ticker']` work as lookup keys

Full `pytest tests/` — **2632 passed, 1 skipped, 0 failed (7m 56s).**

## Test plan

- [x] All 5 new tests pass
- [x] Full test suite passes (no regressions in `_validate_iron_condor` or condor flow)
- [x] No profile JSON edits → zero behavior change in production until per-commodity values are added by a follow-up


rozavala:
## MECE approval — infrastructure-only, identical behavior

- **Value:** MODERATE — adds the per-commodity tuning lane for `condor_iv_rank_breach` without changing any current threshold. The discipline of "infrastructure first, tuning later" matches `feedback_eval_persona_model_swap_fragile.md` — keeps the experimental tuning step isolated.

- **Risk:** **LOW** — verified independently:
  - **`CommodityProfile.condor_iv_rank_breach: Optional[float] = None`** at `config/commodity_profiles.py:198` — `None` default = "fall back to global"
  - **`_load_profile_from_json` at `:507`** wires `data.get('condor_iv_rank_breach', None)` — JSON profiles can opt in
  - **`_resolve_condor_iv_rank_breach(config, exit_cfg)` at `trading_bot/thesis_validation.py:113`** documents the resolution order (profile → global → 70 fallback) and is **fail-open on exception** (`except (ValueError, KeyError, ImportError, AttributeError): pass` → falls through to global) — correct: early-exit should NOT force-close on a config-load issue
  - **Call site migration at `:150`** — `iv_breach_threshold = _resolve_condor_iv_rank_breach(config, exit_cfg)` replaces the direct `exit_cfg.get('condor_iv_rank_breach', 70)` so the helper actually controls the threshold
  - **No JSON profile sets `condor_iv_rank_breach`** post-PR — every commodity still resolves to the global `70` (or whatever `exit_logic.condor_iv_rank_breach` is on main). Behavior identical to pre-PR.

- **Prior art:** #2039 (umbrella), #2009 (real IV-rank wiring, the predecessor), #2173 (sibling IV cache merged today).

- **Tests:** new `tests/test_condor_iv_rank_threshold_2039.py` covers the 3 resolution branches (global-used, per-commodity-overrides, fallback-to-70). Targeted + exit-logic + thesis-validation suites: **106/106 pass** in 42s.

**Follow-up:** the actual per-commodity tuning experiment (the empirical analysis to set NG/KC/SB/ZW thresholds) should be a separate PR/issue once enough live IV data is collected. The infrastructure here unblocks that work.

---

PR #2184, opened 2026-05-11
fix(sentinels): include exception type in Price-Sentinel error logs (#2151)

rozavala:
Closes #2151.

## Summary

The daily-summary error report for 2026-05-11 surfaced 7 instances of:

```
ERROR  Price Sentinel check failed:
```

…with **nothing after the colon**. The bug is the f-string at `sentinels.py:494` that interpolates `str(e)` — fine when `e` has a message, useless when it's an `asyncio.TimeoutError` (whose default `str()` is empty).

The 12s inner `asyncio.wait_for` budget at line 477 (#1975) is the single most common source of these exceptions, so the empty-message case is the dominant production path.

## Fix

One-line per site: prefix the message with `type(e).__name__`. Pattern lifted from `sentinels.py:1988` (X search EXCEPTION logging) — same approach is already used elsewhere in this file. Two call sites, both in `PriceSentinel.check`:

- Line 494 (ERROR, hourly intraday-move check) — the production-visible one
- Line 549 (DEBUG, cumulative multi-day check) — same bug shape, debug-only so noise is low, but the fix is free

After this PR, the log reads:

```
ERROR  Price Sentinel check failed: TimeoutError:
```

…enough for the error reporter to bucket and for an operator to triage at a glance.

## Tests

`test_price_sentinel_logs_exception_type_when_message_empty` — pins the new format. Uses a direct handler attached to the `trading_bot.sentinels` logger because that logger has `propagate = False` set at import time (line 45), so pytest's `caplog` (which attaches to the root logger) wouldn't see these records.

Full `pytest tests/` — **2635 passed, 1 skipped, 0 failed (9m 35s).**

## Test plan

- [x] New test pins the type-prefix format
- [x] Full test suite passes
- [x] No semantics change — sentinel still returns None on TimeoutError (per #1975), only the log message changes


rozavala:
**Claude Code review — LGTM ✅**

Minimal, targeted observability fix with zero behavioral change.

**Change analysis:**
- `type(e).__name__` is always a non-empty string — it cannot be `None` or raise — so the log is strictly more informative with no regression risk.
- Pattern already established at `sentinels.py:1988` (X search EXCEPTION logging). This PR makes the two `PriceSentinel` error sites consistent with the established convention.
- Fix applied at both the ERROR site (line 494, the production-visible one) and the DEBUG site (line 549) — both have the same bug shape, fixing both is correct.

**Test:**
- Direct `logging.Handler` attached to `trading_bot.sentinels` is the right approach because that logger has `propagate = False` at import time (line 45). pytest's `caplog` would have silently missed these records. The `try/finally` handler cleanup is correct.

**CI:** Ruff F821 ✅

Ready to merge.

---

Issue #2147, opened 2026-05-11
bug: Gemini 60s timeouts now fail emergency cycles outright — 4 ERROR cascades today (Council 300s + Emergency Cycle Failed)


rozavala:
## Summary

Today (2026-05-11, Mon RTH) had 4 ERROR-level emergency-cycle failures rooted in Gemini 60s timeouts: 2× `Council timed out after 300s` (SB 07:35 UTC, NG 13:10 UTC) and 2× `Emergency Cycle Failed: Gemini timed out after 60.0s` (NG 13:13, 13:20 UTC). The pattern is loud cascades replacing what was previously silent agent-dropping behavior — directly exposed by PR #2075 (closed 2026-05-10), which removed the dead-letter Gemini retry fallback that was masking the same upstream slowness investigated in #2069.

## Verbatim evidence — NG emergency cycle 13:10–13:20 (Mon RTH)

```
13:04:48,786  WARNING  SELF-HEAL: 14 deferred triggers accumulated...
13:09:33,452  WARNING  Fallback event: permabull xai -> openai. Reason: Request timed out.
13:10:34,465  ERROR    Council timed out after 300s during emergency cycle
13:10:34,467  INFO     Emergency Decision: NEUTRAL (0.0)   ← forced fallback, conf=0.0
13:13:35,598  ERROR    Emergency Cycle Failed: Gemini timed out after 60.0s
13:19:21,243  WARNING  Attempt 1 failed for generate, retrying in 1.0s: Gemini timed out after 60.0s
13:20:22,249  ERROR    Emergency Cycle Failed: Gemini timed out after 60.0s
```

## Gemini-timeout volume per engine (24h, weekday RTH)

- KC: 8× Gemini 60s timeout retries
- NG: 10× retries + 3 ERROR-level cycle failures
- SB: 10× retries + 1 ERROR Council timeout
- ZW: 1×

## First-seen + spread

- 2026-05-11 07:32:36 UTC — SB pre-RTH first timeout
- 2026-05-11 07:35:22 UTC — first Council 300s timeout (SB emergency cycle)
- Pattern then recurs through the morning/early afternoon across all 4 engines

## Likely origin

PR #2075 (merged 2026-05-10) — `fix(router): remove dead-letter Gemini-retry fallback in _route_call (#2069)`. #2069's fix removed silent agent-dropping; this is the post-fix manifestation of the same upstream Gemini latency, now visible as loud emergency-cycle failures.

[...]

PR #2239, opened 2026-05-12
feat(abstention-monitor): BULL_DRIFT / BEAR_DRIFT direction-anchor detection (#2234 — Option A)

rozavala:
## Summary

Ships **Option A** from #2234 — symmetric direction ceilings — as the minimum-viable mirror of the existing NEU ceiling code. The abstention monitor now catches three families of persona-locking pathology, not just one:

- ``NEUTRAL_DRIFT`` (existing) — persona returns NEUTRAL regardless of input
- ``EXCLUSION_DRIFT`` (existing) — persona absent from cycles
- **``BULL_DRIFT`` / ``BEAR_DRIFT`` (new)** — persona anchored on one direction regardless of input

The motivation, from the 2026-05-12 thread: the post-#2097 sentiment results look great today (NG 5/5 BULL, ZW 7/9 BEAR), but we have no automated check that the directional firing keeps *moving* as the input data moves. The eval-harness 39/39 PASS is a snapshot, not a multi-week responsiveness check. This PR closes that gap for the snapshot — Options B (input-output correlation) and C (Shannon entropy) from the issue body are deferred.

## Threshold defaults — and why 0.85 (vs 0.40-0.70 for NEU)

The asymmetry is intentional. Legitimate trending markets produce sustained one-direction firing as *correct behavior* — a commodity in a 90-day drawdown will have its bearish specialists firing BEAR 70-80% of the time. The 0.85 default is the threshold for "this is persona-locking, not trend-following."

Per #2234's recommendation, tune downward once we have the post-#2161 / #2200 / #2210 baseline settled (~2 weeks of production data). The issue stays OPEN for that calibration plus Options B/C.

## Backward compat

- ``compute_neutral_rates`` function name preserved (return shape strictly grows: ``bull_pct``, ``bear_pct``, ``n_bullish``, ``n_bearish`` added).
- ``detect_breaches`` gains four optional kwargs with sensible defaults — existing callers that pass only NEU + ABSENT ceilings continue to work.
- Legacy breach records without ``breach_type`` render with the NEU label in the Pushover formatter (existing behavior preserved).

## Test plan

- [x] 9 new tests in ``tests/test_abstention_monitor.py``:
  - BULL/BEAR drift fires above ceiling (per direction)
  - Default ceiling applies when per-agent override absent
  - ``min_cycles`` guards direction ceilings same as NEU
  - Strict ``observed > ceiling`` (not ``>=``)
  - Backward compat: rates dicts without ``bull_pct``/``bear_pct`` don't raise
  - ``compute_neutral_rates`` emits the new keys
  - Pushover formatter renders direction labels distinctly (e.g., BULL_DRIFT line says "BULL" not "NEU")
  - Legacy breach records render with NEU label
- [x] 1 updated: ``test_run_abstention_monitor_per_ticker`` now correctly asserts that 100%-one-direction synthetic data trips BULL/BEAR drift. Pre-#2234 it was asserting these were NOT flagged — that was the gap.
- [x] Full ``pytest tests/`` green: 2742 passed, 1 skipped.
- [ ] Production verification: with the 0.85 default, no current specialist should trip at deploy (today's most-extreme cases — NG sentiment 100% BULL, SB sentiment 100% BULL — have n < 20 so won't fire yet). By next week if the persona is properly responsive, percentages will have moved off 100% as the underlying data shifts. If they don't, we'll see BULL_DRIFT / BEAR_DRIFT Pushover alerts — which is the correct behavior.

## What stays open under #2234

- Threshold calibration after the ~2-week baseline.
- Option B (input-output correlation against per-specialist proxies) — needs explicit per-specialist proxy registry.
- Option C (Shannon entropy of direction distribution) — clean but adds operator-interpretation complexity over A.

Closes the snapshot half of #2234.


rozavala:
## MECE Review — Approved (label applied)

(Self-approval blocked; verdict captured as comment + `Ready to be merged` label.)

**Value (HIGH):** Closes the symmetric observability gap. NEUTRAL_DRIFT + EXCLUSION_DRIFT alone miss the third persona-locking shape (always-BULL / always-BEAR). With sentiment now firing directionally post-#2097, this is precisely the responsiveness check we lacked.

**Risk (LOW, controlled):**
- **Symmetry verified.** BULL_DRIFT (`abstention_monitor.py` L241-258) and BEAR_DRIFT (L260-274) are structural mirrors — same `n_cycles >= min_cycles and "X_pct" in stats` guard, same ceiling lookup, same `common` payload. Pushover formatter parametrised by `_BREACH_TYPE_LABEL` so future directions extend the map without render-code changes.
- **Rate denominator correct.** `bull_pct = 100 * n_bullish / n_valid` (L138) — same denominator as `neu_pct`, excludes absences. EXCLUSION_DRIFT keeps its independent `total_cycles` denominator.
- **Bootstrap safe.** Empty window short-circuits at L121-122. `total < min_cycles` guard at L194 covers all four breach checks. `test_detect_breaches_direction_ceilings_skip_low_volume` verifies 5 cycles 100% BULL with min_cycles=20 produces zero breaches.
- **Defaults sensible and symmetric.** Both `max_bullish_rate` and `max_bearish_rate` default to `0.85`. Intentional asymmetry vs NEU (0.40-0.70) well-justified inline: legitimate trending markets produce sustained one-direction firing as correct behavior. Tune-after-baseline plan documented in `_comment_2234`.
- **Backward compat preserved.** Function name `compute_neutral_rates` kept; return shape strict superset; new `detect_breaches` kwargs default empty/0.85. `test_detect_breaches_direction_ceilings_skip_when_pct_absent` covers legacy rate dicts; `test_format_breach_alert_legacy_breach_without_type_renders_as_neu` covers replay of pre-#2234 persisted reports.
- **Per-engine isolation preserved** (`test_run_abstention_monitor_resolves_per_engine_overlay` still green).

**Prior art:** Extends #2147 → #2198 → #2204 → #2213; #2234 is the authorizing umbrella that explicitly recommended Option A first. Correctly leaves #2234 open for threshold calibration plus Options B/C.

**Tests:** 33/33 pass locally (0.86s). 9 new tests cover BULL+BEAR fires, default fallback, min_cycles guard, strict `>` semantics, backward compat, new-key emission, integration (KC sentiment 100% BULL → BULL_DRIFT; NG macro 100% BEAR → BEAR_DRIFT), Pushover label distinction, legacy NEU fallback. PR body reports full `pytest tests/` green (2742 passed, 1 skipped).

PR #2251, opened 2026-05-13
fix(notifications): correct send_pushover_notification import path (#2246)

rozavala:
Closes #2246 (HIGH severity). One-line correction at two lazy-import sites + 6-test regression guard.

## Summary

`send_pushover_notification` is defined in **top-level** `notifications.py:29` — not in `trading_bot/notifications.py` (which only defines `NotificationTier` + `get_notification_tier`). Two lazy-import sites used the wrong path:

- `trading_bot/compliance.py:791` (compliance fail-closed Pushover alert)
- `trading_bot/master_orchestrator.py:52` (engine-dead Pushover alert)

Both imports are wrapped in `except Exception: pass`, which silently swallows the `ImportError` — leaving the operator with **no Pushover** on two high-severity paths:

- Engine crashes past `max_restarts` (capital-protection signal)
- Compliance veto with fail-closed LLM parse (decision-quality signal)

The operator thinks notifications are wired; they aren't.

## Fix

```diff
-from trading_bot.notifications import send_pushover_notification
+from notifications import send_pushover_notification
```

The eager-import sites elsewhere (`order_manager.py`, `order_close.py`, `drawdown_circuit_breaker.py`, etc.) already use the correct top-level path — only these two lazy imports had drifted.

## Regression guard

`tests/test_send_pushover_notification_import_paths_2246.py` (6 tests):

- Each known site uses `from notifications import …` (parametrized)
- Each known site does NOT re-introduce `from trading_bot.notifications` (separate inverse assertion, so the failure message points directly at the regression mode)
- Top-level `notifications.send_pushover_notification` is importable (catches a future move)
- `trading_bot.notifications` does NOT define the symbol (sanity-check the bug premise — if the alternative fix from #2246's body is adopted, this test fails and prompts updating the call sites)

Source-grep rather than monkey-patched-import because the bug is a static import path — runtime behavior depends only on the literal string in the source.

## Test plan

- [x] 6 new tests pass
- [x] **Full suite passes: 2750 passed, 1 skipped in 7:41**
- [ ] CI gate passes on PR

## Risk

Low. Two-character path correction on lazy imports that are reached only on critical incidents (engine-dead, compliance fail-closed). Before this fix the imports raise ImportError and the surrounding `except` swallows it; after the fix they resolve cleanly and the Pushover actually fires. The change is operationally lossless except in the failure mode we want to recover.

## Refs

- #2246 — origin issue (HIGH severity)
- PR #2243 — where the bug was discovered during mypy scope review

https://claude.ai/code/session_01EmBXjcsngyG2oL99oPLhjC

---
_Generated by [Claude Code](https://claude.ai/code/session_01EmBXjcsngyG2oL99oPLhjC)_


rozavala:
**Review: ✅ Ready to merge — critical HIGH-severity bug fix, correct and safe.**

## What was verified

**Root cause confirmed on `main`:** Both broken import sites are present in the current codebase:
```
trading_bot/compliance.py:791    from trading_bot.notifications import send_pushover_notification
trading_bot/master_orchestrator.py:52  from trading_bot.notifications import send_pushover_notification
```
`trading_bot/notifications.py` only exports `NotificationTier` and `get_notification_tier`. `send_pushover_notification` lives exclusively in top-level `notifications.py`. Both sites are wrapped in `except Exception: pass`, so the `ImportError` is silently swallowed — the operator receives **no Pushover** on engine-dead and compliance fail-closed paths. All other callers (`order_manager.py:20`, `order_close.py:39`, `drawdown_circuit_breaker.py:22`) already use the correct `from notifications import` path.

**Diff:** Two one-line changes — only the import string, no logic altered. The lazy-import pattern means these code paths execute only on critical incidents, so there is no risk of accidental regression during normal trading operation.

**Regression tests:** 4 static-grep tests (the core regression protection) pass cleanly. The 1 test that does a live import (`test_send_pushover_notification_resolves_from_top_level_module`) fails in the stripped worktree environment due to `requests` not being installed — not a code defect. CI with full deps passes green. Full suite reported in the PR body (2750 passed) confirms.

**Safety review:** `compliance.py` is safety-critical per CLAUDE.md. This change makes a previously silently-broken path work correctly — strictly safer than before. No execution logic is altered; only a broken import string is corrected.

---

PR #2287, opened 2026-05-13
feat(iv-rank): persist per-cycle iv_rank observations + 1Y backfill (#2186)

rozavala:
## Summary
- New `trading_bot/iv_rank_logger.py` — best-effort append-only logger writing `data/{ticker}/iv_rank_observations.csv`
- Hook in `regime_engine._get_current_regime_and_iv` emits one `source='runtime'` row per invocation (with 300s per-ticker dedupe)
- New `scripts/backfill_iv_rank_observations.py` — one-shot operator-run script that backfills from 1Y of daily IBKR IV history
- 9 new tests pinning behavior + 2818 existing tests still pass

## Why

Per [my #2186 triage comment](https://github.com/rozavala/real_options/issues/2186#issuecomment-4433392561): the per-commodity `condor_iv_rank_breach` tuning experiment is blocked because `iv_rank` is computed at runtime in `regime_engine._get_current_regime_and_iv` but **never persisted anywhere**. With 4 active commodities and effectively N≈3 distinct observed values in 7 days of logs, the experiment's "≥60 trading days per commodity" acceptance criterion has no source of data to compute percentiles over.

This PR is the prerequisite. It does **not** ship per-commodity threshold values — those still wait on the data accumulating. It just makes the experiment runnable in ~3 months once 60 days of runtime observations stack up, OR runnable today once the operator runs the backfill against IBKR's 1Y history.

### Runtime logger (`trading_bot/iv_rank_logger.py`)
Append-only CSV at `data/{ticker}/iv_rank_observations.csv`. Schema:
```
timestamp,iv_rank,iv_percentile,current_iv,regime,source
```
- In-memory per-ticker dedupe at 300s — exit-validation can call `_get_current_regime_and_iv` multiple times per cycle (once per IC position), and we don't want to bloat the CSV with redundant rows from the same cached 1Y window.
- Write failures log a warning and return `False`; never propagate. The exit-validation hot path cannot afford a telemetry-induced exception.
- Backfill writes bypass the dedupe window — their timestamps are historical and shouldn't be suppressed by an in-memory cache entry from an unrelated runtime call.

### regime_engine hook
After the existing metric+regime computation in `_get_current_regime_and_iv`, emit one `source='runtime'` row. Pulls `iv_percentile` and `current_iv` from the already-fetched `metrics` dict (zero extra IB calls). Logger failures are swallowed; the `(regime, iv_rank)` return contract is byte-identical.

### Backfill script
Standalone — not a migration, because the migration runner runs after `trading-bot.service` is stopped and an IB connection is needed for `reqHistoricalDataAsync`. Operator runs once with IB Gateway up:
```
python scripts/backfill_iv_rank_observations.py
```
- Defaults to `config.active_commodities`; `--tickers KC,NG` to scope.
- `--dry-run` prints counts without writing.
- Idempotent: skips a ticker whose observations file already has ≥60 rows. Re-run to safety after a partial failure.
- Uses `clientId=199` by default (overridable via `IV_RANK_BACKFILL_CLIENT_ID` env) to avoid colliding with running engine processes.

For each historical day `d`, computes `_compute_iv_rank_and_percentile(closes[:d], close[d]_as_fraction)` — exactly the same helper the runtime uses. Scale (percent vs fraction) is detected once on the full series for stable interpretation across the backfilled range.

## Caveat: early-day approximation

The earliest ~30 days in each backfilled series get a shorter trailing window than the runtime definition (60-day minimum). `_compute_iv_rank_and_percentile` emits `'N/A'` for those rows rather than approximating, so the experiment will simply ignore them. Percentile interpretation is full-window-accurate only for the freshest ~7-12 months — comfortably above the 60-day acceptance threshold from #2186.

## Test plan
- [x] `pytest tests/test_iv_rank_logger.py` — 7 passed
- [x] `pytest tests/test_regime_engine.py` — 23 passed (incl. 2 new integration tests)
- [x] `pytest tests/` (full suite) — 2818 passed, 1 skipped
- [x] `ruff check` clean on all new files
- [ ] After merge: operator runs `python scripts/backfill_iv_rank_observations.py` once with IB Gateway up to seed historical observations
- [ ] After ~60 days of runtime observations (or immediately post-backfill): the #2186 tuning experiment becomes runnable as designed

## Out of scope

- Picking per-commodity `condor_iv_rank_breach` values — that's the experiment in #2186 itself, which still waits on this data
- Dashboard surfacing of the new CSV — analysts can read the CSV directly for now; dashboard integration is a separate follow-up if useful
- Changing the global `exit_logic.condor_iv_rank_breach` default

Refs #2186, #2039 (parent).


rozavala:
Thanks for the careful review — the path bug was a real one and the diagnosis was exactly right. Fixed at `f0fbc17ca` (force-pushed; PR is fresh and unreviewed before this push).

**What changed:**

- `iv_rank_logger._observations_path(data_dir)` no longer joins a ticker segment — accepts the already-per-ticker `data_dir` and writes `data_dir/iv_rank_observations.csv`. Matches the canonical pattern in `utils.log_council_decision`. Logger still takes the `ticker` arg for the dedupe key, just doesn't use it in the path.
- `regime_engine` hook unchanged at the call site — still passes `config['data_dir']` (already `data/<ticker>` in prod).
- `backfill_iv_rank_observations.py` now resolves `base_data_dir = resolve_base_data_dir(config)` and joins per-ticker inside the loop. Fixes the secondary bug where multi-ticker invocations would have all routed under the first ticker's dir.
- Module docstring + `_last_write_ts_by_ticker` now have explicit comments documenting why per-ticker keying makes the module-level state safe for multi-engine-single-process callers (your non-blocking observation #1).
- Moved `parse_iv_to_fraction` import to the top-level imports alongside the other `ib_interface` imports (non-blocking #2).

**New regression pin:** `test_observations_path_does_not_nest_under_ticker` asserts directly that `_observations_path(per_ticker_dir)` returns `per_ticker_dir/iv_rank_observations.csv` and NOT `per_ticker_dir/KC/iv_rank_observations.csv`. The integration test `test_get_current_regime_and_iv_writes_observation_row` was also updated to use a production-shaped `data_dir` (`tmp_path/KC`, not bare `tmp_path`) and explicitly asserts the double-nested path does NOT exist.

Full suite (2819 passed, 1 skipped, ~7m22s) green at the new HEAD.

PR #2307, opened 2026-05-14
fix(var): widen IV fetch timeout 8s→12s + log first-ready latency (#2283)

rozavala:
## Summary

- Bump default ``compliance.iv_fetch_timeout`` 8.0 → 12.0s
- Log time at which the first ticker resolved alongside the timeout warning to distinguish partial-fail (slow-tail individual contracts) from all-fail (off-RTH / no subscription)
- Config override path unchanged — operators can still tune via ``compliance.iv_fetch_timeout``

## Evidence from production logs

Two distinct failure modes observed in ``logs/orchestrator_*.log``:

| Shape | Examples | Likely cause |
|---|---|---|
| Partial timeouts during RTH | 56× 2/6, 32× 2/19, 30× 10/19, 24× 12/19 | Slow-tail individual IB streams; median 8-11s |
| All-fail off-RTH | 32× 6/6, 12× 19/19 (mostly 21:11+ UTC / 00:30+ UTC) | No streaming-IV subscription outside RTH |

Issue #2283 reports 6/14 VaR runs on 2026-05-13 fully degraded (19/19 fallback). That's the all-fail shape — widening doesn't *help* but doesn't *hurt* (the poll exits early once every ticker resolves). The 8 remaining partial-fails almost certainly benefit from 12s.

## What the new log line tells you

`IV fetch timed out after 12.0s — 5/19 tickers still pending (first ready at 9.3s)` — partial-fail, slow tail. Widen further if the gap between first-ready and timeout is consistently > 3s.

`IV fetch timed out after 12.0s — 19/19 tickers still pending (first ready at never)` — all-fail, no subscription. Widening won't fix this; needs market-open gating, which is left to a follow-up after we have enough data.

## Test plan

- [x] `pytest tests/test_var_calculator.py -x -q --timeout=60` — 49 passed (47 pre-existing + 2 new)
- [x] Test `test_iv_polling_timeout_default_is_12s` pins the new default
- [x] Test `test_iv_polling_logs_first_ready_when_all_fail` pins the new instrumentation log
- [x] Existing tests unchanged — the override path (``compliance.iv_fetch_timeout`` in config) still wins

## Out of scope

- Per-ticker fast/slow split — speculative until we have first-ready data on it
- Per-commodity market-open gating for the all-fail case — separate, larger change
- Live-IV caching pattern analogous to #2173 — separate experiment

Closes #2283.


rozavala:
## Verdict: Ready to merge

**Value (MECE 1)**: Closes #2283. Production log evidence cited in the PR shows the partial-fail shape (5-12 of 19 stragglers, median 8-11s) is the dominant timeout pattern during RTH — a 50% timeout widening (8 → 12s) captures the slow tail without compromising fail-fast behavior. The new "first ready at X / never" log line is genuine observability: it lets the next round of tuning distinguish partial-fail (widen further) from all-fail off-RTH (gate on market-open, separate work). Proportionate fix.

**Risk (MECE 2)**: VaR-adjacent (`trading_bot/var_calculator.py`); compliance feeds off VaR; this is safety-relevant. Mitigations verified:

- The poll loop exits early once every ticker resolves OR all fail, so the worst-case cycle latency increase is bounded by `min(slowest_resolved + poll_interval, 12s)` — a 4s ceiling above today's behavior.
- Config override path (`compliance.iv_fetch_timeout`) preserved — operators can revert per-deployment without a code change.
- Two new tests pin both invariants: `test_iv_polling_timeout_default_is_12s` (default value), `test_iv_polling_logs_first_ready_when_all_fail` (the all-fail log shape). 49/49 var_calculator tests pass on this worktree.
- 33/33 compliance + risk_management tests pass.

**Prior art (MECE 3)**: #2100 raised the original timeout concern; #2118 fixed reporting; #2283 is the residual root-cause work this PR addresses. Not re-litigated.

**Trajectory (MECE 4)**: The first-ready-at instrumentation is forward-looking — it generates the empirical data needed to make the next tuning decision (widen further vs. market-open gating). Better than a one-shot timeout bump that flies blind.

LGTM.

PR #2318, opened 2026-05-14
fix(backfill-iv-rank): per-ticker outcome + non-zero exit on partial fail (#2315)

rozavala:
## Summary

- ``TickerStatus`` tri-state: ``SUCCESS / SKIPPED / FAILED``. ``SKIPPED`` (file already at threshold) counts as success for exit-code purposes.
- ``_backfill_one`` returns ``(status, rows)``. Every former silent-fail path (insufficient bars, missing profile, no active futures, exceptions) now surfaces as FAILED.
- ``_run`` returns the per-ticker status dict and wraps each call in try/except so one ticker's exception doesn't strand the rest.
- ``main`` exits 1 if any requested ticker is FAILED — the migration framework will retry on the next deploy.

## Why the change matters

PR #2313 wired the backfill into the migration framework. Pre-#2315, the framework saw a zero exit even when the script silently fell through on a ticker (the typical KC off-RTH shape: ICE RTH 04:15-13:30 ET is narrower than NG/SB/ZW; an off-RTH first deploy returns no historical bars / Error 200). The migration got marked applied, KC's observations file stayed empty, and the framework had no signal to retry.

Now: any ticker the operator asked for that doesn't end the run at SUCCESS or SKIPPED forces a non-zero exit. The migration framework retries; eventually every ticker's RTH window is open during a deploy and the migration completes cleanly.

## Side fix

The pre-existing `if profile is None` check was dead code — `get_commodity_profile` raises `ValueError` for unknown tickers rather than returning None. The script would crash with an uncaught exception, stranding any tickers later in the list. Wrapped with try/except so an unknown ticker is surfaced as FAILED for that ticker and the script proceeds to the next.

## Test plan

- [x] `pytest tests/test_backfill_iv_rank_observations.py -v --timeout=60` — 9 passed
- [x] Tests cover all five `_backfill_one` return-shape branches (success, skipped, fail-no-bars, fail-no-profile, fail-no-futures)
- [x] `_run` aggregation + exception-to-FAILED behavior pinned
- [x] `main` exit-code contract pinned for both partial-fail and all-good cases

Closes #2315.


rozavala:
Closing as duplicate of #2317, which merged just before this PR was opened.

Side-by-side, #2317 (merged) and #2318 (this) are the same scope and same shape:
- `(status, rows)` return from `_backfill_one`
- Status tri-state `success / skipped / failed`
- `main()` exits 1 on any failed ticker → migration framework retries
- Both close #2315

That said, this PR caught **three things #2317 missed** that I'll port in a follow-up:

1. **Real bug**: `get_commodity_profile` raises `ValueError` for unknown tickers — does not return `None`. #2317 kept the existing `if profile is None` check as a `'failed'` return, but that branch is unreachable. An unknown ticker still crashes `_run` and strands every later ticker. Port: replace with `try/except ValueError`.
2. **Broader defense**: #2317's `try/except` is only around `_fetch_dated_iv_history`. This PR wraps the whole `_backfill_one` call in `_run`'s loop, so any other raise (e.g., `log_observation` writing to a missing dir) doesn't strand subsequent tickers. Port: broaden the catch site.
3. **Better tests**: 9 tests covering every `_backfill_one` return-shape branch vs #2317's 6 aggregator-only tests. Port: add the missing branch tests.

I won't try to rebase this branch — it's built off pre-#2317 `main` and would re-rewrite every line of `_backfill_one`. The follow-up will be a clean diff on top of merged `main`.

Thanks for catching what I missed. Closing — follow-up PR incoming.

---

PR #2401, opened 2026-05-14
fix(weighted-voting): VaR dampener skips non-OK cached results (#2393)

rozavala:
## Summary

Closes #2393. Makes the VaR confidence dampener in `weighted_voting.calculate_weighted_decision` respect `cached_var.last_attempt_status` — non-OK readings (`DEGRADED` / `NO_DATA` / `UNAVAILABLE` / `FAILED`) bypass the dampener instead of damping council confidence with known-bad input.

## Why

Pre-#2393, the dampener consumed `var_95_pct` regardless of status. The #2100-family fixes (#2100 / #2118 / #2120) added the status field but never wired the dampener to read it — so a `DEGRADED` result (priced entirely from `profile.fallback_iv` because option markets were closed or IV fetch returned empty) could halve council confidence on the next emergency cycle.

Today's logs (2026-05-14, post-close window):

```
21:38:08,913 — VaR reliability degraded: 16/16 positions used fallback IV
21:38:10,832 — VaR DEGRADED: All 16 option position(s) priced with fallback IV
```

Any emergency-cycle that fires next reads that DEGRADED result via `get_cached_var()` and dampens confidence on a utilization derived from guess-vol scenarios. Same mechanism produced the original 187% pre-market spike that motivated parent #2065.

## Why only Option B from the issue

Issue #2393 proposed two layers: Option A (gate `recompute_var_state` on market-open) and Option B (status-aware dampener). After tracing the post-close audit cycle, **Option A is wrong**: the post-close recompute is *intentional* — it snapshots the day's risk state for overnight planning and pre-open trigger gating. Skipping it during market-closed defeats its design purpose. The cleaner architecture is "let the recompute run, correctly stamp DEGRADED when it can't get live data, and have downstream consumers respect the stamp." Option B is exactly that.

Option B also covers cases A wouldn't have: mid-session IV-fetch failure, `ib.portfolio()` errors, generic compute exceptions. The breadth is the load-bearing piece.

## Changes

- `trading_bot/weighted_voting.py:861-895` — `cached_ok = cached_var.last_attempt_status == "OK"` guard around the dampener block. Uses `getattr(cached_var, 'last_attempt_status', 'OK')` for backward compat with legacy results on disk. Adds a DEBUG breadcrumb when a non-OK status is skipped so forensics can correlate.
- `tests/test_var_dampener_status_aware.py` (new) — 8 cases covering all four non-OK statuses, the OK regression path, legacy-without-field backward compat, no-cached-var, and below-threshold-but-DEGRADED.

## Test plan

- [x] `pytest tests/test_var_dampener_status_aware.py` → 8 passed.
- [x] Related-modules sweep (`test_agent_validation.py` + `test_var_calculator.py` + `test_signal_flow.py` + `test_guardrails.py` + `test_row_builder_symmetry.py`) → 98 passed.
- [x] Full suite (`pytest tests/`) → 19 failures, **all pre-existing on `origin/main`** (same set documented in #2376's PR body — `test_schedule_*`, `test_sentinel_circuit_breaker_throttle`, `test_weekend_behavior`, `test_order_manager::TestValidateICWingsLiquidity` order-pollution, etc. — none touch `weighted_voting.py` or the new test file).
- [ ] DEV observation after merge:
  - Zero `0.50 FLOOR` dampener log lines whose preceding `VaR=…` line was stamped `DEGRADED` (today's logs would have shown several such pairings).
  - Legitimate FLOOR firings during in-session high-utilization continue to fire as today (regression guard — covered by `test_dampener_applies_when_status_ok`).

## Rollback

If the un-dampened paths during DEGRADED windows turn out to be too permissive in practice, the gate is one boolean — revert the `cached_ok` guard or invert it. No state-shape changes; no migration required.


rozavala:
## Re-reviewed — fixed, Ready to be merged

Commit `b308fb3ac` restructured the guard exactly as suggested:

```python
if cached_var is not None:
    cached_status = getattr(cached_var, 'last_attempt_status', 'OK')
    if cached_status != 'OK':
        logger.debug(...)
    else:
        var_limit_pct = ...
        # dampener block
```

Single `is not None` narrowing — mypy strict can now prove `cached_var.var_95_pct` safe inside the OK branch. The legacy-missing-field path (covered by `test_dampener_applies_when_status_field_missing_legacy`) is preserved via the `getattr(..., 'OK')` default. The DEBUG log for non-OK rejections kept for forensics.

### Verified

- CI green: Ruff F821 + Mypy strict-scoped both passed at 00:01:38 UTC
- Worktree test run: **90/90 pass** across `test_var_dampener_status_aware.py` (8 cases) + `test_agent_validation.py` (incl. `TestWeightedVotingForceAbstain` regression check) + `test_var_calculator.py` (status-assignment tests)
- Real code restructure — tree hash changed, not a no-op fix-commit

Approving — original MECE assessment still holds: high-value bug fix for the post-#2065 reliability layer, low-risk (DEBUG log only for non-OK; runtime behavior preserved for the common OK path), no prior-art conflict, lands on the signal-quality + risk-management workstream alongside #2407.

rozavala:
## Re-review — previously-flagged issue resolved ✅

The `claude-fix` commit (`b308fb3ac`) correctly addressed the mypy strict `union-attr` error. Verified:

**Fix applied exactly as suggested:**
```python
if cached_var is not None:           # ← single is-not-None narrows VaRResult|None → VaRResult
    cached_status = getattr(cached_var, 'last_attempt_status', 'OK')
    if cached_status != 'OK':
        logger.debug(...)             # non-OK rejection is logged for forensics
    else:
        var_utilization = cached_var.var_95_pct / var_limit_pct ...  # mypy-safe ✓
```

**CI (post-fix commit):**
- `Mypy (strict, scoped)` → ✅ success
- `Ruff F821` → ✅ success

**MECE verdict:**
1. **Value**: Real production bug — DEGRADED VaR (all 16 positions priced from fallback IV during post-close audit) was feeding the confidence dampener and could halve council confidence on the next emergency cycle. Demonstrated with today's (2026-05-14) logs. The `#2100` family added the status field to flag exactly this; this PR wires the downstream consumer to respect it.
2. **Risk**: Low. The change is a narrowing guard on a safety-critical file. When VaR is DEGRADED the dampener stays at 1.0 — more permissive than before, but correct: the VaR was unreliable, so suppressing council confidence based on it amplified known-bad input. The `getattr` fallback correctly preserves legacy-on-disk behavior.
3. **Prior art**: Natural follow-on to the `#2100` family; no rejected prior attempts.
4. **Trajectory-fit**: Squarely within the `#2100`-family VaR reliability workstream. Tracked issue (#2393). Option A/B decision is well-reasoned and documented.

All 8 tests cover the full status matrix (OK, DEGRADED, NO_DATA, UNAVAILABLE, FAILED, legacy-missing-field, no-cached-var, below-threshold-DEGRADED). Marking **Ready to be merged**.

---

PR #2323, opened 2026-05-14
feat(dashboard): Cycle Viewer — Slack-like chat replay per cycle

rozavala:
## Summary

Adds a new dashboard page (**11_Cycle_Viewer**) that replays any trading cycle as a Slack-like chat timeline: sentinel trigger → market snapshot → every analyst report → full Permabear/Permabull debate transcript → Devil's Advocate → master decision → compliance → execution events → exit. The motivating use case is forensic debugging — "why was this signal generated, why did compliance veto, why did this trade lose money".

Also closes a real forensic gap: the debate text, Devil's Advocate output, and full sentinel payloads were built in memory each cycle and **discarded after compliance** — only `debate_order` + `debate_models` survived in `council_history.csv`. This PR persists them in per-cycle JSON artifacts.

## Three stacked commits

1. **`feat(cycle-artifacts): persist debate, devils-advocate, sentinel payload per cycle`**
   - New `trading_bot/cycle_artifacts.py` — `CycleArtifactBuilder` accumulator + atomic JSON write (`tempfile` + `os.replace`).
   - Wired into `agents.py` (`run_debate` / `decide` / `run_specialized_cycle` take optional `artifact_builder`), `emergency_cycle.py` (build at cycle_id generation, flush in inner `finally`), `signal_generator.py` (per-contract builder, flush in new `finally`).
   - Files live at `data/{ticker}/cycle_artifacts/{cycle_id}.json`.
   - Empty-flush short-circuits — gate-blocked cycles that bail before any `record_*` call don't litter the dir.

2. **`feat(dashboard): add Cycle Viewer page`**
   - `dashboard_cycle_viewer.py` — pure-Python loaders/joins/timeline builder. No Streamlit imports → fully unit-testable.
   - `pages/11_Cycle_Viewer.py` — Streamlit shell with sentiment-colored picker, message-kind → renderer registry, query-param routing (`?cycle_id=…`), "Download as Markdown" export.

3. **`feat(dashboard): cross-page deep links into Cycle Viewer`**
   - `cycle_link()` helper in `dashboard_utils.py`.
   - Deep links from `3_The_Council` (inline link in selected-decision header), `6_Signal_Overlay` (LinkColumn in Raw Signal Log), `10_The_Funnel` (LinkColumn in per-trade outcome table).
   - Brier (#7) intentionally not wired — its rows are agent-aggregate, not per-cycle.

## Why JSON artifacts vs. more CSV columns

[...]

PR #2425, opened 2026-05-15
feat(cycle-viewer): IV snapshot in the market-snapshot section

rozavala:
## Summary

Closes **#2423**. Adds implied-volatility context to each cycle's snapshot, joined from `data/{ticker}/iv_rank_observations.csv`. There's no `cycle_id` on the IV rows — observations are timestamped only — so the join picks the most recent observation **at or before** the cycle's decision time (can't predict the future).

Renders as one line below the existing 4 metric columns:

> **IV** 35.5% · **rank** 28 _(low-range)_ · _obs 3 min before · runtime_

The rank-band label (low / mid / high) helps an operator quickly read *"this council picked a LONG_STRADDLE when IV rank was 18 (low-range)"* — strategy choices like IRON_CONDOR / LONG_STRADDLE are IV-conditional, so seeing the IV the council saw is high-leverage context that wasn't previously visible.

## Why observation-source tagging matters

| Source | Frequency | Carries |
|---|---|---|
| `runtime` | Per cycle (sparse, only when one fires) | Full `iv_rank` + `iv_percentile` + `current_iv` |
| `backfill` | Daily | `current_iv` only (`iv_rank`/`iv_percentile` are `N/A`) |

Stale observations (`obs 2 days before`) are still shown — better than nothing — but the age + source tag tells the operator exactly how trustworthy the value is for *this* cycle.

## Implementation

- New **`load_iv_snapshot(ticker, decision_time, base_dir)`** pure function in `dashboard_cycle_viewer.py`. Tolerates `"N/A"` on backfill rows, naive vs tz-aware decision times (treated as UTC), decision times that pre-date all observations (returns None).
- `build_timeline` calls it once per cycle and attaches the result to the snapshot payload's `iv_snapshot` key.
- `render_snapshot` in `pages/11_Cycle_Viewer.py` adds the IV chip below the existing metric columns; skipped silently when the iv_snapshot dict is None or empty.
- Small helpers: `_humanize_age(seconds)` and `_iv_rank_label(rank)` for the chip text.

## Test plan

- [x] **+9 new tests** covering missing file, latest-before-decision selection (NOT after, no future peeking), `N/A` iv_rank handling on backfill rows, naive decision time treated as UTC, exact-match age 0, None decision time, end-to-end timeline payload propagation
- [x] **104/104** cycle_viewer + cycle_artifacts tests pass locally
- [x] Ruff clean repo-wide
- [ ] CI green
- [ ] Manual on DEV after deploy: open any recent cycle, verify the snapshot expander now has the IV chip line below the metrics. Both runtime cycles (full chip) and backfill-only periods (IV%-only chip) should render gracefully.

https://claude.ai/code/session_01GHBsk95maucGJAJaGZpigh

---
_Generated by [Claude Code](https://claude.ai/code/session_01GHBsk95maucGJAJaGZpigh)_


rozavala:
## Ready to be merged

Closes **#2423** (the IV-at-order-time issue I filed off the #2421 follow-up call-out, less than an hour ago).

### Design notes

The implementation goes *beyond* the merge_asof + 5min-tolerance shape I sketched in #2423 — instead it uses **strict "at or before" semantics**: pick the most recent IV observation whose timestamp ≤ decision_time. That's more conservative than nearest-match (no future-peeking) and the right call. Stale observations are still shown but tagged with their age so operators can grade trustworthiness.

The runtime-vs-backfill source tagging is a smart addition I didn't anticipate in the issue:
- `runtime` rows (per-cycle, sparse): full `iv_rank` + `iv_percentile` + `current_iv`
- `backfill` rows (daily, dense): `current_iv` only — `iv_rank`/`iv_percentile` are `N/A`

Showing `obs 3 min before · runtime` vs `obs 2 days before · backfill` lets operators distinguish "this IV is exactly what the council saw" from "this is the nearest snapshot we have, decision-time IV may differ." That's the load-bearing nuance.

### Code-shape verification

- **`load_iv_snapshot(ticker, decision_time, base_dir)`** is a pure function — easy to test, easy to reason about
- Handles: missing file, `N/A` iv_rank on backfill rows, naive-vs-tz-aware decision times (UTC-coerce), pre-history decision times (returns None)
- `build_timeline` calls it once per cycle; `render_snapshot` renders the chip only when non-None
- Skipped silently when the snapshot is None/empty — graceful no-data fallback

### Tests verified

- **104/104 cycle_viewer + cycle_artifacts tests pass locally** in worktree (9 new + 95 pre-existing)
- New tests pin: missing file → None, latest-before-decision selection (with explicit NOT-after assertion), `N/A` handling, naive UTC coercion, exact-match age 0, None decision time → None, end-to-end timeline propagation
- CI green: Ruff F821 + Mypy strict-scoped at 04:38 UTC

### MECE

- **Value**: real, high for a forensic surface — IV-rank context tells the operator *why* the council picked LONG_STRADDLE vs IRON_CONDOR. Strategy choices are IV-conditional; previously this had to be cross-referenced manually against `iv_rank_observations.csv`
- **Risk**: low — dashboard-only, pure-function join, graceful no-data fallback. No trading orchestrator path touched
- **Prior art**: not re-litigating — this is the implementation of the #2423 follow-up I filed yesterday. Clean closing of the loop
- **Trajectory**: Cycle Viewer thread (#2323 → #2349/2380/2396/2402/2414/2421). Consumes the iv_rank backfill chain output (#2186 → #2287/2295/2305/2313/2317/2320/2321/2331/2333) — perfect example of trajectory-fit work that gives the backfill data a downstream consumer

Closes the #2423 follow-up the same day it was filed. The conscious-scope-split between #2421 (read what's in the funnel) and #2425 (read across CSVs) was the right call.

PR #2539, opened 2026-05-19
fix(execution): implement exposure-priority leg sorting for BAG fallback (#2348)

rozavala:
### Goal
Hardens the fallback liquidation logic for BAG (combo) orders to prevent directional exposure during partial fills. Closes #2348.

### Findings
Forensic audit of production logs revealed up to 18s gaps between leg closes when a BAG order failed. This was due to naive sequential closing without regard for market risk.

### Proposed Changes
- **Delta-Prioritized Sorting**: Legs are now sorted by absolute Black-Scholes delta (highest risk first).
- **Aggressive Trailing Walks**: Trailing legs now use halved intervals and 50% larger increments to 'chase' the market and close the loop.

### Lessons Learned
Avoided the mistake of #1640 (which focused only on speed) by anchoring the priority in Delta risk.

### Verification
- PASSED 
- PASSED


rozavala:
**Re-review: APPROVE. Both previously-flagged issues are fully resolved.**

**v2.5 scope creep reverted.** \`gh pr diff 2539 --name-only\` now lists only 3 files: \`tests/test_bag_close.py\`, \`trading_bot/contribution_scorer.py\`, \`trading_bot/order_close.py\`. None of the smuggled files remain (\`reconciliation.py\`, \`master_orchestrator.py\`, \`commodity_engine.py\`, \`scheduling.py\`, \`reconcile_and_analyze.py\`, \`reconcile_trades.py\`, \`scripts/sync_worktree.sh\`).

**NaN regression test added** at \`tests/test_bag_close.py::TestBagLegDeltaSort\` — 5 tests covering:
1. NaN IV doesn't get passed to B-S (sigma=0.3 fallback)
2. NaN IV → finite delta
3. Deterministic sort order with NaN-IV leg
4. \`delta=0.0\` from modelGreeks preserved (the \`is not None\` check)
5. Pre-fix bug reproduction

Tests use an extracted \`_resolve_leg_delta\` helper that replicates the prod logic byte-for-byte rather than importing \`order_close.py\` (avoiding the \`ib_insync\` dependency for unit testing). Minor concession but the logic match is faithful.

**NaN fix correctness preserved**: \`order_close.py:933\` has \`sigma=(ticker.impliedVol if (ticker.impliedVol and not util.isNan(ticker.impliedVol)) else 0.3)\`. Adjacent fix at line 927: \`ticker.modelGreeks.delta is not None\` (not truthy).

**#2533 contribution_scorer logic preserved** — zero-score exclusion at \`contribution_scorer.py:155-171\` intact (guards storage + trim + save inside \`if score != 0.0:\` with a debug-log else branch).

**Full test suite**: 3428 passed, 1 skipped, 3 failures — all 3 are pre-existing on main (the documented \`test_notifications_pr_e::TestMaxRiskTMSLookup\` TMS flake plus 2 \`test_var_calculator\` failures unrelated to this PR). The 7 v2.5-induced failures from the prior review are gone.

Ready to merge.

Issue #1648, opened 2026-04-17
bug: VaR calculator returns $0 at 95th percentile while $1,701 at 99th — and substitutes B-S model prices that deviate 99% from market


rozavala:
## Summary

Two related VaR calculation issues:

### 1. VaR(95%) = \$0 while VaR(99%) = \$1,701 (with 13 positions)

\`\`\`
VaR(95%)=\$0 but VaR(99%)=\$1,702 with 13 positions — tail risk exists beyond 95% confidence level.
\`\`\`

Confirmed in current \`data/var_state.json\` (computed 2026-04-17T14:31:55 UTC). The portfolio holds 13 option spread positions. A VaR of \$0 at 95% while \$1,701+ at 99% suggests the historical simulation has insufficient return observations to populate the 95th percentile, or the options delta/gamma computation is producing near-zero P&L on most scenarios.

### 2. B-S model prices deviate 99% from market prices for NG deep-OTM puts

\`\`\`
Market price for SOH7 P0.15 (0.0116) deviates 99% from B-S model (1.4974) — likely stale/averageCost. Substituting model price.
Market price for SOH7 P0.1275 (0.0029) deviates 99% from B-S model (0.4195) — substituting model price.
Market price for SON6 P0.1275 (0.0017) deviates 99% from B-S model (0.1511) — substituting model price.
\`\`\`

The VaR calculator flags the IB-reported price as stale (probably \`averageCost\` leaking in as market price) and substitutes a Black-Scholes theoretical price. However, the B-S model itself appears to have a pricing error — the theoretical prices are 100× the market price for these options, indicating a unit mismatch or IV scaling bug in the B-S formula.

## Impact

- VaR(95%)=\$0 prevents the Compliance Guardian from triggering any 95% VaR breach alerts regardless of actual portfolio risk.
- Substituting incorrect B-S prices for stale market prices adds model error that may overstate or understate position risk.

## Log Evidence

\`\`\`
2026-04-17 14:31:55 - var_calculator - WARNING - VaR(95%)=\$0 but VaR(99%)=\$1,702 with 13 positions
2026-04-13 17:55:28 - var_calculator - WARNING - Market price for SOH7 P0.15 (0.0116) deviates 99% from B-S model (1.4974)
2026-04-14 12:15:26 - var_calculator - WARNING - Market price for SOH7 P0.15 (0.0108) deviates 99% from B-S model (1.3726)
\`\`\`

## Component

\`trading_bot/var_calculator.py\`

PR #2558, opened 2026-05-21
fix(scheduling): register 3 missing RECOVERY_POLICY entries (#2553)

rozavala:
**Closes #2553.**

Three tasks scheduled in the config-driven \`session_template\` had no \`RECOVERY_POLICY\` entry in \`trading_bot/scheduling.py\`, fell through to the runtime safe default \`{'policy': 'MARKET_OPEN', 'idempotent': False}\`, and triggered a WARNING on every engine boot (4 lines per boot — one per engine). The implicit policy was wrong for two of the three.

## Entries added

| func_name | policy | idempotent | Rationale |
|---|---|---|---|
| \`run_pulse_reconciliation\` (#2544) | MARKET_OPEN | True | Light reqExecutions reconciler scheduled at 20/40/60/80% of session — session-only by design, idempotent reads of IB state |
| \`generate_eod_portfolio_pnl_task\` | ALWAYS | True | EOD P&L summary at session_close + 28min. Reporting — recover even hours later; overwrites with current state |
| \`check_catastrophe_stop_gtc_drift\` | ALWAYS | True | GTC catastrophe-stop drift check at session_close + 38min. Pure read-side check; recover regardless of market state |

The runtime default would have flagged the pulse reconciler correctly (MARKET_OPEN matches its session-only design), but marked the two EOD tasks as MARKET_OPEN too — which would skip them on any recovery outside market hours, defeating their purpose.

## Regression guard (the part that prevents recurrence)

\`tests/test_schedule_config.py\` — 3 new tests:

- \`test_recovery_policy_covers_all_default_schedule_funcs\` — hoists the runtime validator into a unit test against \`_build_default_schedule()\`.
- \`test_recovery_policy_covers_all_config_schedule_funcs\` — same coverage guarantee for the config-driven path (\`build_schedule(load_config())\`). This is the canonical path \`commodity_engine\` actually consumes at boot, so any new \`session_template\` entry without a policy entry now **fails CI** rather than printing a runtime warning that ships to PROD.
- \`test_pulse_reconciliation_policy_registered\` — pins the three specific picks above so a future cleanup that re-removes them is caught.

## Tests

- [x] \`pytest tests/test_schedule_config.py\` → **18 passed** (incl. 3 new)
- [x] Schedule-adjacent suites (test_session_schedule, test_recovery_intertask_sleep, test_optimized_prompt_monitoring_schedule, test_session_schedule_commodity_filter) → **66 passed**
- [x] \`pytest tests/\` (production-touching PR) → **3461 passed, 3 pre-existing unrelated failures** (confirmed by re-running on bare \`origin/main\` without my changes):
  - The known \`test_notifications_pr_e::TestMaxRiskTMSLookup\` TMS-notification flake from prior sessions
  - \`test_var_calculator::test_iv_unavailable_uses_fallback\`
  - \`test_var_calculator::test_option_underlying_price_from_future\`
  
  Both VaR failures fail on the B-S model-price substitution path introduced in #1648; tracked separately, not caused by this PR.
- [x] \`ruff check\` clean on both touched files

## What still surfaces after this lands

The boot-time warning about uncovered tasks will be gone. The two pre-existing VaR test failures are orthogonal and remain as separate work.


rozavala:
**Approved — clean config fix with a regression guard that prevents recurrence.**

**The fix.** 3 missing RECOVERY_POLICY entries added with per-task rationale documented inline in \`trading_bot/scheduling.py\`:

| Task | Policy | Idempotent | Why |
|---|---|---|---|
| \`run_pulse_reconciliation\` | MARKET_OPEN | True | Session-only by design; reads IB state, no transactional side effects |
| \`generate_eod_portfolio_pnl_task\` | ALWAYS | True | EOD summary writer — even hours-late, we still want it (overwrites with current state) |
| \`check_catastrophe_stop_gtc_drift\` | ALWAYS | True | Pure read-side check; safe to re-run regardless of market state |

**The real bug** wasn't just the boot-time warning. The implicit safe-default \`{'policy': 'MARKET_OPEN', 'idempotent': False}\` was **wrong for 2 of 3 entries**: the two EOD tasks would have been skipped on any recovery outside market hours, defeating their purpose. This PR fixes that bug.

**The regression guard** is the strong design move. The new test \`test_recovery_policy_covers_all_config_schedule_funcs\` hoists the runtime validator into a unit test against \`build_schedule(load_config())\` — the canonical path \`commodity_engine\` consumes at boot. Any future \`session_template\` entry without a matching policy entry now **fails CI** rather than printing a runtime warning that ships to PROD. Plus a default-schedule coverage test and a sanity pin on the 3 specific picks.

**Risk: zero.** Pure additive config. The validator was logging warnings, not blocking boot — so this is a hygiene fix that also catches a real correctness gap.

**Tests:** 3461 passed + 3 documented pre-existing failures (confirmed by re-running on bare origin/main). 18 schedule-config tests pass (incl. 3 new). 66 schedule-adjacent tests pass. CI Ruff + Mypy green.

Ready to merge.

Issue #2096, opened 2026-05-10
bug: abstention_monitor reads empty data in production cron — "0 specialist-tickers within ceiling" despite live data showing 20 breaches


rozavala:
## Summary

PR #2064 wired \`abstention_monitor\` into the session schedule. First cron run (2026-05-10 00:44 UTC, recovery path) fired correctly and persisted \`logs/abstention_monitor/2026-05-10.json\`, but reports **0 breaches** and the orchestrator log says:

\`\`\`
Abstention monitor: OK (0 specialist-tickers within ceiling, 7d window)
\`\`\`

Manual reproduction from \`/home/rodrigo/real_options/\` CWD with the same config and same data → **20 breaches** found. So the monitor IS running, but seeing no data.

## Mechanism

The "0 specialist-tickers" message tells us \`report['by_ticker'][<ticker>]['rates']\` is empty for all 4 tickers. That means \`compute_neutral_rates(df, ...)\` returned \`{}\` for all of them — which happens when \`df\` is empty.

\`_safe_read_council_history\` currently returns an empty DataFrame on:
- CSV file not found
- Read errors
- Empty file

…and the empty DataFrame propagates silently through to the "0 breaches → OK" path.

The monitor uses:

\`\`\`python
base_dir = config.get("data_dir", "data")
for ticker in active:
    data_dir = os.path.join(base_dir, ticker)
    df = _safe_read_council_history(data_dir)
\`\`\`

\`config['data_dir']\` is absent, so \`base_dir = "data"\` (relative). From the orchestrator's CWD \`/home/rodrigo/real_options/\` this resolves correctly to \`data/KC/council_history.csv\`. But the cron run is still seeing empty data.

## Likely cause

The cron task fires per-engine, and each engine has its own \`data_dir_context\` runtime where \`get_engine_data_dir() = "data/<ticker>"\`. If the orchestrator's task-invocation layer is overriding \`config['data_dir']\` to the per-engine value before the call, then:

\`\`\`
base_dir = "data/KC"  (per-engine override)
data_dir = os.path.join("data/KC", "KC") = "data/KC/KC"  ← does not exist
\`\`\`

→ all reads return empty → 0 breaches.

\`generate_system_digest_task\` doesn't have this issue presumably because it iterates differently or because its data-dir resolution is robust to the per-engine overlay (need to verify).

## Proposed fix

Two-part:

[...]

PR #2610, opened 2026-05-22
feat(var): add skip-reason affordance to var_state.json (#2608)

### LINKED ISSUE #2601: enhancement(var): skip off-RTH VaR recompute per-commodity when futures market is closed (588 noise lines/24h)
## Summary

`recompute_var_state` and post-close VaR cycles fire **regardless of futures-market state**. When the relevant futures market is closed (off-RTH), no underlying-future live price is available, so `var_calculator` correctly falls back to the documented "exclude position from VaR, log WARNING" path (CLAUDE.md: *"The VaR calculation ... properly downgrades severity to WARNING for expected off-hours `NO_DATA` states when underlying future prices are unavailable."*).

This is **not a bug** — it's the documented graceful-degradation behavior. But it produces **588 WARNING lines per 24h** across all 4 commodities, none of which represent any new information once the operator knows "yeah, the market is closed."

## Distribution (2026-05-21 17:07 → 2026-05-22 17:07 UTC)

All 588 events cluster strictly in off-RTH hours; **zero** during normal RTH or during today's two Gateway-wedge windows:

```
Hour (UTC) | Total | SB  | NG | KC | ZW
   00:00   |  144  | 64  | -  | -  | -
   01:00   |  144  | 96  | -  | -  | -
   02:00   |   32  | 32  | -  | -  | -
   03:00   |  208  | 112 | -  | -  | -    ← peak
   07:00   |    4
   21:00   |   40
   22:00   |   16
   --- RTH hours (13:30–20:00 UTC for KC/SB): zero events ---
```

**Per-ticker totals:** SB 304 (worst), NG 120, KC 84, ZW 80.

## Sample (verbatim)

```
2026-05-22 02:45:10 trading_bot.var_calculator - WARNING - VaR: No valid price for SB underlying future
2026-05-22 02:45:11 trading_bot.var_calculator - WARNING - No underlying future price for SOV6 C0.16 (symbol=SB) — excluded from VaR
2026-05-22 02:45:11 trading_bot.var_calculator - WARNING - No underlying future price for SOV6 C0.175 (symbol=SB) — excluded from VaR
2026-05-22 02:45:11 trading_bot.var_calculator - WARNING - No underlying future price for SOH7 P0.14 (symbol=SB) — excluded from VaR
2026-05-22 02:45:11 trading_bot.var_calculator - WARNING - No underlying future price for SOH7 P0.16 (symbol=SB) — excluded from VaR
2026-05-22 02:45:13 Orchestrator - INFO - Post-close VaR updated: 10 positions.
```

The final "Post-close VaR updated: 10 positions" log is misleading — the VaR snapshot has positions silently excluded; the 10-count doesn't mean those positions are at zero risk, just that they couldn't be priced.

## Proposed enhancement

**Option A — Per-commodity gate (preferred):** in the scheduling layer (or at `recompute_var_state` entry), check the commodity's futures-market state via `calendars.is_market_open(ticker, now)`. If closed for that ticker, skip the per-position VaR recompute for that ticker's positions. The portfolio-level VaR still recomputes for tickers whose markets ARE open at that time (this matters because the post-close cycles for KC, NG, SB, ZW fire at different UTC hours).

**Option B — Global gate:** if ALL active commodities' markets are closed, skip the post-close VaR cycle entirely until the next market-open boundary. Coarser but simpler.

**Either way:**
- Last successful VaR state should be preserved on disk (CLAUDE.md notes this is already done via atomic file locking + "preserves previous known-good results on disk during transient fetch failures")
- The dashboard / `var_state.json` should explicitly flag "stale due to market-closed" vs "stale due to fetch failure" so operators don't misread one as the other

## Why this matters

Beyond the 588 lines/24h log-noise reduction:

1. **Dashboard accuracy** — current `var_state.json` may show a "10 positions" portfolio VaR that has 4 positions silently excluded; an operator reading the dashboard mid-overnight could read it as the full portfolio risk
2. **Error-reporter triage** — these warnings count against `error_reporter_state.json` fingerprint dedup budget; suppressing them via the gate (vs filtering) lets the budget go to real signals
3. **Operationally cleaner** — "VaR isn't running because the market is closed" is a simpler mental model than "VaR is running but most positions are excluded"

## What this is NOT

- NOT a bug — the current behavior is correctly documented and fails gracefully
- NOT urgent — VaR state preservation works; no risk of crashes or data loss
- NOT a per-commodity adjustment that needs operator intervention — purely a scheduling improvement

## Related

- CLAUDE.md "Portfolio Risk Guard" section (documents the current off-hours WARNING fallback)
- PR #1648 (`var_iv_unavailable_uses_fallback` — sibling pattern at the IV layer)
- `trading_bot/calendars.py` (existing market-hours helper that would be reused)
- `trading_bot/state_manager.py` (atomic var_state.json persistence — already in place)

---
*Filed by `/review-logs` skill on 2026-05-22.*

rozavala:
Closes #2608. Follow-up to PR #2602 (closes #2601).

## What this PR adds

A schema field + dashboard surface that distinguishes the three VaR-state shapes operators care about:

| Badge | Condition |
|---|---|
| 🟢 OK | recomputed within expected cadence |
| 🟦 STALE (market closed) | gate fired, last skip is more recent than last successful recompute |
| 🔴 {status} | compute error (existing behavior, unchanged) |

### Schema (`trading_bot/var_calculator.py::VaRResult`)

Two new fields, both default empty:

```python
last_skip_reason: str = ""        # e.g. "market_closed"
last_skip_at: str = ""            # ISO timestamp of the skip event
```

`from_dict` defaults both to `""` for backward compat with pre-#2608 state files. The dataclass defaults `last_skip_reason=""` is also what clears the skip-reason on a successful recompute — `compute_portfolio_var` builds a fresh `VaRResult()`, which `_save_state` writes with `last_skip_reason=""`. No special clearing path needed.

### Write path (`orchestrator.py::_stamp_var_state_skip_reason`)

New helper called from the off-RTH gate. Reads existing `var_state.json`, sets the two skip-fields, writes back via the same locked-write helper that `var_calculator._save_state` uses. Preserves every other field — this is a metadata-only update.

```python
if not is_market_open(config):
    ...
    logger.info(f"--- VAR STATE RECOMPUTE ({trigger_source}) skipped: ...")
    _stamp_var_state_skip_reason(config, reason="market_closed")  # ← NEW
    return
```

IO errors are swallowed + logged. The gate's primary job (skip the recompute) must not fail because of a cosmetic write failure. On failure, the dashboard falls back to the legacy badge until the next recompute.

### Dashboard (`pages/9_Portfolio.py`)

Three-state status badge replaces the previous binary OK/error one:

```python
last_skip_reason = var_state.get("last_skip_reason", "")
last_skip_at = var_state.get("last_skip_at", "")
skip_is_recent = ...  # True iff last_skip_at > computed_epoch

if status != "OK":
    status_icon, status_label = "🔴", status
elif skip_is_recent and last_skip_reason == "market_closed":
    status_icon, status_label = "🟦", "STALE (market closed)"
else:
    status_icon, status_label = "🟢", "OK"
```

The `skip_is_recent` check is load-bearing — a historical skip already overwritten by a later successful recompute does NOT show the stale-clean badge. Avoids "we skipped at 22:00 last night, but RTH has been running for 3 hours and the state is fresh" false-positive blue badges.

- **The gate itself** (PR #2602) — unchanged. This is metadata only.
- **`var_calculator.compute_portfolio_var`** — unchanged. Naturally clears skip-reason via fresh `VaRResult()` default values on every save.
- **Pre-#2608 readers** — work unchanged. Missing fields default to `""` everywhere.
- **The Cockpit page** — currently doesn't surface VaR freshness explicitly; left untouched. Worth a follow-up if you want symmetric treatment.

## Tests — 9 new cases

`tests/test_var_state_skip_reason_2608.py`:

**Schema (4):**
- `test_var_result_from_dict_defaults_skip_fields_to_empty` — pre-#2608 state files still load
- `test_var_result_to_dict_includes_skip_fields`
- `test_var_result_round_trip_preserves_skip_fields`
- `test_var_result_default_skip_reason_is_empty` — clearing on successful recompute

**Stamp helper (3):**
- `test_stamp_skip_reason_preserves_existing_fields` — last-known-good VaR numbers stay intact
- `test_stamp_skip_reason_creates_stub_when_no_file` — cold-start path
- `test_stamp_skip_reason_swallows_io_errors` — fault-tolerance pin

**Integration (2):**
- `test_recompute_var_state_stamps_skip_reason_on_off_rth_skip`
- `test_recompute_var_state_does_not_stamp_when_market_open` — guard against over-stamping

Full suite: 3603 passed, 1 skipped, 0 failed.

## Rollback

Two-file revert on `var_calculator.py` + `orchestrator.py` (the test file can stay — regression pins remain valid against the unfixed code). Dashboard change at `pages/9_Portfolio.py` is a third file; reverts independently.

## Related

- **PR #2602** (closes #2601) — adds the gate this PR adds the affordance for
- **#2368** — VaR outer budget + freshness-aware Pushover suppression (orthogonal — operates on the failure path, not the skip path)


rozavala:
**✅ Ready to merge — clean observability addition with correct fail-safe semantics.**

**Verification**

- `VaRResult.from_dict` defaults both new fields to `""` — correct backward compat with pre-#2608 state files.
- `to_dict` uses `asdict(self)` which automatically includes all dataclass fields — round-trips correctly.
- `save_state_to_file_locked` / `load_state_from_file_locked` confirmed to accept `lock_path` as a keyword arg (verified in `trading_bot/state_manager.py:185–263`).
- Dashboard `computed_epoch` is a Unix float from `var_state.get("computed_epoch", 0)`. The `skip_is_recent = _skip_ts.timestamp() > computed_epoch` comparison is correct; it correctly returns `False` when `computed_epoch = 0` (cold start), so blue badge doesn't false-alarm before any VaR compute has ever run.
- The `_stamp_var_state_skip_reason` call at line ~807 is inside `recompute_var_state`'s body — Python resolves it at runtime (the function is defined later in the same module), not at import time. Fine.
- Fail-closed semantics preserved: the `try/except Exception` swallow in `_stamp_var_state_skip_reason` means an IO error on the cosmetic write never blocks the gate's primary job (skip the recompute). ✓
- Tests couldn't run locally (env missing `holidays`, `pytz`, and the full `requirements.txt` stack that `orchestrator.py` imports at module level), but the logic was verified manually. PR body reports 3603 passed, 1 skipped on the actual branch.

**Minor note (non-blocking)**: The stdlib imports inside the `try` block use underscore aliases (`import os as _os`, `import json as _json`) — an unusual style for a function that's not in a hot path. Worth cleaning up in a follow-up polish commit, but doesn't affect correctness.

**Follow-up filed**: #2613 — extend the three-state VaR badge to `pages/1_Cockpit.py`, which currently shows red staleness during the expected off-RTH market-closed case.

**MECE**
1. **Value**: Operators can distinguish expected overnight staleness ("market closed") from compute failures requiring attention. Reduces false-alarm pressure from the Cockpit's existing age-based red indicator (addressed fully by #2613).
2. **Risk**: Very low. Schema change is backward-compatible (empty defaults). Write failures are swallowed. No changes to VaR compute logic, gate logic, or compliance paths.
3. **Prior art**: Explicit follow-up to merged PR #2602 (#2601). No prior rejected attempts.
4. **Trajectory**: Fits the VaR reliability/observability workstream cleanly.

Labelled **Ready to be merged**.

---

rozavala:
**REQUEST CHANGES** — Path-resolution bug; writes will silently no-op in production.

## The bug

The new `_stamp_var_state_skip_reason` writes to:
```python
data_dir = config.get('data_dir', 'data')   # → 'data/KC/' in prod
state_path = _os.path.join(data_dir, 'var_state.json')
# → data/KC/var_state.json   ← orphan; nothing reads from here
```

But `var_state.json` lives at the **base** data dir, not per-engine. The canonical write path is `var_calculator._save_state` which uses module-level `_var_data_dir` set explicitly by `orchestrator.py:1604`:
```python
set_var_data_dir(os.path.dirname(data_dir))  # data/{ticker}/ → data/
```

Production verification:
```
$ ls -la /home/rodrigo/real_options/data/var_state.json
-rw-r--r-- 1 rodrigo rodrigo 1464 May 22 18:32 .../data/var_state.json
$ ls -la /home/rodrigo/real_options/data/KC/var_state.json
ls: cannot access ...: No such file or directory
```

The dashboard read path (`pages/9_Portfolio.py:192`) also reads from base, so the new skip-reason field will never surface — silent no-op.

## Why the tests don't catch it

Every test in `tests/test_var_state_skip_reason_2608.py` does `config = {"data_dir": str(tmp_path)}` with no per-ticker subdir. Writer and reader paths collapse to the same `tmp_path`, hiding the divergence that exists in production.

This is the per-ticker `data_dir` double-nest pattern — same bug class as #2096, #2110, missed in #2287.

## Fix (one-line)

In `_stamp_var_state_skip_reason`, mirror the canonical pattern:
```python
# Replace:
data_dir = config.get('data_dir', 'data')
# With either:
data_dir = _os.path.dirname(config.get('data_dir', 'data'))    # mirrors orchestrator.py:1604
# OR (preferred — uses the existing helper):
from trading_bot.data_dir_context import resolve_base_data_dir
data_dir = resolve_base_data_dir(config)
```

## Test fix

Add at least one case where `data_dir = tmp_path / "KC"` so writer/reader divergence is exercised. The current shape unifies them and bypasses the real failure mode.

## Pre-existing sibling bug (out of scope but flagging)

`_var_state_is_fresh` at `orchestrator.py:902` has the **identical** wrong pattern — `_os.path.join(config.get('data_dir', 'data'), 'var_state.json')`. In production this always returns False (the per-engine file doesn't exist), so the Pushover-suppression at `orchestrator.py:849` has been a silent no-op since #2368. Filing as a separate issue — your call whether to bundle into this PR.

CI green, schema is additive (backward-compat verified). Once the write path is corrected this PR cleanly closes #2608.

PR #2617, opened 2026-05-23
fix(var): correct per-ticker data_dir double-nest in _var_state_is_fresh (#2614)

### LINKED ISSUE #2614: bug(var): _var_state_is_fresh silent no-op in prod since #2368 — wrong data_dir resolution (#2610 review follow-up)
## Motivation

While reviewing PR #2610 (var_state.json skip-reason affordance), I found that the existing `_var_state_is_fresh` helper at `orchestrator.py:902` (introduced by #2368) has been a silent no-op in production since it landed.

## The bug

```python
def _var_state_is_fresh(config: dict, max_age_seconds: int = 300) -> bool:
    ...
    var_path = _os.path.join(config.get('data_dir', 'data'), 'var_state.json')
    if not _os.path.exists(var_path):
        return False                                # ← always returns here in prod
```

`config['data_dir']` is `data/<ticker>/` per-engine in production. `var_state.json` lives at the **base** data dir, written via `var_calculator._save_state` which uses the module-level `_var_data_dir` set by `orchestrator.py:1604`:

```python
set_var_data_dir(os.path.dirname(data_dir))   # data/{ticker}/ → data/
```

So `_var_state_is_fresh` looks at `data/KC/var_state.json` (or `data/NG/`, `data/SB/`, `data/ZW/`) which never exists. The freshness check always falls into the `not exists` branch and returns False.

Production verification:
```
$ ls -la /home/rodrigo/real_options/data/var_state.json
-rw-r--r-- 1 rodrigo rodrigo 1464 May 22 18:32 .../data/var_state.json
$ ls /home/rodrigo/real_options/data/KC/var_state.json
ls: cannot access ...: No such file or directory
```

## Behavioral impact

The intended Pushover-suppression at `orchestrator.py:849` never fires:

```python
if isinstance(e, (asyncio.TimeoutError, ConnectionError, ConnectionResetError)) \
        and not _in_ib_startup_grace():
    if _var_state_is_fresh(config, max_age_seconds=300):
        logger.info("VaR recompute ... var_state.json is fresh — suppressing Pushover")
    else:
        # always falls into this branch — Pushover fires every time
        ...
```

So every VaR recompute timeout / connection error has been generating a Pushover, including the cases where a sibling engine just produced a fresh state file. This may explain part of the Pushover-noise we've seen around the wedge-event clusters.

## Same pattern as MEMORY.md `feedback_per_ticker_data_dir_double_nest`

Same bug class as #2096, #2110, missed in #2287 review, and the bug currently being flagged on #2610. The per-ticker `data_dir` double-nest is a recurring failure mode.

## Suggested fix

Mirror the canonical pattern at `orchestrator.py:1604`:

```python
# In _var_state_is_fresh, replace:
var_path = _os.path.join(config.get('data_dir', 'data'), 'var_state.json')
# With either:
from trading_bot.data_dir_context import resolve_base_data_dir
var_path = _os.path.join(resolve_base_data_dir(config), 'var_state.json')
# OR (terser):
var_path = _os.path.join(_os.path.dirname(config.get('data_dir', 'data')), 'var_state.json')
```

Add a test where `config['data_dir'] = tmp_path / "KC"` and a `var_state.json` is placed at `tmp_path / "var_state.json"` — verifies the read path actually finds it.

## Why this isn't in PR #2610

PR #2610 introduces a *new* helper with the same wrong pattern; flagging the new instance is straightforward (now in the claude-fix comment thread). The pre-existing `_var_state_is_fresh` site is the same fix shape but operator-callers and impact differ — splitting keeps each PR's blast radius focused. Author of #2610 may choose to bundle both fixes if desired.

## Related

- PR #2610 (introduces the same pattern in `_stamp_var_state_skip_reason`)
- PR #2368 (introduced `_var_state_is_fresh`)
- PR #2096, #2110, #2287 (prior bugs of this class)
- MEMORY.md `feedback_per_ticker_data_dir_double_nest.md`

---
*Filed by `/review-prs` skill on 2026-05-23.*

rozavala:
Closes #2614.

Sibling fix to PR #2610 (#2608) — same `config['data_dir']` pattern slip, in the helper `_var_state_is_fresh` that escaped that PR's scope.

## The bug

```python
# orchestrator.py:_var_state_is_fresh (line 902, pre-fix)
var_path = _os.path.join(config.get('data_dir', 'data'), 'var_state.json')
if not _os.path.exists(var_path):
    return False
```

`config['data_dir']` is `data/<ticker>/` in per-engine task contexts (per `orchestrator.py:1604`'s canonical `set_var_data_dir(os.path.dirname(data_dir))`). The function therefore looked for `data/KC/var_state.json` — an orphan path that **never existed in production**. `os.path.exists` returned False, function returned False, and the entire #2368 Pushover-suppression heuristic became a **silent no-op**.

```
$ ls /home/rodrigo/real_options/data/var_state.json     # exists, recent
$ ls /home/rodrigo/real_options/data/KC/var_state.json  # never existed
ls: cannot access ...: No such file or directory
```

## Operational impact

Since PR #2368 landed on 2026-04-19, operators received misleading Pushovers (`🟠 VaR recompute TIMEOUT` / `🟠 VaR recompute SKIPPED — IB connection`) in cases where a sibling engine had successfully written fresh state within the 5-minute suppression window — the exact false-positive class the #2368 suppression was designed to prevent. **~4 weeks of degraded operator surface.**

## Fix

Use `resolve_base_data_dir(config)` from `trading_bot.data_dir_context` to resolve the BASE account-wide data dir from a per-engine config. Mirrors PR #2610's fix in the sibling helper.

```python
# After:
from trading_bot.data_dir_context import resolve_base_data_dir
var_path = _os.path.join(resolve_base_data_dir(config), 'var_state.json')
```

## Why the existing tests missed it

All five pre-#2614 `_var_state_is_fresh` tests use `config = {'data_dir': str(tmp_path)}` — the legacy single-engine layout. `resolve_base_data_dir` falls back to `config['data_dir']` when `base_data_dir` is missing, so under that test shape **writer and reader paths collapse to the same tmp_path**. The per-engine divergence that exists in production was never exercised.

This is the per-ticker `data_dir` double-nest bug class (#2096 / #2110 / PR #2287 review). Tests that don't model the engine_dir vs base_data_dir split silently mask the entire family. Documented as `feedback_per_ticker_data_dir_double_nest.md` in my memory; bit me yesterday in #2608/#2610 too.

## Tests

**Existing five preserved** — they pin the legacy fallback path (still load-bearing for older deploys).

**Two new regression cases:**

- `test_var_state_is_fresh_finds_base_path_under_per_engine_config` — primary repro. Builds `config` with `data_dir = tmp/KC/` and `base_data_dir = tmp/`, writes `var_state.json` at the base location only, asserts the function finds it. Would have caught the original #2368 bug.
- `test_var_state_is_fresh_does_not_read_per_engine_orphan_path` — defense-in-depth: even if someone wrote a stale file at the orphan `data/<ticker>/var_state.json` path, the function must NOT pick it up.

Both pass after the fix; both would have failed against the un-fixed code.

```
================= 11 passed (9 existing + 2 new) =================
```

- **Fail-closed semantics** — unchanged. Read/parse errors still return False.
- **Legacy single-engine fallback** (via `resolve_base_data_dir`'s tier-3 path) — preserved.
- **PR #2368's Pushover-suppression branch** in `recompute_var_state` — unchanged; just now actually works.
- **The #2608 sibling helper** `_stamp_var_state_skip_reason` (fixed in PR #2610) — unchanged.

## Rollback

Single-file revert on `orchestrator.py` + the test additions can stay (they continue to pass against the legacy fallback path).

## Related

- **PR #2610** (#2608) — adjacent fix in `_stamp_var_state_skip_reason`. Identical pattern slip.
- **PR #2368** — introduced `_var_state_is_fresh` with the buggy path. Pushover-suppression has been inoperative since.
- **#2096 / #2110 / PR #2287 (review)** — the per-ticker double-nest pattern this fits into.
- `feedback_per_ticker_data_dir_double_nest.md` (agent memory) — flagging this pattern for future surfaces.


rozavala:
Approved.

Mirrors PR #2610's fix shape (commit `2d630857`) one-for-one — uses `resolve_base_data_dir(config)` from `trading_bot.data_dir_context` to resolve to the base data dir where `var_state.json` actually lives.

**Tests catch the regression**:
- `_make_per_engine_config(tmp_path, ticker="KC")` sets `data_dir = tmp/KC/`, `base_data_dir = tmp/` — the exact production layout
- `test_var_state_is_fresh_finds_base_path_under_per_engine_config` — writes `var_state.json` at the BASE, asserts the function finds it. Against the OLD code (`config.get('data_dir')` → `tmp/KC/`), this assertion fails. Regression-catching confirmed.
- `test_var_state_is_fresh_does_not_read_per_engine_orphan_path` — defense-in-depth

**Full suite**: 3608 passed, 1 skipped, 0 failed (707s).

**Impact**: Restores the #2368 Pushover-suppression at `orchestrator.py:849` that has been a silent no-op for ~4 weeks. Every VaR recompute timeout/connection error has been triggering Pushovers — including the cases where a sibling engine had just written a fresh state. This may explain part of the Pushover noise around wedge-event clusters.

Ready to merge.

PR #2655, opened 2026-05-26
fix(emergency-close): thread config['data_dir'] into ledger lookup so PID reuse actually works (#2654 root-cause A)

rozavala:
Addresses **root cause A** of #2654 — the 16 `EMERGENCY_<sym>_<date>_<ts>` synthetic PIDs in the live ledger.

The other 3 root causes (reconciler PIDs, catastrophe writers, partial-close mismatches) are out of scope for this PR per CLAUDE.md safety rules + operator preference for "one root cause per PR."

## The bug

`emergency_close._lookup_original_position_info` calls `get_trade_ledger_df()` with **no `data_dir`** argument. The reader's default falls back to the legacy single-engine path under the project root (`<repo>/trade_ledger.csv` + `<repo>/archive_ledger/`).

In multi-engine production the ledger lives at `data/<ticker>/archive_ledger/` — the legacy paths don't exist. So the lookup always returned `{}` and every emergency close fell into the synthetic-PID escape hatch:

```python
# _resolve_emergency_position_id:106 (the escape hatch)
synthetic = f"EMERGENCY_{sym}_{today_date}_{int(_time.time())}"
logger.warning(
    f"Emergency close: no matching open entry in ledger for {sym}; "
    f"using synthetic position_id {synthetic!r} (ledger continuity lost)."
)
```

**The escape hatch was DESIGNED to fire only when no matching entry exists.** Pre-#2654 it fired on every emergency close because the lookup couldn't find the ledger at all. Classic per-ticker `data_dir` double-nest pattern (same shape as #2096/#2110/#2614, documented in agent memory `feedback_per_ticker_data_dir_double_nest`).

## Fix

One-line plumbing: thread `config['data_dir']` from `emergency_hard_close` through `_lookup_original_position_info` into `get_trade_ledger_df(data_dir=...)`.

```python
# emergency_close.py
def _lookup_original_position_info(
    open_positions: list[Position],
    data_dir: str | None = None,  # NEW
) -> dict[str, tuple[str, str | None]]:
    ...
    trade_ledger = get_trade_ledger_df(data_dir=data_dir)  # was: get_trade_ledger_df()
    ...

# emergency_hard_close call site
original_position_info = _lookup_original_position_info(
    open_positions,
    data_dir=config.get('data_dir'),  # NEW
)
```

The escape-hatch behavior is preserved verbatim — it still fires when no matching entry exists. The fix just ensures the lookup actually finds the ledger first.

- `_resolve_emergency_position_id` escape-hatch logic — unchanged. Synthetic PIDs still fire on true no-match.
- The other 3 root-cause categories from #2654 (reconciler PIDs, catastrophe writers, partial-close mismatches). Filed separately per the operator's "one root cause per PR" preference.
- Ledger read semantics — same `get_trade_ledger_df` reader, just called with the right argument.
- **Trade-execution path** — no changes to order placement, fill handling, or any IB-facing logic. Only the post-fill ledger-write PID lookup.

## Tests — 3 new + 6 existing mock updates

**New** (`tests/test_emergency_close_ledger_continuity.py::TestDataDirPlumbing2654`):

- `test_data_dir_forwarded_to_get_trade_ledger_df` — spies on the ledger reader, asserts the passed-in `data_dir` reaches it.
- `test_no_data_dir_arg_still_calls_get_trade_ledger_df` — backward-compat: callers without `data_dir` get the original behavior.
- `test_emergency_hard_close_call_site_passes_config_data_dir` — source-text pin on the call site. Prevents future regression where the kwarg is dropped from the call expression.

**Existing test updates** (`tests/test_emergency_hard_close_characterization.py`):

The 6 mocks that used `lambda: ledger` (no-arg) needed `**kw` to accept the new `data_dir` kwarg the helper now passes through. Mechanical update — same lambda body, same intent.

**All 14 existing `TestLookupOriginalPositionInfo` / `TestResolveEmergencyPositionId` tests pass unchanged.**

Full suite: 3706 passed, 1 skipped. The 2 failures (`test_var_calculator::test_iv_unavailable_uses_fallback`, `test_option_underlying_price_from_future`) are pre-existing flakes called out by previous PR bodies — unrelated.

## Expected operational effect

After deploy:
- Emergency hard closes on existing open positions will reuse the original entry's `position_id` instead of generating `EMERGENCY_<sym>_<date>_<ts>`.
- The 16 orphan `EMERGENCY_*` PIDs in the live ledger represent PAST emergency closes — the fix doesn't retroactively heal them. They'll persist until manual cleanup (separate operator action).
- The aggregator's open-position count will be 16 lower going forward as new emergency closes correctly link to their original PIDs.

## Risk

Per CLAUDE.md "Never modify execution logic without thorough testing": this PR doesn't touch any IB-facing execution code. The trade is already FILLED when `_lookup_original_position_info` is called. The function only affects which `position_id` string is written to a ledger CSV row. **Worst-case regression**: the function returns `{}` (as it did pre-fix) and synthetic PIDs are generated — i.e., we land back at the pre-fix state, not a worse state.

## Rollback

Three-file revert. The function-signature change is backward-compatible (`data_dir` defaults to `None` = pre-fix behavior), so reverting the call-site change alone is sufficient if you want to keep the test mock updates.

## Related

- **#2654** — umbrella issue covering all 4 root causes
- **#2649** — parent (operator-correction comment thread)
- **PR #2653** — sibling: per-leg net_qty fix for spread classification (same module family)
- **#1760** — original `_lookup_original_position_info` introduction (this PR makes the function actually work in multi-engine prod)
- Memory: `feedback_per_ticker_data_dir_double_nest` (#2096 / #2110 / #2287 / #2614 — same bug class)


rozavala (commented):
## ✅ Review complete — Ready to be merged

### What was reviewed

**Scope:** `trading_bot/emergency_close.py` — one-line plumbing fix threading `config['data_dir']` through `_lookup_original_position_info` into `get_trade_ledger_df`. Two test files updated (3 new tests + 6 mock-signature fixes).

---

### MECE assessment

**1. Does this add value?**
Yes — fixes a real, operator-observed production bug. In multi-engine production the legacy project-root path doesn't exist, so `_lookup_original_position_info` always returned `{}`, and every emergency close wrote a synthetic `EMERGENCY_{sym}_{date}_{ts}` position_id. This created 16 orphan PIDs in the live ledger: original open entries look "never closed", synthetic close entries look "never opened" — a double-count that inflates the open-position aggregator. The fix is proportionate (one argument, backward-compatible default, worst-case regresses to pre-fix behavior).

**2. What is the risk?**
Very low. `data_dir=None` default preserves the exact pre-fix calling contract for all other callers. The function runs _after_ the FILL is placed (post-fill ledger write only), so there is no path from this change back to order placement or IB-facing execution. The only safety-critical question — "could the correct path cause a crash?" — is covered: the reader already handles missing files cleanly and returns an empty DataFrame, which the existing escape hatch then handles.

**3. Prior art / rejected alternatives?**
This is the fourth instance of `feedback_per_ticker_data_dir_double_nest` (#2096/#2110/#2287/#2614). Consistent with all prior fixes. No variant of this was previously rejected.

**4. Trajectory-fit?**
Directly closes Root Cause A of umbrella issue #2654. Part of the tracked PID-discontinuity workstream. The other three root causes (B/C/D) are consciously deferred per operator preference for one-root-cause-per-PR.

---

### Code correctness

- `get_trade_ledger_df(data_dir=...)` signature confirmed in `order_manager.py:4501` — the kwarg already existed.
- `config.get('data_dir')` at the call site is safe (`None` fall-through if key absent = pre-fix behavior).
- The escape-hatch path in `_resolve_emergency_position_id` is **unchanged** — synthetic PIDs still fire on true no-match.
- Six mock updates (`lambda: ledger` → `lambda **kw: ledger`) are mechanically correct; no behavioral change to the mocked functions.
- Source-text pin test (`test_emergency_hard_close_call_site_passes_config_data_dir`) is a sound regression guard.

---

### Non-blocking follow-up filed

Two more `get_trade_ledger_df()` calls in the same function (`emergency_hard_close`) are also missing `data_dir` — tracked in **#2657** (not blocking this PR per "one root cause per PR" discipline):

- **Stale-detection gate** (~line 231 on this branch): in multi-engine prod, reads empty ledger → non-Friday emergency cycles silently skip all stale enforcement; Friday cycles bypass WITH_TREND exemption (re-introduces the #1946 regression at this callsite). Higher severity — recommend prioritizing immediately after this merges.
- **TMS sweep** (~line 593): no trading effect, but stale theses don't get cleaned up after emergency closes.


---

Issue #2145, opened 2026-05-11
strangler-fig step 8: characterization tests for orchestrator's live-trading hot paths


rozavala:
## Context

PR #2140 completed strangler-fig step 7 by extracting six pure scheduling primitives into `trading_bot/scheduling.py`. Cumulative reduction: `orchestrator.py` 8,249 → 6,220 lines (-24.6%) and `agents.py` 2,646 → 2,134 (-19.4%) across 8 new focused modules.

The PR body itself flags that this is the **\"easy half done\"** point for `orchestrator.py`. The remaining large functions are all live-trading hot paths:

| Function | Lines | Risk |
|---|---|---|
| `run_emergency_cycle` | ~1,030 | Very high — live trading, sentinel-driven |
| `run_sentinels` | ~680 | High — live trading |
| `emergency_hard_close` | ~510 | Very high — capital protection |
| `run_position_audit_cycle` | ~430 | High — position state mutation |
| `main` | ~340 | High — startup wiring |
| `_reconcile_orphaned_theses` | ~300 | High — state reconciliation |
| `guarded_generate_orders` | ~200 | High — order placement |

Extracting any of these naively risks live-trading regressions. The #2111 → #2117 close-path pair demonstrated the safe pattern: **characterization tests first, byte-identical extraction second.**

## Scope

**Phase A: characterization tests** for one (or more) of the above functions. Pick one to start — `emergency_hard_close` (510 lines) and `guarded_generate_orders` (200 lines) are the strongest candidates because:

- Both have well-defined input contracts (config + IB connection)
- Both have observable side effects via the trade ledger and IB Gateway mocks
- Both are smaller than `run_emergency_cycle` or `run_sentinels`, so the test surface is tractable

Goal: pin current behavior with enough specificity that a byte-identical move would not pass review on tests alone — i.e. the tests catch every documented behavior (success path, hold-off, abstention, mismatched contract, dry-run, etc.).

**Phase B: byte-identical extraction** in a separate PR, mirroring #2111 → #2117 and the #2114 hygiene rule:
1. Pure-move commit (AST + raw-text identical, comment preservation, issue-ref preservation).
2. Re-export at top of `orchestrator.py` under original (underscore-prefixed) names.
3. No back-import of `orchestrator` in the new module.
4. Optional cleanup commit on the same PR.

## Discipline (per #2114 + #2117 precedent)

[...]

PR #2690, opened 2026-05-28
fix(scheduling): anchor EOD portfolio P&L to latest active-commodity close (#2689)

rozavala:
## Summary

The **🏁 Daily Portfolio P&L** Pushover (`generate_eod_portfolio_pnl_task`) fired at **13:58 ET** — KC close (13:30) + `offset_minutes: 28` — **before NG (14:30) and ZW (14:20) closed**, let alone reconciled (close + 8). The cross-commodity aggregator reads filesystem ledgers only, so the headline number summed still-open, un-reconciled NG/ZW ledgers. (Fixes #2689.)

Root cause: in session mode, `post_close_tasks` anchor to *the running engine's own close*, and `eod_portfolio_pnl` is pinned to the **primary** engine (KC) via `commodity_filter: "primary"`. KC is one of the earliest closers, so the portfolio-wide task ran ahead of the later ones. No single PR broke it — the `"primary"` dedup (#2087) and session-mode anchoring (#2145) compose into it.

## Fix (latest-close anchor — the structural fix, not an offset bump)

- **`trading_bot/scheduling.py`** — add `_latest_active_close(active_commodities, fallback_close)` and an `anchor: "last_close"` branch in the `post_close_tasks` loop. Tagged entries offset from `max(close)` across active commodities instead of the engine's own close. A ticker whose profile can't load is skipped with a warning (degrades to the pre-fix behaviour, never crashes the build). The default (no `anchor`) path is byte-for-byte unchanged.
- **`config.json`** — `eod_portfolio_pnl` opts in with `"anchor": "last_close"`. Still `commodity_filter: "primary"` (anchor is orthogonal to the once-per-day dedup). Now fires at NG close + 28 = **14:58 ET**, after every close (max 14:30) and reconcile (max 14:38).
- **`trading_bot/reporting_tasks.py`** — replace the stale `18:30 ET / NG-18:25` docstring (a false invariant from the legacy absolute schedule) with the real anchor.

Why not just bump `offset_minutes`: that hardcodes the KC↔NG close gap and silently re-breaks if the commodity set, the primary, or the latest closer changes. The anchor self-heals (it's computed from whatever's active).

## Verification

`active_commodities = [KC, NG, SB, ZW]` → primary = KC; closes KC 13:30 / NG 14:30 / SB 13:00 / ZW 14:20.
- Before: KC 13:30 + 28 = **13:58 ET** (before NG/ZW close ❌)
- After: NG 14:30 + 28 = **14:58 ET** (after all closes + reconciles ✅)

## Test plan

- [x] New `tests/test_session_schedule_last_close_anchor.py` — unit tests for `_latest_active_close` (latest wins, fallback floor, empty list, unresolvable ticker skipped), integration through `_build_session_schedule` (anchored → 14:58, unanchored → 13:58, no-op on NG engine), and a production-config guard asserting the real `eod_portfolio_pnl` entry keeps `anchor: "last_close"` and builds at 14:58 (> last close, > last reconcile).
- [x] Full suite: **3869 passed, 1 skipped** (`pytest tests/`).
- [x] `mypy` clean on both gated files (`scheduling.py`, `reporting_tasks.py`).


rozavala (commented):
**Daily PR review — LGTM. Labelled `Ready to be merged`.**

## What the bug was

`eod_portfolio_pnl` carries `commodity_filter: "primary"` so it fires once per day on the KC engine. In session mode `post_close_tasks` anchor to the *running engine's own close*, so the task fired at KC-close + 28 = **13:58 ET** — before NG (14:30) and ZW (14:20) had even closed, let alone reconciled (close+8). The daily headline P&L was aggregating stale, un-reconciled NG/ZW ledgers.

Root-cause lineage: the original task lived in `_build_default_schedule` at hardcoded 18:30 ET (correctly after all closes). PR #2419 migrated it to the session template with `offset_minutes: 28`, but the anchor silently shifted from the legacy absolute time to the KC engine's own close, composing with the primary-engine dedup (#2087) and session-mode anchoring (#2145) to produce the timing regression.

## Fix assessment

The `anchor: "last_close"` mechanism is the right structural fix:
- `_latest_active_close` correctly floors at the engine's own close (never moves *earlier*), gracefully skips unresolvable tickers, and self-heals if the active-commodity set or primary engine changes.
- `active_commodities` and `close_t` are both in scope at the step-5 post_close_tasks loop (defined unconditionally at lines 482–483); the default (no `anchor`) path is byte-for-byte unchanged.
- The one task that opts in (`eod_portfolio_pnl`) is observability-only — Pushover + filesystem ledger reads, no IB calls, no execution decisions.

## MECE

1. **Value**: Fixes a systematic correctness bug in the daily P&L summary (headline number was wrong on every day NG or ZW positions were active). Structural solution that won't re-break if the commodity set or primary engine changes.
2. **Risk**: Low. Production-touched surface is two files (`scheduling.py`: new 34-line helper + 8-line branch; `config.json`: 1 line). Only tasks explicitly tagged `anchor: "last_close"` are affected. Degradation is graceful (falls back to pre-fix behaviour, never crashes the schedule build).
3. **Prior art**: No rejected attempts at this pattern. The fix restores the semantic intent of the original 18:30 ET absolute schedule.
4. **Trajectory fit**: Fits the session-mode scheduling infrastructure workstream (#2145, #2419). No orphan infrastructure — consumed immediately by the existing task.

## Verification

- All CI passes: 3869 passed, 1 skipped; mypy clean; ruff F821 clean.
- No transcription drift risk: `post_close_tasks` are processed only in `_build_session_schedule`; `_build_default_schedule` uses a hardcoded list unaffected by this change.
- 10 new tests covering the helper (5 unit cases), integration through `_build_session_schedule` (3 cases), and a production config guard (2 cases) that will catch any future `anchor` key removal.

---

PR #2720, opened 2026-05-31
test: make date-dependent fixtures relative to stop calendar time-bombs

rozavala:
## Summary

5 tests started failing on **2026-05-31** (they passed 2026-05-30) — `main`'s CI went red, so every open PR's full suite shows them. All are **date-dependent fixture time-bombs**: they hardcode dates that aged past a runtime guard as the wall-clock advanced (the same class as #2670). **Test-only — no production code touched.**

| Test(s) | Why it tripped | Fix |
|---|---|---|
| `test_var_calculator::test_iv_unavailable_uses_smile_neighbor` | Options hardcoded expiry `20260601`; once ~1 day out, `var_calculator` excluded them ("Expired or invalid expiry") → 0 snapshots → `[s for s ...][0]` `IndexError` | relative `+30d` expiry (matches the #2670 fix already applied to sibling tests in this file) |
| `test_weekly_close_reconciliation::test_reconcile_friday_entry_uses_friday_close` | Contract `202603` (expiry 2026-03-31) crossed the reconciler's *"skip contracts expired >60 days ago"* guard on day 61 (today) → row skipped → `exit_price` stayed `None` | relative near-future contract month; the known-good Friday entry/bars are unchanged (entry age trips no gate) |
| `test_council_history_batching_2552` ×3 | `_make_minimal_df` hardcoded the same `202603`; the >60-day skip starved the reconciliation walk before the IB call, so the reconnect/abort paths under test never ran (`mock_reconnect.await_count 0 != 1`) | relative near-future contract month + recent-but-settled entry base |

## Root-cause discipline

Each failure was traced to the *specific* runtime guard it tripped (not assumed): the var_calculator expiry-exclusion, and the reconciler's 60-day expired-contract skip (`reconciliation.py`). The weekly_close and council_history failures share the **same** 60-day-skip bomb via a hardcoded `202603` — the entry dates themselves trip no gate, so they were left fixed where harmless and made relative only where it keeps the fixture coherent.

## Verification

- Each test **fails on clean `origin/main` (62e1b9148)** and **passes** after the change (verified per test).
- Full suite: **3951 passed, 1 skipped, 0 failed** (previously 5 failed).

## Follow-up (not in this PR)

[...]

Issue #2641, opened 2026-05-26
enhancement(pattern-gate): promote log-only detector to blocking after audit period (#2640 follow-up)


rozavala:
## Motivation

PR #2640 ships the SB-style catastrophic-loss pattern gate as **log-only** to allow false-positive measurement before enforcement. Issue #2637's acceptance criteria explicitly requires a tracking issue for the log-only → blocking flip. Filing this proactively so it doesn't fall through the cracks during the audit period.

## What this issue tracks

After the audit period (~1 week of production data), review `data/<TICKER>/pattern_gate_log.csv` (or whichever path #2640 settles on) and decide whether to promote the gate from log-only to blocking:

- **Calibration**: measure observed false-positive rate per commodity. Are KC/NG/ZW firing the gate at rates that suggest the SB-tuned thresholds (15% SMA-200 deviation, 26% range bands) are too tight for other commodities? Consider per-commodity calibration.
- **Promotion mechanism**: in `trading_bot/signal_generator.py` (and the still-to-be-added `emergency_cycle.py` integration, per the claude-fix on #2640), replace the log-only WARNING with a `continue` / `return` that drops the signal. Surface the rejection via `log_funnel_event` (mirrors PR #2620's IC-rejection-funnel pattern).
- **Override pathway**: decide whether a "force-through" mechanism is needed for any legitimate post-audit pattern matches.

## Suggested implementation

```python
# In signal_generator.py + emergency_cycle.py:
fires, reason = pattern_gate.check(council_log_entry, market_ctx, decision)
if fires:
    if ENFORCE_PATTERN_GATE:  # config-driven flag once tuning settled
        log_funnel_event(stage=FunnelStage.PATTERN_GATE_BLOCK, ...)
        logger.warning(f"Pattern gate BLOCKED: {reason}")
        continue
    else:
        pattern_gate.log_observation(...)  # current log-only behavior
```

## Audit-end criteria

- Minimum 1 week of production logging across all 4 active commodities
- At least 1 EMERGENCY-trigger signal observed per commodity (after #2640's claude-fix lands the emergency-path integration)
- False-positive rate measured and compared to the SB-incident base rate
- Per-commodity calibration decision recorded

## Why this isn't in PR #2640

#2640's scope is the detector only. The promotion decision is data-driven and requires the audit to complete. Separating keeps both PRs reviewable independently.

## Related

- PR #2640 (log-only detector implementation)
- Issue #2637 (SB-style catastrophic-loss combination)
[...]

Issue #2625, opened 2026-05-24
investigate: SB specialist agents called BEARISH during 9% sugar rally — why?


rozavala:
## Summary

#2622's v2 deep-dive (linked: https://github.com/rozavala/real_options/issues/2622#issuecomment-4527942511) localized SB's 3% WITH_TREND win rate to a **directionally-wrong-call problem at the analyst layer**, not a trend-signal infrastructure problem. Sugar rallied +9.13% during the Apr-May 2026 trading window while SB specialist agents called BEARISH:

- supply_chain: 41% BEARISH (vs 1-3% on other tickers)
- technical: 65% BEARISH
- macro: 24% BEARISH (vs 0-6% on other tickers)
- master: 74% BEARISH

This issue tracks the focused investigation: **why did SB specialists systematically call bearish during a 9% sugar rally?**

## What needs investigation

For each of the ~60 closed SB WITH_TREND trades during 2026-04-20 → 2026-05-24:

1. **Pull the agent evidence text** for the load-bearing bearish calls (supply_chain, technical). What was the analyst seeing that drove the BEARISH conclusion?
2. **Cross-check the evidence against the actual sugar price action**. If the agent cited (say) "Brazilian production glut" as bearish, was the sugar price already incorporating this, or was the agent fighting the tape?
3. **Identify the sentinel / data feed** driving the bearish narrative. If a specific sentinel (e.g., logistics, X-sentiment, news, prediction-market) was feeding bearish sugar signals while the price rallied, that's the load-bearing fix point.
4. **Check whether sentinel outputs match observed price action**. The system may have a sentinel-to-price misalignment specific to sugar (or the Brazilian-coverage data sources may have stale narratives).

## Hypotheses

- **H-A**: Sentinel feeds carried persistent bearish Brazilian-production / Brazilian-glut narratives that the market had already priced in (or was now reversing).
- **H-B**: Inventory data (USDA / industry stocks) was bearish on paper, but spot prices ran up on weather/mill closures the inventory data lags.
- **H-C**: The agents over-weighted supply factors vs demand-side / weather-side / vol-side factors that were the actual price drivers.
- **H-D**: A specific data source (Databento? EIA? USDA NASS?) had a bias in coverage that propagated.

## What this is NOT

- Not asking to rewrite SB personas. Per memory: persona text lives in the renderer (not config.json), and persona changes are fragile. Audit first; persona changes only if a load-bearing layer is identified.
[...]

PR #2721, opened 2026-05-31
feat(schema): persist raw_confidence (VaR pre-clamp) to council_history (#2714)

rozavala:
## Summary
- Register `raw_confidence` (the VaR pre-clamp confidence — `var_dampener`'s partner) in `COUNCIL_HISTORY_FIELDNAMES` + `COUNCIL_HISTORY_BACKFILL_DEFAULTS`, and surface it from `signal_generator`'s `council_log_entry`.
- Completes the VaR/trend-extreme forensics symmetry started in #2709 (#2647): a clamped signal's VaR *and* trend-extreme pre/post magnitudes are now both readable from one `council_history.csv` row.

## Root cause
#2709 registered `te_dampener`/`raw_conf_pre_te`. The `/review-prs` reviewer flagged that the VaR side was only half-persisted: `var_dampener` reaches the CSV but `raw_confidence` (the pre-VaR-clamp confidence) was never in `COUNCIL_HISTORY_FIELDNAMES`. `log_council_decision` builds rows only from those fieldnames and drops unlisted keys via `extrasaction='ignore'`, so even though `calculate_weighted_decision` returns `raw_confidence`, it never reached the forensic record.

## What changed
- **`trading_bot/schema.py`** — add `raw_confidence` to `COUNCIL_HISTORY_FIELDNAMES` (v12 block) and `COUNCIL_HISTORY_BACKFILL_DEFAULTS` (`""` = VaR didn't clamp / pre-feature), appended to preserve append-only migration ordering.
- **`trading_bot/signal_generator.py`** — add `"raw_confidence": weighted_result.get('raw_confidence')` to `council_log_entry`, next to `var_dampener`.
- **`tests/test_row_builder_symmetry.py`** — add `raw_confidence` to `DEFERRED_SCHED_ONLY` (SCHEDULED-only by construction, same as `var_dampener`/`te_dampener`: the EMERGENCY path never calls `calculate_weighted_decision`).
- **`tests/test_debate_schema.py`** — update the append-order guard to reflect the v12 tail.
- **`tests/test_raw_confidence_persistence_2714.py`** (new) — CSV-roundtrip test: calls `log_council_decision` and reads `council_history.csv` back, asserting `raw_confidence` persists when the VaR clamp fired (and defaults empty otherwise). Confirmed failing without the schema registration.

## Test plan
- [x] Affected tests green (schema, symmetry, forensics, new persistence test — 14 passed)
- [x] New persistence test confirmed failing on `main` (column dropped) and passing with the fix
[...]

PR #2726, opened 2026-05-31
feat(pattern-gate): link gate matches to opened position_id for outcome auditing (#2641)

rozavala:
## Summary

Partially addresses #2641 (promote the #2637 catastrophic-loss gate from log-only to blocking). **The gate stays log-only here** — this PR builds the prerequisite that the audit showed was missing.

## Why not just flip to blocking

I audited the ~5 days of #2640 log-only data. It's **inconclusive for the blocking decision**:

- The signature fired **3× (KC + 2 NG)**, all `WITH_TREND` + conf 0.80 + BEARISH at −15 to −17% below SMA-200 — recurring and cross-commodity, matching #2625's 60-loss SB signature.
- But `pattern_gate_log.csv` recorded only **1 of 3** (the 2 NG matches predate the gate going live on the DEV engine on merge day — it logs correctly once live).
- **Decisively: none of the 3 matched signals filled** → there is zero realized-outcome data → the would-be false-positive rate is **unmeasurable**.
- Root blocker: **no key links a gate-fire to its trade.** The gate logs `cycle_id` at signal time; `trade_ledger.csv` is keyed by `position_id`; `council_history` has `cycle_id` but no `position_id` (and its `pnl_realized` is unreliable). So you can't tell whether a fire became a winning or losing trade.

## What this adds (the missing link, log-only preserved)

- **`signal_generator`** — stamps `_pattern_gate_matched` / `_pattern_gate_reason` onto the returned signal dict so the flag rides `decision_data` to the fill handler. The `'… in dir()'` guard mirrors the existing `decision`/`thesis_strength` guards in that dict (no NameError on any return path).
- **`order_manager._handle_and_log_fill`** — once per opened position (deduped like `POSITION_OPENED`), when the originating decision carried the gate flag, appends `data/<ticker>/pattern_gate_outcomes.csv` keyed by the ledger **`position_id`**. Fully **fail-open**, gated on `_ledger_logged`, and a **no-op for the ~all non-matching trades** — never disturbs the fill path.
- **`pattern_gate.audit_outcomes(data_dir)`** — joins fires (`pattern_gate_log`) + fills (`pattern_gate_outcomes`) + `trade_ledger` P&L by `position_id` → `NOT_FILLED` / `OPEN` / `FILLED_WIN` (would-be false positive) / `FILLED_LOSS` (true positive) + `false_positive_rate` (`None` until something closes — the honest "not yet measurable" state). P&L from the ledger's pre-signed `total_value_usd` (the trustworthy source; `council_history.pnl_realized` is unreliable).

Once fires start filling and closing, `audit_outcomes` yields a real FP rate, and the promote-to-blocking decision (the rest of #2641) can be made on data.

## Testing

- **Additive only** — no existing logic changed (315 insertions, 0 deletions).
- 5 new tests (outcome-link write + fail-open; audit join classification incl. archived ledger; `None` FP rate when nothing closed).
- Targeted fill/signal/order integration batch (`test_order_manager` + `test_signal_flow` + emergency hot-path + pattern_gate): **105 passed**.
- Full suite: **3951 passed**. The 5 failures are the pre-existing date-bomb fixtures fixed by #2720 (verified identical on clean `origin/main`), unrelated to this change.


rozavala:
**Ready to merge** (after #2720 lands — see note).

### Value
Builds the missing link the #2641 audit needs: the #2637 pattern-gate is log-only, but a fire couldn't be tied to its trade outcome (gate logs `cycle_id`; ledger is keyed by `position_id`; `council_history.pnl_realized` is unreliable). This stamps the match flag onto `decision_data` at signal time, records the opened `position_id` at fill time, and adds `audit_outcomes` to join fires+fills+ledger P&L into a would-be false-positive rate — so the promote-to-blocking decision (rest of #2641) can be made on data instead of guesswork. The gate stays **log-only** here.

### Risk
Low. Additive (315 insertions, 0 deletions); the fill-path hook is `_ledger_logged`-gated, deduped once per position (like `POSITION_OPENED`), fully fail-open (unwritable path → WARNING, never disturbs the fill), and a no-op for ~all non-matching trades. Transcription-drift check: both `decision_data` return-builders in `signal_generator.py` (1381, 1413) carry the stamp — no third unstamped path. `position_uuid` is confirmed the same value written to the ledger as `position_id` (order_manager.py:2557), so the audit join key is sound.

### Verification (full suite, prod venv, head d4639d029)
`PYTEST_EXIT=1` on **only** the 5 pre-existing date-bombs fixed by #2720 — zero new failures. 5 new outcome-link/audit tests pass (incl. archived-ledger join + `None` FP-rate when nothing's closed).

### One gap (non-blocking, follow-up filed)
The outcome link is wired on the **SCHEDULED** path only. `emergency_cycle.py` logs the gate *fire* but never stamps `_pattern_gate_matched` onto its decision_data (it builds via `run_specialized_cycle`), so an emergency-origin match that fills classifies as `NOT_FILLED` and biases the FP-rate denominator. Minor for now (the #2625 signature is a SCHEDULED-path phenomenon), tracked in **#2728**.

### Prior art / fit
Documented follow-up to merged #2640 (#2637 gate); partially addresses #2641.

> **Merge note:** touches `signal_generator.py` in a different region than #2721; rebase after #2720.

PR #2796, opened 2026-06-05
feat(strategy): symmetric IV-driven vol-lane selector (v2, opt-in) (#2795)

rozavala:
The structural rewrite behind #2793 — addresses #2795. **Stacked on #2794** (base is its branch, so this diff shows only the v2 work; it retargets to `main` when #2794 merges).

## Why
The vol lane is a one-way valve to `LONG_STRADDLE`. In `route_strategy` (v1):
- `IRON_CONDOR` is gated on `regime=='RANGE_BOUND' and vol_sentiment=='BEARISH'` — but `vol_sentiment` lands NEUTRAL ~93% (NG: **BEARISH 0/123** cycles in 30d), so **IC is mathematically impossible** for NG and rare elsewhere.
- `LONG_STRADDLE` is gated on directional **conflict** — not a vol forecast.

So NEUTRAL cycles (60–74% of all cycles) resolve to **straddle-or-nothing**, and the system can never sell premium even in textbook range-bound/normal-vol conditions. That asymmetry produced the #2793 NG straddle book (6/6 losers → −8.7% HALT).

## What v2 does
Drives **both** legs off one continuous **implied**-vol signal (`iv_rank`), symmetric:

| condition | strategy |
|---|---|
| `iv_rank > condor_iv_rank_breach` (rich IV) + `RANGE_BOUND` | IRON_CONDOR |
| `iv_rank < straddle_iv_rank_max` (cheap IV) + real move trigger | LONG_STRADDLE |
| otherwise | NO_TRADE |

- **Conflict demoted** — it only fires a straddle when IV is already cheap.
- Needs **implied** vol (`iv_rank`); realized `vol_zscore` is ~0 in `RANGE_BOUND` so it can't express "rich options on a quiet tape".
- The #2793 vega/theta gates and the #2512 cooldowns **still compose** on top.
- **Reuses existing infra**: `condor_iv_rank_breach` (per-commodity NG=80 / SB=90 / global 70) — the *same* threshold the IC exit-validation uses (#2186), so selection and exit agree on "rich IV". `straddle_iv_rank_max` defaults 30 (the regime-engine's low-IV marker).

## Safety / rollout
- **Default OFF** (`config.strategy_tuning.vol_lane_v2_enabled=false`). With the flag off, behavior is **bit-for-bit v1** (the v2 branch precedes the v1 elif chain; v1 changed only `if`→`elif`).
- `build_market_context` fetches `iv_rank` **best-effort and only when the flag is on** — default IB load and behavior unchanged; any failure ⇒ `iv_rank=None` ⇒ **v1 fallback**.
- **Flipping the flag changes live strategy selection**, so it must be **backtest/shadow-validated** before enabling. (Mirrors the team's default-off pattern, e.g. #2738.)

## Plumbing
`route_strategy` gains `iv_rank` / `condor_iv_rank_breach` / `straddle_iv_rank_max` / `vol_lane_v2`; `signal_generator` + `emergency_cycle_helpers` resolve + pass them; `build_market_context` populates `iv_rank`.

## Tests (run under the prod venv)
- **router** 44 (12 new v2): symmetric IC on rich-IV+range, straddle on cheap-IV+catalyst/conflict, mid-IV→NO_TRADE, rich-IV-but-not-range→NO_TRADE, per-commodity breach (85 rich for NG@80 / not SB@90), vega+theta+cooldown composition, v1 fallback when `iv_rank` None, flag-off ignores `iv_rank`.
- **market_data** 11 (4 new): flag-off ⇒ no IB fetch + `iv_rank=None`; flag-on populates; fetch-failure best-effort; `'N/A'`→None.
- Full `pytest tests/` running — will confirm green before requesting merge.

## Deferred (tracked on #2795)
- Delete the now-dead inline NEUTRAL block in `signal_generator` (shadowed since v4.1; behavior-preserving but a careful separate refactor).
- Extend the consecutive-loss cooldown to the emergency/sentinel path (adds a `council_history` read to the hot path).
- Retire `vol_sentiment` as a selector once v2 is validated.

## Validation ask
Before flipping the flag: a backtest/shadow comparing v2 vs v1 selection on recent cycles (does IC actually fire in the right spots? does straddle volume drop to cheap-IV setups?). Happy to build the shadow harness next.


rozavala:
Added the **shadow-validation harness** as commit 2 (`vol_lane_v2_shadow`, default off): with v2 OFF, every NEUTRAL cycle still trades v1 but the router is re-run in v2 mode on the same inputs (live iv_rank) and the v1-vs-v2 diff is logged to `data/<ticker>/vol_lane_shadow.csv` — no trade changes. `scripts/analyze_vol_lane_shadow.py` summarizes it (divergence rate, v1→v2 transition matrix, IC-revival count). This is the gate to flipping `vol_lane_v2_enabled`.

Why forward shadow and not an offline backtest: historical `iv_rank` is logged only during exit-validation, so it's far too sparse at selection time — a nearest-time join just smears a few observations across hundreds of cycles. A quick offline pass confirmed the direction (SB: v2 adds ~19 ICs where v1 traded nothing; ZW: flips 5 rich-IV straddles to IC + adds 24) but NG/KC came out as pure join artifacts. Real per-cycle iv_rank only exists going forward.

Issue #2670, opened 2026-05-28
bug(test): test_var_calculator::test_iv_unavailable_uses_fallback RED on main — stale fixture trips the #1648 stale-price substitution guard


rozavala:
## Summary

`tests/test_var_calculator.py::test_iv_unavailable_uses_fallback` fails on current `main` (confirmed `fb63a20a5` — `var_calculator.py` + `test_var_calculator.py` are byte-identical between main and the branch this surfaced on). It's almost certainly a **stale test**, not a code regression: the test fixture feeds an unrealistic option market price that correctly trips the stale-averageCost substitution guard (#1648), but the assertion still expects the raw market price.

## The failure

```
assert option_snap.current_price == 2.5, "Should use option market price"
E  AssertionError: Should use option market price
E  assert np.float64(0.2833) == 2.5
```

```
WARNING trading_bot.var_calculator:var_calculator.py:768 IV unavailable for KC 400C (conId=123) — using profile fallback IV=0.35
WARNING trading_bot.var_calculator:var_calculator.py:853 Market price for KC 400C (2.5000) deviates 782% from B-S model (0.2833) [S=380.0000 K=400.0000 T=0.008 σ=0.350] — likely stale/averageCost. Substituting model price.
```

## Why this is a stale test, not a regression

The fixture: KC 400-strike call, ~3 days to expiry (T=0.008yr), underlying at 380 (20 points OTM), σ=0.35, "market price" = **2.5**.

A 20-point-OTM call with 3 days left at 35% vol is genuinely worth **~0.28**, not 2.5. The guard at `trading_bot/var_calculator.py:852-860` is doing exactly its job:

```python
elif model_price > 0 and not model_broken:
    # Option A: stale-averageCost substitution guard. Require both a
    # large percentage deviation AND a meaningful absolute difference
    # (≥5 ticks in per-unit terms). ... (#1648)
    if _tick_size is not None:
        deviation = abs(opt_price - model_price) / model_price
        abs_diff = abs(opt_price - model_price)
        if deviation > 0.80 and abs_diff > 5 * _tick_size and opt_price > 5 * _tick_size:
            logger.warning(f"Market price ... deviates {deviation:.0%} ... Substituting model price.")
            opt_price = model_price
```

[...]

PR #2621, opened 2026-05-23
feat(realized-move): daily runtime writer for observation sidecar (#2580)

rozavala:
Closes #2580. Follow-up to PR #2577 (closes #2535).

## What this PR adds

A per-engine post-close task that keeps `realized_move_observations.csv` growing past the one-time backfill, so the 252-day rolling window `realized_move_rank.move_to_rank` reads from doesn't go stale.

PR #2577 explicitly deferred this in its body:

> **Deferred (follow-up issue)**: The **daily runtime writer** for fresh `source='runtime'` rows. Without it, the sidecar ages — operator can periodically re-run the backfill as a stopgap.

## Design

New `log_realized_move(config)` in `trading_bot/realized_move_runtime.py`:

```
post_close + 6 min
  → get_active_ticker(config)
  → get_commodity_profile(ticker)
  → get IB connection ("var_audit" pool)
  → get_active_futures(symbol, exchange)   ← ICE→NYBOT fallback for off-RTH KC
  → reqHistoricalDataAsync(durationStr='2 D', barSize='1 day', TRADES, useRTH)
  → pick most recent valid bar
  → move_pct = |close - open| / open * 100
  → log_observation(data_dir, ticker, move_pct, source='runtime', timestamp=bar_ts)
```

- **Per-engine**: each commodity engine runs the task for ITS ticker — mirrors `log_equity_snapshot` / `recompute_var_state`.
- **Shape parity with backfill**: same OHLC fetch + same per-ticker per-UTC-date dedupe, so runtime rows are byte-compatible with backfill rows in the distribution.
- **ICE → NYBOT fallback**: matches the backfill script's behavior for off-RTH KC qualification rejections.
- **2-day window** (not 1): defensive against the "exactly 1 day at session close returns 0 bars" IB edge case. The latest valid bar is picked.
- **Fault-tolerant**: any IB outage, profile bug, or fetch failure logs WARNING and returns silently. The 252-day window absorbs single-day misses; next day's task fills the gap. Cannot block the rest of the post-close chain.

## Schedule placement

```diff
  post_close_tasks:
    - recompute_var_state         (+2 min)
    - log_equity_snapshot         (+5 min)
+   - log_realized_move           (+6 min)   ← NEW
    - reconcile_and_analyze       (+8 min)
    - brier_reconciliation        (+20 min)
    - sentinel_effectiveness      (+23 min)
    - eod_shutdown                (+26 min)
    - eod_portfolio_pnl           (+28 min)
```

[...]

PR #2865, opened 2026-06-11
fix(realized-move): file-scoped dedupe, per-ticker exchange resolution, degenerate-bar guard (#2854)

rozavala:
## Summary

Fixes all three data-quality defects in the realized-move observation sidecar (`realized_move_observations.csv`, written by `trading_bot/realized_move_runtime.py`, shipped PR #2621 for #2580) that #2854 surfaced during the 06-10/06-11 deploy-wave restarts:

1. **Restart-recovery appended duplicate rows** — dedupe is now file-scoped, not process-cache-scoped.
2. **KC qualified on ICE first → daily ERROR-level Error 200** — exchange now resolves via the #2741 per-ticker pattern; KC queries NYBOT first.
3. **Exact-0.000000 artifact rows** — degenerate `open == close` IB stub bars are skipped with a warning, never written.

## Root cause

**Symptom 1 — duplicates.** `realized_move_logger`'s per-ticker per-day dedupe lived only in the process-local `_last_write_date_by_ticker` cache. `log_realized_move` has `RECOVERY_POLICY {'policy': 'ALWAYS', 'idempotent': True}` — its claimed idempotency rested entirely on that cache, which resets on every restart. Each restart re-ran the task and re-appended the same session: confirmed in all four prod sidecars (KC/NG/SB/ZW each show duplicate dates on every restart day — 2026-05-29 ×5, 06-10 ×3, etc.). Same-date rows sometimes carry *different* values (KC 06-10: 1.304078 ×2 then 1.262679 from the 06-11 03:27 recovery) because the re-fetch picks up settlement revisions. Beyond over-weighting restart days, the inflation would have opened the `MIN_OBSERVATIONS=60` cold-start gate early on a corrupted distribution (the prod files are runtime-only, 25–26 rows each).

**Symptom 2 — Error 200.** `realized_move_runtime.py` took `exchange = profile.contract.exchange` — `ICE` for KC, the exact profile-vs-prod-proven divergence #2741 already established (prod-proven KC exchange is `NYBOT`). Every run burned an IB request and emitted an ERROR-level Error 200 before the line-94 `ICE→NYBOT` fallback rescued the fetch — confirmed daily in retained prod logs since 2026-06-02 (1–6×/day).

**Symptom 3 — exact-0.0 rows.** Proven proximate mechanism from prod logs: on 2026-06-09, IB returned a daily TRADES bar with `open=241.65 close=241.65 (bar date=2026-06-09)` on **both** write paths — the 17:36 UTC scheduled post-close fetch *and* the 22:30 UTC restart-recovery fetch (deterministic, so not a partial-bar race) — while KC actually moved ~0.9%. The bar passed the validity filter (positive, non-NaN) and was written verbatim as `0.000000`. **Open question (upstream, unproven):** *why* IB built that stub bar. It correlates with the 06-08/06-09 HMDS-farm trouble window (the 06-08 17:36 and 06-09 03:04 fetches both died with `TimeoutError`), so an HMDS settlement-placeholder/no-trade stub is the likely cause, but that's IB-side and not provable from our logs. The guard is therefore shape-based (`open == close` → skip + WARNING), which covers the mechanism regardless of IB's reason, on both write paths (they share the single computation).

## What changed

- **`trading_bot/realized_move_logger.py`** — new `_has_row_for_utc_date()`: before any `source='runtime'` append, the sidecar file is scanned for a row on the same UTC date (any source — a backfill row covering the day also suppresses). The file is the dedupe source of truth; the in-process cache is demoted to a fast path and is now set only **after** a successful append (a failed write no longer suppresses a same-day retry). Scan failures fail open (append; a missed dedupe is bounded noise, a false positive loses the day's observation).
- **`trading_bot/realized_move_runtime.py`** — new `_resolve_observation_exchange()` using the #2741 accessor chain: `config['commodity_overrides'][ticker]['exchange']` → top-level `config['exchange']` (per-engine injected) → profile with the ICE→NYBOT IBKR mapping. The raw profile exchange (ICE for KC) and the legacy ICE→NYBOT hop are kept as **true fallbacks**, tried only when the resolved exchange returns no contracts; non-ICE tickers (NG/ZW) no longer get any wasted cross-exchange attempt. Plus the degenerate-bar guard: `open == close` → WARNING + skip, never a 0.0 row (a single skipped day is absorbed by the 252-day window, per the module's own design).
- **`trading_bot/realized_move_rank.py`** — consumer-side per-date collapse in `_load_recent_moves()` (one observation per UTC date, last row wins = most settled fetch) to neutralize the legacy duplicate rows already on disk without a data-mutation script, and to keep the `MIN_OBSERVATIONS` gate counting unique dates. This **does not mask the writer fix**: all writer tests assert raw file contents directly, and the writer restart test proves a single physical row.
- **Tests** — `tests/test_realized_move_runtime_2580.py`: restart simulation (dedupe cache cleared between two task runs, settlement-revised re-fetch → exactly one row, first write wins), NYBOT-first/no-ICE-attempt for KC, ICE retained as true fallback, explicit-ICE-override-first, 6 resolver unit tests, degenerate-bar skip + warning. `tests/test_realized_move_rank.py`: 4 file-scoped-dedupe tests (incl. fail-open and backfill-suppresses-runtime) + 3 per-date-collapse tests (percentile invariance under 4× duplicated restart rows, last-wins, gate-counts-unique-dates). Also fixed pre-existing fixtures that recycled dates (`i % 28` cycling — would silently collapse under dedupe-on-read) to unique **relative** dates per the #2670 date-bomb rule.

One legacy KC 0.0 row (2026-06-09, after the duplicates collapse) remains on disk; it self-ages out of the 252-day window and its rank impact is ~1/n — flagging rather than scripting a one-off data mutation.

## Test plan

- Full suite: `4229 passed, 1 skipped, 13 warnings, 9 subtests passed in 639.09s — EXIT=0`
- Targeted (`tests/test_realized_move_rank.py` + `tests/test_realized_move_runtime_2580.py`): `44 passed`
- Fail-on-unpatched proof (source fixes reverted, tests kept): `15 failed, 29 passed` — every new pin failed, covering all three symptoms, the resolver, and the consumer collapse.

Closes #2854


rozavala:
**Review: APPROVE — Ready to be merged.**

**Value:** Closes #2854, all three symptoms at the root: file-scoped dedupe (the process-cache "idempotency" claim under `RECOVERY_POLICY ALWAYS` was the contradiction that produced every restart-day duplicate), the #2741-pattern exchange resolver (KC stops burning a daily ERROR-200 request on ICE; the legacy hop is retained as a true fallback only), and a shape-based degenerate-bar guard that covers the 0.000000 mechanism on both write paths regardless of why IB built the stub bar. The body is honest about the unproven upstream cause (HMDS placeholder hypothesis flagged, not asserted) and about the one legacy 0.0 row left to age out rather than scripting a data mutation — both the right calls.

**Verification (fresh worktree @ a667c35d7, prod venv):** targeted `44 passed`; full suite `4229 passed, 1 skipped — EXIT=0` (626s). Code-read: `_resolve_observation_exchange` follows the #2741 accessor chain exactly (commodity_overrides → engine-injected `config['exchange']` → mapped profile); dedupe polarity is right (scan failure fails *open* — a missed dedupe is bounded noise, a false positive loses the day; cache set only after successful append so a failed write no longer suppresses the retry); the consumer-side per-date collapse in `realized_move_rank` neutralizes legacy duplicates without masking the writer fix (writer tests assert raw file contents, restart test proves one physical row). Fixing the pre-existing `i % 28` date-cycling fixtures was necessary, not incidental — they'd have silently collapsed under dedupe-on-read.

Nice cross-check available day-1 post-deploy: the next restart day should produce zero new duplicate dates in any sidecar, and KC's run should log a NYBOT fetch with no Error 200 in `orchestrator_multi`.

PR #2945, opened 2026-06-22
fix(profiles): wire two execution fields the loader silently dropped (#2742 class)

rozavala:
## Summary
Two `CommodityProfile` dataclass fields were never wired into `_load_profile_from_json`'s constructor, so their JSON values were **silently dropped** and every commodity fell back to the dataclass defaults — the exact #2742 class (expiry fields, PR #2748).

## Root cause
`config/commodity_profiles.py::_load_profile_from_json` maps ~50 fields explicitly into `CommodityProfile(...)` but omitted two:
- **`market_order_fallback_timeout_seconds`** — **active divergence**: `kc.json`/`sb.json` set `90`, `cc.json` sets `120`, but all ran at the `60` default. It bounds the close-path market-order fill wait (`order_close.py:200/365`, `order_manager.py:6260` — a `for _ in range(timeout): sleep(1); if Filled: break` loop), so KC/SB/CC were giving close orders less time to fill than the operator configured.
- **`catastrophe_stop_nlv_minimum`** — inert today (all 5 profiles set `100000` = the default, consumed at `order_manager.py:4353`), but a latent trap: any future per-commodity NLV override would have been silently ignored.

## What changed
- `config/commodity_profiles.py`: pass both fields through `data.get(..., <dataclass default>)` in the constructor.

## Safety
- Behaviour change is limited to **honoring config the operator already wrote** (KC/SB close window 60→90, CC 60→120 — a *longer* fill wait, strictly more patient on the close path; NG/ZW unchanged at 60). `catastrophe_stop_nlv_minimum` is unchanged today (all profiles = default).

## Test plan
- [x] `tests/test_profile_loader_exec_fields_unmapped.py` (3): real profiles load 90/90/120; absent → defaults (60/100000); a synthetic profile round-trips arbitrary values. **Fails on `main` (2 failed), passes with the fix.**
- [x] Full suite (loader feeds execution params): **4417 passed, 1 skipped, 0 failed.**


rozavala (commented):
**PR Review — Ready to merge. Label applied.**

Verified against source. This is a real, active close-path execution bug.

**What I checked:**
- `_load_profile_from_json` on `main` (SHA `ca0f0ca9`) ends its constructor call at `market_states=...` — both `market_order_fallback_timeout_seconds` and `catastrophe_stop_nlv_minimum` are absent. Bug confirmed.
- JSON values confirmed: `kc.json` → 90, `sb.json` → 90, `cc.json` → 120, `ng.json` → 60 (= default), `zw.json` → 60 (= default).
- `market_order_fallback_timeout_seconds` feeds a `for _ in range(timeout): sleep(1); if Filled: break` loop in `order_close.py` / `order_manager.py`. KC/SB were cut off 30s short of their configured window; CC was cut off 60s short. Fix is strictly more patient — no tighter fill windows, no new risk.
- CC is deactivated in production, so 60→120 is currently inert but closes the latent trap.
- All 5 profiles' `catastrophe_stop_nlv_minimum` already equal the dataclass default (100000.0) — zero behavior change there today.
- 3 new tests fail on `main`, pass here. Pattern matches prior #2748 expiry-field fix.
- Full suite 4417 passed / 1 skipped; eval gate 448/136 all green, zero baseline drift.

**MECE:**
1. *Value* — Active close-path bug: KC/SB/CC market-order fallback was systematically shorter than operator config. Real fix, real execution impact.
2. *Risk* — Direction is safer (longer fill wait). NG/ZW unchanged.
3. *Prior art* — #2748 fixed expiry fields by the same pattern; no prior fix for these two fields.
4. *Trajectory* — Fits the #2742-class profile-loader correctness workstream.

Non-blocking follow-up filed as #2948: a completeness test that verifies all `CommodityProfile` dataclass fields are wired into the constructor, so future additions can't silently drop config again.

---

PR #2961, opened 2026-06-23
fix(order-manager): straddle gate reads compliance.max_position_pct, not a dead sub-path (#2944)

rozavala:
## Summary
- The pre-compliance straddle affordability gate now reads `compliance.max_position_pct` (the real, configurable key) instead of a non-existent `compliance.limits.max_position_pct` sub-path, and the dead `compliance.max_straddle_pct` is removed.

## Root cause
`order_manager.py` (straddle early-rejection gate, `VOLATILITY`+`HIGH`):
```python
limit_pct = config.get('compliance', {}).get('limits', {}).get('max_position_pct', 0.40)
```
`compliance` has no `limits` sub-dict (its keys are flat), so this always missed → silently `0.40`. Investigation of the compliance layer settled the "0.40 vs 0.55" question:
- **ComplianceGuardian sizes straddles with `max_position_pct` (0.40), same as directional** (`compliance.py:671`), plus a per-contract `straddle_risk_threshold` **hard block**. The "higher % limit for straddles" approach was **deliberately abandoned** ("G2 FIX: was previously just increasing limit").
- So **0.40 is correct** — it matches what compliance enforces. Loosening the pre-filter to 0.55 would admit straddles compliance then rejects at 0.40, wasting the LLM/IB cost the gate exists to save.
- `compliance.max_straddle_pct: 0.55` is **dead vestigial config** from the abandoned approach (read nowhere).

## What changed
- `trading_bot/order_manager.py`: extract `_compliance_max_position_pct(config)` (reads `compliance.max_position_pct`, default 0.40, documented to match ComplianceGuardian) and use it in the gate. **No behavior change today (0.40 → 0.40)** — the key is now actually configurable.
- `config.json`: delete the dead `compliance.max_straddle_pct`.

## Test plan
- [x] `tests/test_straddle_gate_config_path_2944.py`: the gate helper respects a configured `max_position_pct` (the broken sub-path ignored it) and defaults to 0.40; the dead `max_straddle_pct` is gone from config.json. The config-deletion assertion **fails on main** (key still present), passes with the fix.
- [x] Full suite (order_manager.py is safety-critical): **4425 passed, 1 skipped, 0 failed.**

Closes #2944


rozavala:
**Approved.**

**Value:** Genuine bug fix. The straddle affordability gate was reading `compliance.limits.max_position_pct` (a non-existent sub-path) and silently falling through to the hardcoded 0.40 default on every call. The fix reads `compliance.max_position_pct` — the same key `ComplianceGuardian` uses at `compliance.py:671` (`self.limits.get('max_position_pct', 0.40)`). The gate now actually tracks config instead of ignoring it.

**No behavior change with current config** (config has `max_position_pct: 0.40`, so 0.40 → 0.40), but the key is now correctly configurable — someone tuning `compliance.max_position_pct` will now see the pre-filter respect that value.

**Dead config cleanup:** `compliance.max_straddle_pct: 0.55` removed from `config.json`. Confirmed it's not read anywhere in the codebase (the "higher % limit for straddles" approach was abandoned in favor of the per-contract `straddle_risk_threshold` hard block in `compliance.py`). Note: `test_compliance_var.py` has this key in a local test-fixture dict — that's harmless since the code ignores it and the fixture isn't loading from `config.json`.

**Tests:** 3 focused tests verify the gate respects a configured value (the old broken sub-path silently ignored it), defaults to 0.40 matching compliance, and the dead key is gone. Full suite 4425 passed / 0 failed per PR.

<sub>Reviewed via /review-prs · config-path audit + dead-key sweep across codebase</sub>

---

PR #3064, opened 2026-07-01
fix(tests): isolate _target_env tests from ambient IB_PORT (#3062)

### LINKED ISSUE #3062: test_operate_cli::test_target_env_unknown is order-dependent — IB_PORT leaks from .env via os.environ (#3051 follow-up)
## Motivation
Full-suite runs on a machine with a real `.env` fail `tests/test_operate_cli.py::test_target_env_unknown` while the test passes in isolation. Confirmed during the #3058/#3060 reviews (two independent full-suite runs, same single failure):

```
def test_target_env_unknown():
>       assert "UNKNOWN" in operate._target_env({})
E       AssertionError: assert 'UNKNOWN' in 'PAPER gateway (4002)'
```

## Root cause
`scripts/operate.py::_target_env` deliberately falls back to `os.environ.get('IB_PORT')` (correct production behavior). Any earlier test that triggers `load_config()` loads `.env` into `os.environ`, so `IB_PORT=4002` is present by the time `test_target_env_unknown` runs → the empty-config case resolves to PAPER instead of UNKNOWN. **The production code is right; the test is under-isolated.** CI stays green only because the CI runner has no `.env` — so this is a local-full-suite trap, the exact class that makes "1 failed" noise on every future PR review.

## What this issue tracks
- `tests/test_operate_cli.py::test_target_env_unknown` — add `monkeypatch.delenv("IB_PORT", raising=False)`.
- Same hygiene for the sibling `_target_env` label tests (`test_target_env_live`/`_paper`) so they pin rather than inherit the ambient env.

## Suggested implementation
```python
def test_target_env_unknown(monkeypatch):
    monkeypatch.delenv("IB_PORT", raising=False)
    assert "UNKNOWN" in operate._target_env({})
```

## Why this isn't in #3058/#3060
Both PRs' full-suite runs merely *surfaced* the pre-existing failure; neither touches `operate.py` or its tests. Bundling an unrelated test fix would muddy their scopes.

## Related
- PR #3051 (introduced the test), PRs #3058/#3060 (surfaced it in full-suite runs)

rozavala:
## Summary
- `monkeypatch.delenv("IB_PORT")` on all three `_target_env` label tests so they assert the config lookup instead of inheriting the ambient environment.
- New `test_target_env_env_fallback` covering the deliberate production fallback (`IB_PORT=4001` env → LIVE label), so that path is now pinned by a test instead of leaking into unrelated ones.

## Root cause
`_target_env` deliberately falls back to `os.environ['IB_PORT']` (correct production behavior). Any earlier test that triggers `load_config()` loads `.env` into `os.environ`, so by the time `test_target_env_unknown` runs in a full-suite pass on a machine with a real `.env`, the empty-config case resolves to `PAPER` instead of `UNKNOWN`. The production code is right; the test was under-isolated — a local-full-suite trap invisible in CI (no `.env` there).

## What changed
- `tests/test_operate_cli.py` — env isolation on the three label tests + 1 new fallback test. Test-only; no production code touched.

## Test plan
- [x] Reproduced on main: `IB_PORT=4002 pytest tests/test_operate_cli.py::test_target_env_unknown` → FAILED
- [x] With fix: same command → 14 passed
- [x] Clean env → 14 passed

Closes #3062


rozavala:
✅ **APPROVE** — exactly the fix #3062 prescribed, with a bonus.

**Verified the order-dependence is dead by reproducing the failure condition:** `tests/test_operate_cli.py` passes standalone (**14 passed**) *and* with `IB_PORT=4002` exported into the environment (**14 passed**) — the precise condition under which `test_target_env_unknown` failed in both of my prior full-suite runs (4511 passed / 1 failed, twice). The three `_target_env` tests now `monkeypatch.delenv("IB_PORT")` so they assert the config lookup, not the ambient env.

**The bonus:** `test_target_env_env_fallback` pins the *deliberate* production behavior (empty config + `IB_PORT` in env → labelled) — so the fallback that caused the pollution is now itself under test rather than merely tolerated.

**MECE:** value=real (removes the "1 failed" noise from every local full-suite run — which had to be explained away in two PR reviews already); risk=none (tests-only, no production code); prior=implements #3062 as filed; fit=direct. Closes #3062 on merge (plain "Fixes" — verify auto-close anyway given the recent bold-markdown misses).

Merge note: land this one **first** — it greens the full-suite baseline for the other two open PRs' branches.

PR #3162, opened 2026-07-08
fix(reconcile): make missing-trades an alerting exception queue, not a permanent data layer (#3158)

rozavala:
# Make the missing-trades file an alerting exception queue

## Summary

The reconciler's `trade_ledger_missing_trades.csv` was being treated as a
permanent data layer: fully overwritten on every run and silently
concatenated into all ledger netting forever. This PR turns it into an
**alerting exception queue** — synthesizing a missing trade now alerts the
operator, the write is merge-preserving so an unpromoted row can't be
destroyed, and there is an atomic promote-and-clear workflow to fold queued
rows into the permanent dated archive.

Scoped to the reconciler write/promote path in `reconcile_trades.py`.
`get_trade_ledger_df` (the consumer) is intentionally left unchanged, per the
#2574 guardrail ("fix false positives at the reconciler, never filter at
consumers").

## Root cause

- `write_missing_trades_to_csv` (`reconcile_trades.py`) called
  `final_df.to_csv(output_path, ...)` with `header=True` and no append mode —
  a **full overwrite** of the queue on every run that found any missing trade.
- `get_trade_ledger_df` globs `archive_ledger/` for
  `startswith('trade_ledger_') and endswith('.csv')`, which matches
  `trade_ledger_missing_trades.csv` — so every queued row is concatenated into
  the ledger indefinitely.
- On 2026-07-06, reconciliation synthesized a new OZWU6 C0620/C0660 set. The
  overwrite destroyed the previously-synthesized OZWZ6 C0740/C0830 closes for
  `ZW-8faecb9b56d24c9694c2` (entered 2026-05-22). With only the entry left in
  the archives, the position netted to a phantom-open, producing 2x/day
  `State sync discrepancy: IBKR has 0 positions but ledger shows 2 open`
  WARNINGs and daily #3151 invariant pages.

## What changed (`reconcile_trades.py`)

**(a) Alert on synthesis** — `write_missing_trades_to_csv` now emits a
`logger.warning` and a Pushover notification (`🟠 Missing Trades Synthesized`,
via the existing `send_pushover_notification`) whenever it synthesizes missing
trades. It only fires on genuinely-new discrepancies: once a row is queued it
becomes part of the concatenated local ledger and is no longer flagged
missing, so the alert is naturally rate-limited to real events.

**(b) Merge-preserving write** — the write now reads the existing queue,
unions it with the newly-synthesized rows, and dedupes on a **stable per-fill
key** before writing atomically (temp file + `fsync` + `os.replace`):

- Primary key: `combo_id`, which for synthesized rows is the IB
  `TransactionID` (`parse_flex_csv_to_df` sets `combo_id = TransactionID`) —
  globally unique per execution and reported identically on every Flex pull.
- Fallback (only when `combo_id` is empty — legacy/hand-edited rows):
  `(local_symbol, action, strike, right, fill_date, quantity)`.
- Deliberately **not** keyed on `position_id`: it is re-resolved per run by
  `_resolve_original_position_ids` and is the least stable field.

A run that finds a new missing-trade set now preserves prior rows instead of
destroying them.

**(c) Promote-and-clear** — new `promote_missing_trades()` + CLI:

```
python reconcile_trades.py --promote-missing [--date YYYYMMDD] \
    [--commodity TICKER | --data-dir DIR] [--dry-run]
```

It folds queued rows into the dated `trade_ledger_<YYYYMMDD>_v2.csv` archive
(one file per fill date; the 2026-05-08 precedent format — `strike` stays a
bare `740.0`, prices `%.2f`), then clears them from the queue. If a `_v2` file
for the date already exists it is **merged + deduped** (idempotent), not
overwritten.

**Atomicity / crash-safety.** Consumers concatenate *both* the queue and the
`_v2` archive, so there is no single-syscall way to move rows between the two
files — exactly one intermediate state is unavoidable on a crash. We write the
`_v2` archive **first** and clear the queue **last**:

- crash between → rows in **both** files → a *recoverable, idempotent
  double-count* (re-running `--promote-missing` re-merges to the same `_v2`
  and re-clears the queue). Visible via P&L / the #3151 invariant page.
- the reverse order (clear queue first) → crash **loses** the synthesized
  close → the phantom-open position silently resurrects (the exact #3158
  failure).

We prefer a visible, self-healing double-count over silent data loss. If the
`_v2` write raises, we abort **before** touching the queue, so no row is ever
lost. The failure window is just the gap between the archive rename(s) and the
queue rename.

## Test plan

New: `tests/test_missing_trades_queue_3158.py` (14 tests) — row shapes mirror
the real live files (strike/right/combo_id as strings, tz-aware timestamps):

- merge-preserving write: a new set preserves prior rows; re-synthesis of the
  same set dedupes; same fill re-resolved to a different `position_id` still
  dedupes (combo_id primary); no stray temp files.
- alert fires on synthesis; no alert / no crash when `config` is `None`.
- promote-and-clear: rows land in the dated `_v2`, queue emptied, precedent
  format preserved; **no double-count through `get_trade_ledger_df` before AND
  after** promotion; idempotent re-promote; `--date` filter leaves other dates
  queued; empty-queue no-op; dry-run writes nothing; fail-closed (queue
  preserved) when the `_v2` write raises.

Targeted run (new file + all reconcile/ledger suites touching the changed
functions): **94 passed** —
`test_missing_trades_queue_3158`, `test_reconcile_partial_flex_failure`,
`test_reconcile_2579_flex_dedup`, `test_reconciliation`,
`test_resolve_original_position_ids`, `test_ledger_defensive_read`,
`test_ledger_data_dir_audit_2680`, `test_catastrophe_pid_reuse_2654`,
`test_decision_signals_concat_2712`, `test_pulse_reconciliation`,
`test_post_close_retry_position_id_2458`, `test_active_positions_expiry_filter`,
`test_flex_query_1001_retry`.

## Still open (not in this PR)

- **Live-data cleanup** (operator-side): promote the current OZWU6 queue rows,
  re-synthesize the destroyed OZWZ6 closes, re-baseline the #3151 ZW invariant.
  Runbook in the issue thread.
- **Unrecorded-close root cause** (#2828-adjacent): the orchestrator's
  reconciliation invalidates the TMS thesis on an externally-closed position
  but does not write a ledger close row, so the only ledger record of these
  closes is the fragile Flex-synthesized queue. Needs its own issue.

Full suite (`pytest tests/ -q --timeout=120`): **4761 passed, 1 skipped, 0 failed** (11:49).

Part of #3158

https://claude.ai/code/session_014heGyAvN5VjtLbN2PuQzFQ


rozavala:
**PR Review — Approved ✓**

This is the legitimate evolution of the #2574 decision, not a re-litigation of it — verified in detail first, since that was the crux of this review.

**The #2574 question:** #2574's ruling was "fix false positives at the reconciler, never filter at consumers" — its resolution chain actively reverted a consumer-side skip (PR #2592 re-included `missing_trades*.csv` in `eod_aggregate`). This PR complies on every axis:
- Only `reconcile_trades.py`'s write/promote path is touched — the exact place #2574 designated. Zero consumer changes (verified: `get_trade_ledger_df` ×3, `eod_aggregate.py:147`, `close_logic.py:381`, `strategy_outcome_feedback.py:24` all untouched, still including missing_trades per #2592).
- Every row remains consumer-visible at all times: both the queue and the `_v2` archives match the consumers' `trade_ledger_*.csv` glob; downstream treatment keys on the `reason` column (`RECONCILIATION_MISSING`, preserved through promotion), not the filename. Promotion moves a row between two consumer-visible files — nothing ever silently vanishes from netting.
- #2574's actual invariant — the netting contribution of synthesized rows must never be silently dropped — is *strengthened* here: the pre-PR full-file overwrite is precisely what violated it on 2026-07-06.

**Code verified:**
- Root cause is real: pre-PR `final_df.to_csv(output_path, ...)` full overwrite; the live ZW queue's current content is exactly the two 07-06 OZWU6 rows that destroyed the OZWZ6 closes.
- Dedupe key verified: `combo_id` = IB `TransactionID` (reconcile_trades.py:1116); not keying on `position_id` is correct since `_resolve_original_position_ids` re-resolves per run.
- Merge sort safe: timestamps stringified to a fixed format (:569) before the lexicographic `sort_values` — chronologically correct.
- Alert rate-limiting verified: the local side of the match includes the queue file, so re-runs don't re-alert; new logs are WARNING-level → no `error_reporter` auto-filing.
- Deploy safety: no migration, no deletion — existing ZW/SB queue rows preserved and merged on next run; promotion is a manual operator CLI; `pages/5_Utilities.py:979` (no-arg invocation) unaffected.
- Fail-closed end-to-end: promote writes `_v2` first / clears queue last (crash between = idempotent, visible double-count — test-pinned); corrupt/absent queue → promote no-ops; reconciliation never blocks trading on queue errors.
- Live-data end-to-end (on a COPY of `data/ZW`): dry-run plans correctly; real promote produced a byte-identical-format `_v2` file, cleared the queue, re-promote no-op'd. All 16 existing live `_v2` files across KC/NG/SB/ZW match the 11-column schema, so the `reindex(columns=...)` merge cannot strip columns on real data. Consumer check: `get_trade_ledger_df` from all 3 modules returns identical rows before/after.

**MECE:**
1. **Value**: High — fixes live, recurring production data destruction (2×/day ZW state-sync WARNINGs, daily #3151 invariant pages, permanent ZW P&L corruption from a resurrected phantom-open). Root cause confirmed in code and live data.
2. **Risk**: Low — reconciler write-path only; consumers untouched; atomic temp+fsync+`os.replace` writes; fail-closed promote ordering with the loss-vs-double-count trade-off argued and tested; no deploy migration; backward-compatible CLI.
3. **Prior attempts**: #2574 is the governing prior art and this PR is on the right side of it; PR #2581 (parser dedup) and #2592 (consumer re-inclusion) are the precedents and neither is contradicted.
4. **Workstream fit**: Part of the #3158/#3151/#3160 phantom-position cluster; complements (doesn't mask) the #3160 unrecorded-close root cause — the new alert is what makes the next #3160 occurrence visible within a day.

Residuals filed as #3164 (header-only-queue FutureWarning of the #2698 class — verified reproducible; explicit-commodity guard for promote; `n_added > 0` alert gate; dead `config` param).

**Tests**: 14 new + 134 targeted reconciliation/ledger tests passed; full `pytest tests/` on the PR worktree: **4761 passed, 1 skipped, 9 subtests passed** in 11m42s, exit 0.

---

PR #3167, opened 2026-07-08
polish(reconcile): promote-path residuals — empty-queue FutureWarning, explicit-commodity promote, n_added notify gate, live config param (#3164)

rozavala:
## Summary
Clears the four non-blocking residuals from the PR #3162 (missing-trades exception queue) review, bundled by issue #3164. All four are small polish items on the new promote path; the core merge/dedupe/promote-and-clear design from #3162 is unchanged.

## Per-item root cause & choice made
**1. Header-only queue trips the #2698 empty-concat FutureWarning (fixed both sides).**
`performance_analyzer.get_trade_ledger_df` (`performance_analyzer.py:107`) concatenated every loaded archive frame unconditionally, and a fully-promoted `trade_ledger_missing_trades.csv` reads as a 0-row frame. Reproduced: **1 FutureWarning** with a header-only queue present, **0** without.
- (a) Mirrored `order_manager.get_trade_ledger_df`'s #2698 empty-frame drop (`trading_bot/order_manager.py:5328-5339`) into `performance_analyzer` — kept the comment consistent (transcription-drift pair).
- (b) `promote_missing_trades` now `os.remove`s the queue file when nothing remains after promotion, rather than leaving a header-only file. Verified absence is handled: `_read_missing_trades_queue` returns an empty frame when the file is absent (`reconcile_trades.py:445-446`). The write-`_v2`-first / clear-queue-last crash ordering is preserved (`os.remove` is a single atomic syscall; a crash still leaves at worst a recoverable double-count, never data loss).

**2. Bare `--promote-missing` silently defaulted to KC (footgun) — now errors.**
Old resolution: `(args.commodity or os.environ.get("COMMODITY_TICKER", "KC")).upper()` (`reconcile_trades.py:1411`). **Choice:** require an explicit target (`--commodity` / `--data-dir` / `COMMODITY_TICKER`). Extracted `_resolve_promote_data_dir()` (unit-testable) which raises `ValueError` listing the available `data/` dirs when none is set; the CLI turns that into `parser.error` (exit 2). The per-engine orchestrator's `COMMODITY_TICKER` path is preserved.

**3. `_notify_missing_trades_synthesized` fired on `n_added == 0`.**
`write_missing_trades_to_csv` called the notifier unconditionally (`reconcile_trades.py:598`) even when dedupe absorbed every row. **Choice:** gate the Pushover on `n_added > 0`, and — as the issue asked me to decide — **downgrade the WARNING to an INFO "queue unchanged" line** when `n_added == 0`, since re-detecting already-queued rows (reconciliation runs every cycle) is not a new event and shouldn't warn or alert.

**4. Dead `promote_missing_trades(config=...)` param — now used.**
Per the issue's preferred option, `config` now drives a promote-confirmation Pushover (`🟢 Missing Trades Promoted`) mirroring the synthesis alert, so the operator sees both ends of the queue lifecycle. Gated on rows actually promoted — the early returns cover empty-queue / no-promotable-rows / dry-run, so it only fires after a real `_v2` write.

## What changed
- `performance_analyzer.py` — empty-frame drop before `pd.concat` (item 1a).
- `reconcile_trades.py` — `os.remove` on drained queue (1b); `_resolve_promote_data_dir` + `_available_commodity_data_dirs` + CLI guard (2); `n_added > 0` gate + INFO downgrade (3); `_notify_missing_trades_promoted` wired into `promote_missing_trades` (4).
- `tests/test_missing_trades_queue_3158.py` — extended with 13 #3164 tests; updated one existing #3158 test whose expectation changed (drained queue now removed, not header-only).

## Test plan
- Targeted module: `pytest tests/test_missing_trades_queue_3158.py` → **27 passed** (14 pre-existing + 13 new).
- **Fail-on-unpatched proof:** reverse-applied the prod-file diff (`git apply -R`, not stash) → **9 tests failed** (one per item: 1a header-only FutureWarning, 1b queue removal + the updated existing test, 2× the four resolver tests, 3 the re-fire gate, 4 the promote-notify) → re-applied forward → 27 pass.
- CLI end-to-end: bare `--promote-missing` (no env) → `parser.error`, exit **2**, no KC default; `--data-dir … --dry-run` and `COMMODITY_TICKER=ZW … --dry-run` both resolve correctly.
- FutureWarning reproduction: header-only queue → 1 FutureWarning on unpatched, **0** after fix.
- CI-exact `mypy` (no args, worktree root): **Success: no issues found in 51 source files**.
- Full `pytest tests/` deferred to the reviewer per the 2-core live-host constraint.

https://claude.ai/code/session_01GCtjiLG2jB86PFC2FZJRWr


rozavala:
Rebased onto main after #3166 landed — the two PRs added different functions at the same location in `reconcile_trades.py` (supersede helpers vs promote notifier); resolution keeps both, with `_supersede_external_close_placeholders` correctly outside the `n_added > 0` notify gate per the #3166 re-review instruction. Full suite on the composed tree: **4844 passed, 1 skipped, 9 subtests passed** in 13m02s, exit 0.

PR #3362, opened 2026-07-16
fix(risk): correct credit-branch max_loss units — restores the disabled iron-condor stop-loss (#3352)

rozavala:
## What

Mirror #3334's units correction to the **credit branch** of
`_calculate_combo_risk_metrics` (`trading_bot/risk_management.py`). The
credit-spread `max_loss` used the per-cent multiplier (`get_dollar_multiplier`)
against a **dollar-denominated** strike width, underflowing the width term 100x
on cents-quoted commodities (KC/SB/ZW). Swap it to the full contract multiplier
(`get_contract_multiplier`), exactly as #3353/#3334 did for the debit-branch
`max_profit`. NG is unaffected (the two multipliers are equal there).

Closes #3352.

## Why it matters (this one is a stop-loss decision, not observability)

Unlike the debit branch #3334 fixed, this `max_loss` feeds a **trading
decision**:

```
risk_pct = pnl / max_loss            # risk_management.py
... risk_pct <= -stop_loss_max_risk_pct   # default 0.50 -> Stop-Loss
```

Only **iron condors** exercise the credit branch (bull-call / bear-put /
long-straddle are all debit). With the 100x-underflowed `max_loss`, the
stop-loss on an IC was wrong in one of two ways depending on the credit vs the
underflowed wing width:

- **Typical IC (verified on `main`)**: the underflowed wing term is *smaller*
  than the credit collected, so `max_loss` goes negative, the `<= 0 -> inf`
  guard clamps it, `risk_pct` pins to `0`, and the IC stop-loss is **silently
  disabled**. Confirmed empirically: a realistic KC IC that has lost 50% of its
  real max loss reports `max_loss = inf`, `risk_pct = 0.0` on `main`.
- **Atypical corner (tiny credit / very wide wings)**: a small positive
  `max_loss` makes `risk_pct` ~100x too large and the stop fires on a trivial
  adverse move.

Note this **corrects the issue's "fires ~100x too early" framing**: for
realistic credits the guard clamps to `inf` and the stop never fires at all —
the more dangerous failure mode. Either way the denominator was nonsense; the
fix restores it to the IC's true max loss (`wing_width_$ − credit`).

## Why `stop_loss_max_risk_pct` (0.50) needs NO recalibration — the crux

The 0.50 threshold is the **same value the sound debit branch already uses**,
and after the fix it carries the **identical meaning** on both branches:

- Debit branch: `max_loss = total_entry_cost` (the debit). `risk_pct <= -0.50`
  ⇒ "unrealized loss ≥ 50% of the debit paid" = 50% of the position's real max
  loss (for a debit spread the debit *is* the max loss).
- Credit branch (fixed): `max_loss = wing_width_$ − credit` = the IC's true max
  loss. `risk_pct <= -0.50` ⇒ "unrealized loss ≥ 50% of the IC's real max
  loss."

So the fix makes the credit-branch stop-loss **consistent** with the
(already-correct) debit-branch stop-loss semantics — it does not require a new
threshold. `stop_loss_max_risk_pct` is **not** changed. The load-bearing test
`test_credit_risk_pct_is_scale_correct_and_stop_fires_at_threshold` pins this:
at 50% of real max loss `risk_pct == -0.50`; the stop fires at 55% and holds at
45%.

## The degenerate `max_loss <= 0 -> inf` guard, re-examined

Preserved, and now argued: with **corrected** units, `max_loss <= 0` requires
`credit ≥ wing_width_$` — a genuinely degenerate/mispriced combo (collecting
more credit than the wing width is a risk-free arb). For any sane IC the credit
is a fraction of the wing width, so `max_loss > 0` and the stop is **live**.
Pre-fix, the guard fired on essentially *every* realistic IC (underflowed wing
≪ credit), which is exactly how the stop got silently disabled. The fix makes
the guard fire only where intended, and — mirroring #3334's WARNING-on-fallback
discipline — it now **logs a WARNING** instead of silently disabling the stop
(a silent fallback on the primary path is what hid #3334).

## Latency: currently dormant, but real (not dead code)

- The credit branch is IC-only; the risk monitor's 4-leg IC path (`len == 4`,
  which logs `"Identified Iron Condor"`) does **not** fire in the current logs,
  consistent with the #3334 finding that all 1,708 combos in the examined
  window are 2-leg. So the bug is **dormant on the current position book**.
- It is **not** dead code: the council still selects `IRON_CONDOR` as a strategy
  (KC as recently as 2026-07-16), and iron condors have been chosen historically
  (Jan–Feb 2026). The moment an IC fills and the risk monitor evaluates it, this
  branch runs — so this is a real latent stop-loss bug, not a theoretical one.
- Whether the credit branch ran on historical ICs can't be confirmed from
  currently-retained logs (they span only a narrow recent window); this PR does
  not rely on that.

## Scope / safety

- Change is confined to the credit branch of `_calculate_combo_risk_metrics`
  (does not touch `_execute_risk_closure` / #3356, the debit branch, or
  `stop_loss_max_risk_pct`). No hardcoded constants — uses the canonical
  `get_contract_multiplier` helper.
- Fail-closed preserved (degenerate guard intact + now WARNs).

## Tests (`tests/test_risk_management_max_loss_units.py`, new)

- Per-commodity credit-branch `max_loss` regression (KC 37500, NG 10000,
  SB 112000, ZW 5000) pinned to a hand-computed wing width — **fails on `main`**
  for KC/SB/ZW (inf), NG is the unchanged control.
- `risk_pct` scale + stop-loss decision: `risk_pct == -0.50` at 50% of real max
  loss; Stop-Loss fires at 55%, holds at 45%.
- Degenerate `max_loss <= 0 -> inf` guard still clamps → `risk_pct 0` → stop
  cannot fire, and now emits a WARNING.

Targeted run (green):
`tests/test_risk_management*.py tests/test_exit_*.py` → 132 passed.

---

## Reviewer verification (main session)
- **Full suite: `5511 passed, 1 skipped`** (14:42), serial under flock, `RO_PROD_ROOT` scoped to the worktree. Live tree unchanged (accuracy 941).
- **Rebased onto post-merge main** (all four prior PRs merged at 22:25 mid-run); verified #3334's debit fix + this credit fix coexist (`contract_mult` defined once at function scope, used by both branches), disjoint from the concurrent #3356 (`_execute_risk_closure`) — 0 conflicts.
- **Corrected the issue's severity, empirically.** I filed #3352 as "IC stop fires ~100x too early." The real behavior, verified on main with a realistic KC IC (0.10 wing, ~$1000 credit): the underflowed wing term ($37.50) is *smaller* than the credit → `max_loss` goes negative → the `<=0 → inf` guard clamps it → **the IC stop-loss is silently DISABLED** (`risk_pct = 0`, fail-open on a risk control). "Fires too early" only in the atypical wide-wing/tiny-credit corner. Both are wrong; the fix restores the true `max_loss = wing_width_$ − credit` either way, and now WARNs when the guard clamps instead of silently disabling.
- **Live, not dead code**: 0 executed ICs in the canonical ledger (the only reason it's never bitten), but the council selected IRON_CONDOR **137×** — the next filled IC would have a disabled stop.
- **`0.50` needs no recalibration** — after the fix `risk_pct <= -0.50` means "lost 50% of the IC's real max loss", the identical semantic the sound debit branch already carries. `stop_loss_max_risk_pct` unchanged. Pinned by a risk_pct-scale test (fires at 55%, holds at 45%).
- **Existing IC test resolved as a stale mock, not a regression.** `test_risk_ic_max_loss.py` patched `get_dollar_multiplier=1.0` to simplify the synthetic max_loss; this PR moves the width term to `get_contract_multiplier` (which the test didn't patch → real KC 37500 leaked in → stop stopped firing, full-suite-only failure the agent's targeted run missed). Patched BOTH multipliers to 1.0; the test's intent (IC stop fires past threshold) and expected `max_loss=6` are unchanged.

Closes #3352


rozavala:
**Reviewed — ready to merge** (batch protocol: labeled, awaiting batch call).

Verified independently:
- **Exact mirror of the #3353 correction** on the branch that matters more: the credit-branch `max_loss` feeds `risk_pct` → the stop-loss. The one-line multiplier swap restores `max_loss = wing_width_$ − credit`, and `contract_mult` is reused from #3353's function-scope definition — no drift between the two branches.
- **The severity self-correction is right and matches the code**: with realistic credits the underflowed wing term went *negative* → `<= 0 → inf` guard → `risk_pct` pinned at 0 → **IC stop-loss silently disabled** (fail-open on a risk control), worse than the issue's original "fires too early" framing. The degenerate guard is preserved and now WARNs — with corrected units it fires only on a genuine credit ≥ wing-width mispricing.
- **No threshold recalibration needed** — argued and pinned: post-fix, `risk_pct <= -0.50` means "lost 50% of real max loss" on both branches (fires at 55%, holds at 45%, `test_credit_risk_pct_is_scale_correct_and_stop_fires_at_threshold`).
- **The stale-mock resolution is the honest kind**: `test_risk_ic_max_loss` patched only `get_dollar_multiplier`; the fix moves the width term to `get_contract_multiplier`, so the test now patches both to 1.0 with intent and expected values unchanged.
- **Dormant-but-real** confirmed: 0 filled ICs in the ledger, but the council has selected IRON_CONDOR repeatedly — the next filled IC gets a live stop instead of a disabled one.
- **Full suite: 5510 passed, 1 skipped + only the known #3363 Friday date-bomb** (`test_close_row_inherits_counter_trend`, pre-existing, fails identically on `origin/main` — not this PR). New/updated test files 7/7 with `-v` proof.

Disjoint from #3361 (`_calculate_combo_risk_metrics` vs `_execute_risk_closure`) — both can land in the same batch.

Issue #3364, opened 2026-07-17
bug(tests): test_close_row_inherits_counter_trend is a Friday date-bomb — full suite fails every Friday (#3317 follow-up)


rozavala:
## Summary
`tests/test_trend_close_lane_integration_3317.py::TestCloseStaleLane::test_close_row_inherits_counter_trend` **fails every Friday**. It drives the real `_close_stale_positions_with_connection` lane via `_drive_close_stale`, which does NOT control the clock, and asserts the close reason is `COUNTER_TREND_MAX_HOLD`. On a Friday the lane's internal `datetime.now(timezone.utc).weekday()` triggers the **Friday weekly-close**, which overrides the counter-trend max-hold reason:

```
assert row['reason'] == ExitReason.COUNTER_TREND_MAX_HOLD.value
AssertionError: assert 'Friday Weekly Close' == 'Counter-Trend Max Hold'
```

Reproduced on **clean origin/main** today (2026-07-17, Friday). The sibling `test_close_row_inherits_entry_trend` (WITH_TREND) passes because WITH_TREND is **Friday-exempt** at close time (#1662/#1946/#2456) — only the counter-trend case is Friday-sensitive.

## Impact
Any full `pytest tests/` run on a Friday goes red on this one test. It surfaced blocking the Friday full-suite verification of unrelated PRs (#3356, #3351). It's the classic date-bomb pattern (relative fixtures that assume a weekday).

## Root cause
`_drive_close_stale` (the shared driver for `TestCloseStaleLane`) leaves the lane's `datetime.now()` on the real clock. The counter-trend assertion is only valid on a **non-Friday** weekday; on Friday the weekly-close reason wins.

## Fix
The file already has the seam: `_MostRecentFridayClock(datetime)` — a `datetime` subclass whose `now()` returns the most recent Friday via a **relative** offset (so it can't expire), applied via `patch.object(order_manager, 'datetime', ...)`. Add the non-Friday analogue and pin `_drive_close_stale` to it, so the whole `TestCloseStaleLane` runs on a deterministic non-Friday weekday regardless of the real day. WITH_TREND still closes (10-day cap, closes any weekday); COUNTER_TREND closes with `Counter-Trend Max Hold` (not overridden). Mirrors memory `project_test_date_bomb_pattern` (control the date; never assume a weekday) and `feedback_cross_module_mock_seam` (mock the clock at the caller's namespace — here `order_manager.datetime`).

## Related
[...]

PR #3366, opened 2026-07-17
fix(risk): risk-closure ledger writes inherit the entry position_id + repair the 07-16 split rows (#3219 readout follow-up)

rozavala:
## Summary

The risk-monitor close lane minted a **fresh** `position_id` on every ledger write instead of inheriting the entry's, so each stop-loss/take-profit close split the round trip across two pids and corrupted all position-keyed P&L netting. This threads the entry pid into the risk-closure write (mirroring the already-correct `order_close` / `emergency` lanes) and ships a one-time migration to repair the one round trip already split on disk.

## Root cause

`risk_management._execute_risk_closure` called `log_trade_to_ledger(...)` **without a `position_id` kwarg**, so `log_trade_to_ledger` generated a new id. The first production STOP_LOSS — the 2026-07-16 KC close — wrote its close legs under a brand-new pid `KC-918527ee82ca42458082` (proceeds +$1,541.25) while the entry (07-07) sits under `KC-c32edb70ba9d467d9236` (debit −$3,243.75). Position-keyed netting split the round trip in two, so the console trade-analytics scored a real **−$1,702.50 LOSS as a +$1,541 WIN** (verified live via `/api/trade-analytics`).

PR #3355 fixed the per-engine **directory routing** of this same write; pid identity was out of its scope. The other close lanes (`order_close`, `emergency`) already inherit the entry pid — the #3219 readout confirmed the other 9 recent closes net fine. This completes the #3348 lane fixes: **#3355 = routing, this = identity.**

## Code fix + fail-safe

- `_entry_pid_and_trend_for_combo` (extends the #3311 `_entry_trend_for_combo` lookup) resolves the entry `(position_id, trend_alignment)` from the **same** direction-matched entry row (long leg → most-recent BUY, short leg → most-recent SELL, by `localSymbol`). `_entry_trend_for_combo` is preserved as a thin wrapper (its direct tests stay green).
- `_execute_risk_closure` takes an `entry_position_id` kwarg and passes it to `log_trade_to_ledger`.
- **Fail-safe:** if no entry row resolves, it falls back to today's behavior (fresh pid) and logs a loud `#3219` WARNING — the close write is **never** blocked (an executed fill must always land, #2397).

Notably the 07-16 close rows already carry `WITH_TREND` (the #3311 trend lookup resolved at close time from that same entry row), which confirms the pid lookup would have resolved too.

## Migration (`scripts/migrations/repair_risk_close_split_pid_3219.py`)

Re-keys every ledger row whose `position_id` == the fresh close pid `KC-918527ee82ca42458082` → the entry pid `KC-c32edb70ba9d467d9236`, in both the live `data/KC/trade_ledger.csv` and any dated `archive_ledger/*.csv` snapshot holding them. Matches on the exact, globally-unique close pid, so no other row can be touched.

**Per-host safety / idempotency:** surgical field-bounded token replacement (every other byte preserved, header/schema untouched); atomic temp+replace; apply-by-default + `--dry-run`. A second run finds no row still carrying the old pid → no-op. PROD trades the LIVE account and never held these PAPER rows → every file a no-op. Mirrors the #3275 (`repair_missing_trades_phantom_3158`) migration shape.

## Test plan

- `tests/test_risk_close_pid_inheritance_3219.py` — resolver unit tests (pid+trend from one row; pid resolves even when trend blank; orphan → `(None, None)`; wrapper delegation) + full-lane integration: the STOP_LOSS CLOSE row is keyed to the entry pid (**fail-on-main**), and an unresolvable entry falls back to a fresh pid + WARNING with the close still landing.
- `tests/test_migration_risk_close_split_pid_3219.py` — re-keys exactly the two rows (live + archive), idempotent second run, no-op on data without them, dry-run writes nothing, non-KC dir skipped, byte-for-byte surgical.
- Suites: `test_risk_close_pid_inheritance_3219` + `test_migration_risk_close_split_pid_3219` + `test_trend_on_close_rows_3311` (39 passed); `test_risk_management_new` + `test_risk_ic_max_loss` (passed); `test_order_manager.py` (83 passed). Migration verified end-to-end against a copy of the live KC ledger (dry-run → apply → idempotent re-run → surgical diff).
- **Pre-existing failure, not from this change:** `test_trend_close_lane_integration_3317.py::TestCloseStaleLane::test_close_row_inherits_counter_trend` — a Friday date-bomb in the `order_manager` close-stale lane (that test doesn't pin a fixed weekday clock; today is Friday so weekly forced-close overrides `COUNTER_TREND_MAX_HOLD`). Reproduced on pristine `origin/main`. This PR touches zero lines of `order_manager.py`.
- Scoped-strict mypy clean on both touched source files (zero new errors vs the `origin/main` baseline).

## Collision check vs #3352

PR #3362 (fix/3352-credit-maxloss-units) touches `risk_management._calculate_combo_risk_metrics` (~L351) and its tests. This PR touches `_entry_pid_and_trend_for_combo` / `_entry_trend_for_combo` (~L447) and `_execute_risk_closure` (~L722) + the caller (~L919). **Disjoint regions — no collision.**

https://claude.ai/code/session_01GCtjiLG2jB86PFC2FZJRWr


rozavala:
**Content approved — held for a rebase, not labeled yet.**

Full review done; everything checks out:
- **The identity fix completes the #3348 lane parity** (#3355 = routing, this = identity) and explains the two-pid split my #3348 forensics documented (`KC-918527ee…` close vs `KC-c32edb70b…` entry). The resolver extends the #3311 trend lookup from the same direction-matched entry row (which demonstrably resolved at close time on 07-16 — the close rows carry `WITH_TREND`), the fail-safe direction is right (unresolvable → fresh pid + loud WARNING, close never blocked, #2397), and `_entry_trend_for_combo` survives as a wrapper so its tests stay green.
- **The migration is safe by construction, verified by inspection**: comma-bounded exact-token match on CSV field index 1 (cannot hit the header, numeric combo_ids, or the reconciler's missing_trades synthetics — those already carry the entry pid), atomic mkstemp+fsync+`os.replace` with error cleanup, idempotent (needle absent after one run), non-KC dirs skipped, PROD trivially no-op, dry-run supported. Mirrors the accepted #3275 migration shape.
- **Impact is real**: the console scored the −$1,702.50 round trip as a +$1,541 WIN; the re-key restores position-keyed netting, which also matters for the #3219 exit-timing readout data.
- **Full suite: 5521 passed, 1 skipped + only the known Friday date-bomb** (#3364/PR #3367, pre-existing). New test files 17/17. Fail-on-main holds via import barrier; the behavioral proof is the live incident.

**Why no label yet**: this branch **conflicts with PR #3361** (verified via `git merge-tree`: content conflict in `risk_management.py` — both modify `_execute_risk_closure`'s signature, the `_check_risk_once` call site, and the adjacent resolver-helper region; the PR body's collision check covered #3362 but not #3361). #3361 is already in the approved batch, so this PR flips DIRTY the moment that lands. To avoid the gh auto-close-on-conflict hazard during the batch merge: **rebase onto main after #3361 merges**, resolve the (mechanical) overlap — both kwargs coexist: `thesis_ids` + `entry_position_id` — and re-push. Post-rebase this only needs a conflict-resolution sanity check, not a fresh review; the content verdict above stands.

